PR #14, opened 2017-06-21
Fetch released migrations via the GitHub API

### LINKED ISSUE #6: Fetching the released migrations pollutes the local repo
When the migrations lived in ManageIQ proper, fetching a different branch locally was ok, because it was the same repo.  Now that the schema has been separated out, fetching the ManageIQ/manageiq code *into* the ManageIQ/manageiq-schema repo pollutes the repo with the remote branches and tags.  Unfortunately those tags will collide, for example when we cut gaprindashvili-1 on both repos, then try to fetch, you will get duplicate tag names, and I have no idea what will happen.  We need an alternative now.

cc @carbonin

Fryguy:
This reverts the released migrations to using the GitHub API as was done
in 9fba1a424a3ca3b1a1f534b6ab623527e9fd33d5.  The concern for the GitHub
rate limits in 9568c7aad74c1d5a58d88f04fe824f90fb883293 is no longer a
problem now that migrations are in their own repo which should have much
less churn, unlike the main manageiq repo where the migrations would run
on *every* PR.

Fixes #6 

@carbonin Please review.

### DIFF
--- a/lib/tasks_private/spec.rake
+++ b/lib/tasks_private/spec.rake
@@ -30,22 +30,15 @@ class SetupReleasedMigrations
   private
 
   def released_migrations
-    return [] unless system(fetch_command)
-    files = `git ls-tree -r --name-only #{TEST_BRANCH} db/migrate/`
-    return [] unless $?.success?
+    require 'net/http'
+    json = Net::HTTP.get(URI("https://api.github.com/repos/ManageIQ/manageiq/contents/db/migrate?ref=#{RELEASED_BRANCH}"))
 
-    migrations = files.split.map do |path|
-      filename = path.split("/")[-1]
+    migrations = JSON.parse(json).map do |h|
+      filename = h["path"].split("/")[-1]
       filename.split('_')[0]
-    end
+    end.sort
 
     # eliminate any non-timestamp entries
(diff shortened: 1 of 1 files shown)


carbonin:
Looks good except for what looks like a missing require.

chessbyte:
@Fryguy Travis Failure for Ruby 2.4

Fryguy:
I don't understand it...I run Ruby 2.4 locally and it works just fine.  It can't be something GitHub related or *both* versions would fail.  @carbonin Maybe the sporadic failures in the past weren't because of rate limiting?

carbonin:
Could be ... I also tried locally with 2.4 and didn't have any problems.

Do you think it would pass if we re-run it? (not that this would be a good thing necessarily).

Fryguy:
> Do you think it would pass if we re-run it

Just kicked it...let's see

carbonin:
😟

Fryguy:
aaaand it's green ☹️   (👈  never thought I'd be sad about something going green 😕 )

Fryguy:
OK, let's merge and deal with consequences if it's truly sporadic.

@chessbyte Please merge.  What we have here is better than *not* doing it, as *not* doing it means we leave devs with "broken" repos.

PR #18, opened 2017-06-23
Soft delete instead of disconnect for containers models

Ladas:
Soft delete instead of disconnect for containers models

### DIFF
--- a/db/migrate/20170704102536_use_deleted_in_containers_tables.rb
+++ b/db/migrate/20170704102536_use_deleted_in_containers_tables.rb
@@ -1,0 +1,55 @@
+class UseDeletedInContainersTables < ActiveRecord::Migration[5.0]
+  class ContainerDefinition < ActiveRecord::Base;
+  end
+
+  class ContainerGroup < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class ContainerImage < ActiveRecord::Base;
+  end
+
+  class ContainerProject < ActiveRecord::Base;
+  end
+
+  class ContainerNode < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Container < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def disconnect_to_soft_delete(model)
+    model.where.not(:deleted_on => nil).update_all("ems_id = old_ems_id")
+  end
+
+  def soft_delete_to_disconnect(model)
+    model.where.not(:deleted_on => nil).update_all(:ems_id => nil)
+  end
+
+  MODEL_CLASSES = [
+    ContainerDefinition,
+    ContainerGroup,
+    ContainerImage,
+    ContainerProject,
+    ContainerNode,
+    Container,
+  ].freeze
+
+  def up
+    MODEL_CLASSES.each do |model_class|
+      say_with_time("Change ':deleted_on not nil' :ems_id to :old_ems_id for #{model_class}") do
+        disconnect_to_soft_delete(model_class)
+      end
+    end
+  end
+
+  def down
+    MODEL_CLASSES.each do |model_class|
+      say_with_time("Change ':deleted_on not nil' :ems_id to nil for #{model_class}") do
+        soft_delete_to_disconnect(model_class)
+      end
+    end
+  end
+end
--- a/db/migrate/20170704102536_use_deleted_in_containers_tables.rb
+++ b/db/migrate/20170704102536_use_deleted_in_containers_tables.rb
@@ -1,15 +1,15 @@
 class UseDeletedInContainersTables < ActiveRecord::Migration[5.0]
-  class ContainerDefinition < ActiveRecord::Base;
+  class ContainerDefinition < ActiveRecord::Base
   end
 
   class ContainerGroup < ActiveRecord::Base
     self.inheritance_column = :_type_disabled
   end
 
-  class ContainerImage < ActiveRecord::Base;
+  class ContainerImage < ActiveRecord::Base
   end
 
-  class ContainerProject < ActiveRecord::Base;
+  class ContainerProject < ActiveRecord::Base
   end
 
   class ContainerNode < ActiveRecord::Base
--- a/db/migrate/20170704102506_add_deleted_on_indexes_to_containers_tables.rb
+++ b/db/migrate/20170704102506_add_deleted_on_indexes_to_containers_tables.rb
@@ -1,4 +1,4 @@
-class AddDeletedToContainersTables < ActiveRecord::Migration[5.0]
+class AddDeletedOnIndexesToContainersTables < ActiveRecord::Migration[5.0]
   def change
     add_index :container_definitions, :deleted_on
     add_index :container_groups, :deleted_on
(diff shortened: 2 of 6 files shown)


Ladas:
@miq-bot assign @Fryguy 

cc @zeari @cben @agrare @simon3z

db/migrate/20170530102506_add_deleted_to_containers_tables.rb
```diff
@@ -0,0 +1,14 @@
+class AddDeletedToContainersTables < ActiveRecord::Migration[5.0]
+  def change
+    add_index :container_definitions, :deleted_on,
+              :name  => "index_container_definitions_on_deleted_on"
```
Fryguy: Can you remove the name and let Rails use the default name?

db/migrate/20170530102506_add_deleted_to_containers_tables.rb
```diff
@@ -0,0 +1,14 @@
...
+    add_index :container_groups, :deleted_on,
+              :name  => "container_groups_on_deleted_on"
```
Fryguy: Does this one not have the "index" prefix intentionally?

db/migrate/20170530102536_use_deleted_in_containers_tables.rb
```diff
@@ -0,0 +1,67 @@
...
+      disconnect_to_soft_delete(ContainerImage)
+    end
+
+    say_with_time("Change ':deleted_on not nil' :ems_id to :old_ems_id for ContainerProject") do
+      disconnect_to_soft_delete(ContainerProject)
+    end
+
+    say_with_time("Change ':deleted_on not nil' :ems_id to :old_ems_id for Container") do
+      disconnect_to_soft_delete(Container)
+    end
```
Fryguy: Minor, but you can massively DRY this up with a loop on the models

Fryguy:
Code-wise, looks good to me.  I want a :+1: from @simon3z before merging, though.

Fryguy:
@Ladas Would it make sense to squash commits here?  For example, you have commits in the middle about partial indexes, but you don't actually have any partial indexes in the end.

simon3z:
> Code-wise, looks good to me. I want a  from @simon3z before merging, though.

I was waiting for @cben to take a look at #15250 but probably shouldn't block on that.

+1 for letting Rails use the default name for the indexes.

Overall LGTM :+1:

spec/migrations/20170530102536_use_deleted_in_containers_tables_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    expect(model.where.not(:deleted_on => nil).count).to eq 3
+    expect(model.where(:deleted_on => nil).count).to eq 2
+    expect(model.where(:ems_id => nil).count).to eq 3
+    expect(model.where.not(:ems_id => nil).count).to eq 2
+  end
+
+  def assert_after_migration_data_of(model)
+    expect(model.where.not(:deleted_on => nil).count).to eq 3
+    expect(model.where(:deleted_on => nil).count).to eq 2
+    expect(model.where(:ems_id => nil).count).to eq 1
```
cben: consider also testing `:ems_id => 20` and/or `:ems_id => 10` counts.

db/migrate/20170530102506_add_deleted_to_containers_tables.rb
Ladas: might be that the default name will pass the length limit now, will do

db/migrate/20170530102536_use_deleted_in_containers_tables.rb
Ladas: I know, I was too lazy. @lpichler have said the same thing :-D

Ladas:
@Fryguy @simon3z ok, will squash it as this seems the final version. Will default and dry it up. :-)

db/migrate/20170530102536_use_deleted_in_containers_tables.rb
Ladas: done

db/migrate/20170530102506_add_deleted_to_containers_tables.rb
Ladas: fixed by autonaming

db/migrate/20170530102506_add_deleted_to_containers_tables.rb
Ladas: done

Ladas:
@simon3z @Fryguy @cben done, done and done. :-D Added also the ContainerNode model.

db/migrate/20170704102536_use_deleted_in_containers_tables.rb
```diff
@@ -0,0 +1,55 @@
...
+  class ContainerNode < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Container < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def disconnect_to_soft_delete(model)
+    model.where.not(:deleted_on => nil).update_all("ems_id = old_ems_id")
```
cben: @zeari @zakiva @simon3z 
Is there any scenario where a record could have `ems_id` set and `old_ems_id == nil`, but `deleted_on` set?
Then this would make a bad situation worse, setting `ems_id = old_ems_id = nil`.

It "shouldn't happen", I just remember we saw `ems_id == old_ems_id == nil` in a customer's DB, and suspected it might have resulted from double disconnect overwriting old_ems_id, so trying to think along similar lines...

Same question for `soft_delete_to_disconnect`.

simon3z: > @zeari @zakiva @simon3z
> Is there any scenario where a record could have ems_id set and old_ems_id == nil, but deleted_on set?
> Then this would make a bad situation worse, setting ems_id = old_ems_id = nil.

@cben technically there shouldn't be but I suppose it depends on whether it was possible that a crash left unfinished the updates (if multiple ones and not atomic). Also you're never 100% sure of not having had bugs.

If there's a better logic that we could use to avoid possible mistakes then let's try to improve this (maybe we should use `model.where.not(:deleted_on => nil, :ems_id => nil).update_all("ems_id = old_ems_id")` to be on the safe side?).

Ladas: so doing this, before doing  update_all("ems_id = old_ems_id")
```
    model.where.not(:deleted_on => nil).where.not(:ems_id => nil).update_all(:deleted_on => nil)
```

so having :ems_id will mean the record is not disconnected

Ladas: I wonder if I should also clean up the case where :ems_id is nil but :deleted_on is also nil? Putting Time.now into deleted_on?

Ladas: For :deleted_on => nil, :ems_id => nil, I am filling deleted_on with a date

cben:
Sorry, I've been reviewing another migration today, and my brain is refusing to concentrate 🤕 

But yeah, looks excellent (thanks for separate commits).  In both directions, you make deleted_on consistent before you act based on it. :+1:

Ladas:
@Fryguy @simon3z this should be ready

PR #19, opened 2017-06-23
Add unique indexes for Containers tables

Ladas:
Add unique indexes for Containers tables


Ladas:
@miq-bot assign @Fryguy

Ladas:
cc @Fryguy @agrare @cben @simon3z what do you guys think is missing or should be different?

I've decided to index container_image as [:ems_id, :image_ref], since the image_ref has the host and port of registry inside

Can I remove the :protocol from [:container_service_id, :ems_ref, :protocol] now?

:container_definitions and :containers were nested before, can I trust the :ems_ref is unique under a Manager?

Why is :container_volume not using :ems_ref?

----------

then the last question is, if we are clear to delete a duplicate, or we need to reconnect relations first. I assume we will mostly see duplicated disconnected ContainerImages

Fryguy:
How do you deal with archived / soft-deleted if you are indexing on ems_id ems_ref?  That is, if you have multiple providers and they both archive something with the same ems_ref, you get an error.

kbrock:
@Fryguy I think that use case will still be good. we will not be setting `ems_id` to `nil`.

Fryguy:
@kbrock I wonder if ems_refs ever get reused in a provider (or in a future provider).  That is, if you delete an object, does the provider then have the right to reuse the id number?  If they do, then this would break this idea.

cben:
> 1. I've decided to index container_image as [:ems_id, :image_ref], since the image_ref has the host and port of registry inside

@enoodle please advise.  From https://github.com/ManageIQ/manageiq/pull/14185 it sounds like [:ems_id, :image_ref] is not good enough?

> 2. Can I remove the :protocol from [:container_service_id, :ems_ref, :protocol] now?

I don't think we did anything about https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/35 
parse_service_port_config [sets](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/0f73f292db1f/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L991) `:ems_ref => "#{service_id}_#{port_config.port}_#{port_config.targetPort}"` (where service_id is the service's ems_ref)
so I think you still need protocol for uniqueness ATM.

Perhaps since this is a synthetic ems_ref it's better not to rely on it but on the underlying fields you want eg. `[:port, :target_port, :protocol

> 3. :container_definitions and :containers were nested before, can I trust the :ems_ref is unique under a Manager?

> 4. Why is :container_volume not using :ems_ref?

cc @zakiva.  IIRC ContainerVolume and PersistentVolume share a table (STI) but only one sets ems_ref?

> 5. then the last question is, if we are clear to delete a duplicate, or we need to reconnect relations first. I assume we will mostly see duplicated disconnected ContainerImages

zakiva:
> Why is :container_volume not using :ems_ref?
> cc @zakiva. IIRC ContainerVolume and PersistentVolume share a table (STI) but only one sets ems_ref?

Correct, only Persistent Volumes use :ems_ref

enoodle:
> I've decided to index container_image as [:ems_id, :image_ref], since the image_ref has the host and port of registry inside
@enoodle please advise. From ManageIQ/manageiq#14185 it sounds like [:ems_id, :image_ref] is not good enough?

There is a possibility of two images from different registries that are the same. comparing with just image_ref will show them a different. The best identifier is the digest, this is why it is used if it exists.

cben:
> I wonder if ems_refs ever get reused in a provider (or in a future provider).

Kubernetes UIDs should be permanently unique [https://kubernetes.io/docs/concepts/overview/working-with-objects/names/]
But container refresh_parser synthesizes `ems_ref`s on several tables from other parts, and I'm afraid those might get reused. (it'd be jolly good if we called these something other than `ems_ref`...)

- [ ] services [might get](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/0f73f292db1f/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L532-L534) ems_ref `"#{new_result[:namespace]}_#{new_result[:name]}"`, sounds like can get reused!
  - [ ] this suggests ServicePortConfig ems_ref derived from it can get reused too.
- [x] pods use `metadata.uid` from k8s
  - [ ] some pod's children are derived from it, and sound unique *unless* you can change them via `kubectl edit pod`.

We've generally assumed pods are immutable once created.  Let's see.  Trying to oc edit one port's name I get:
```
# * spec: Forbidden: pod updates may not change fields other than `containers[*].image` or `spec.activeDeadlineSeconds`
```
which is good, but yes I can `oc edit` the image: field!  (Makes it soon go to ErrImagePull -> CrashLoopBackOff status, editing back makes it go Running again, and increments `restartCount`).

ContainerDefinition & Container ems_ref include the image, but they also include the immutable name.  I'm not sure off-hand if it's a problem.

Ladas:
@Fryguy @kbrock right now we count with fact that ems_id, ems_ref is permanently unique, when we reconnect it. The elements that can be reused should not be archivable. Every sane system should use permanently unique uuids though, otherwise the reporting will be a pain. :-)

@enoodle so right now, we just duplicated the image, even if the checksum is the same https://github.com/Ladas/manageiq/blob/6e2ed956d8ec84ec0131e9efce9587ad4edd3c85/app/models/ems_refresh/save_inventory_container.rb#L313-L313 . So I just see that the :container_image_registry_id is redundant here, because image ref already contains unique keys of the container_image_registry.

Or is there a case when we put digest into the image_ref? 

@cben cool, I'll fix the inline FIXMEs then. So I think we just need to agree on 5.

enoodle:
@Ladas We should have the digest inside the image_ref so it is ok as it is in my opinion.

cben:
[responding from gitter.]
> cben what is the story of kubernetes not using UUIDs for references? It's kind of a weird solution
> it's not accessible in API even?
> usually the providers have both UUIDs and human readable names

So an interesting point is that the names are a source of truth!  It's not just the API giving us incomplete information; mostly the API gives us exactly the "desired state" the user declared in etcd, which kubelets and other control loops in k8s are continuously converging to.
For example consider a route from `host: "foo.example.com"` to `kind: "Service", name: "foo"`.
Yesterday it pointed to Fred's service with `name: foo, uid: xxxxxxxx-...`.
Today he destroyed this service, and Barney created a new one with `name: foo, uid: yyyyyyyy-...`.
This makes requests arriving to that hostname go the new service, **with no changes to route object** !

* I believe such link changes would be picked up by full refresh (untested!) but may be challenging for targetted refresh.
  Not a problem for @simon3z's plan of full refresh with partial save by checksums.

@moolitayer and a few others raised the interesting idea that maybe then we shouldn't resolve links to uids or ids at all.  We could (1) store names and do explicit joins by name at access time (2) consider making names the actual primary & foreign key instead of ids, so relations work exactly same?!
Unfortunately, it's unclear whether this can be compatible with past reporting and archiving.
Suppose someone wanted "chargeback by hostname". `foo.example.com` should be charged to Fred yesterday and Barney today — but if name is the primary key, Barney's service actually overwrote Fred's in the same DB row :-(

Anyway this'd require a lot of thinking...  Presently we're keeping row create/update/delete by UID, with links name links resolved at refresh time.

db/migrate/20170714102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,91 @@
...
+    ComputerSystem             => [:managed_entity_id, :managed_entity_type],
+    ContainerEnvVar            => [:container_definition_id, :name, :value, :field_path],
+    ContainerLimitItem         => [:container_limit_id, :resource, :item_type],
+    ContainerPortConfig        => [:container_definition_id, :ems_ref],
+    ContainerQuotaItem         => [:container_quota_id, :resource],
+    ContainerServicePortConfig => [:container_service_id, :ems_ref, :protocol],
+    ContainerTemplateParameter => [:container_template_id, :name],
+    ContainerVolume            => [:parent_id, :parent_type, :name],
+    CustomAttribute            => [:resource_id, :resource_type, :name, :unique_name, :section, :source],
+    Hardware                   => [:vm_or_template_id, :host_id, :computer_system_id],
```
lpichler: There is no stub for `Hardware` and `OperatingSystem`. 🏭

Ladas: fixed

Ladas:
@Fryguy @cben talked with @agrare , it will be safer to merge these as a first thing for next release, leaving us a full release to modify the indexes in a case of bugs.

Ladas:
@cben @agrare reviving this PR, can you review? One last FIXME is https://github.com/ManageIQ/manageiq-schema/pull/19/files#diff-9f048b534014c0bc40f2e38460cc7bd5R65

then we should figure out the https://github.com/ManageIQ/manageiq/pull/16454

cben:
What about archived rows?  IIRC idea was to have unique index scoped to deleted_on IS NULL, but I don't see anything like that in the PR?

I recently added `container_quota_scopes` table (#111), missing here.  `[:container_quota_id, :scope]` should be unique.

We're going to abuse archiving in container_quota* tables for keeping modification history and not just deletions.  I don't think this changes anything for indexes, non-archived portion will still be unique.

db/migrate/20171101102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,90 @@
...
+  class CustomAttribute < ActiveRecord::Base; end
+  class Hardware < ActiveRecord::Base; end
+  class OperatingSystem < ActiveRecord::Base; end
+
+  class Container < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def duplicate_data_query_returning_min_id(model, unique_index_columns)
+    model.group(unique_index_columns).select("min(id)")
```
cben: perhaps `max(id)` is better, likely to be more up-to-date?
(shouldn't matter unless we screwed up, but why not)

Ladas: right, TBH I am not sure, might be that max is better. :-)

cben:
Doesn't provider delete destroy all the things?
If we nullify, who garbage-collects the ems_id=NULL records then?
@Ladas, can you link where we nullify exactly?

cc @zeari

Ladas:
@cben hm, ok I thought we do it also in Containers, but we don't .

We do it for cloud/infra https://github.com/Ladas/manageiq/blob/2b27e62949a27b4620c69197b29a6b17c7cede72/app/models/ext_management_system.rb#L41 (also for hosts). Not sure who cleans the records then, we either have purger for it, or we expect a manual cleanup.

So for containers, I suppose I can rely on :ems_id being there (I would feel better with Foreign Key and Not Null constraints that columns though :-))

db/migrate/20171101102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,91 @@
...
+  class CustomAttribute < ActiveRecord::Base; end
+  class Hardware < ActiveRecord::Base; end
+  class OperatingSystem < ActiveRecord::Base; end
+
+  class Container < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def duplicate_data_query_returning_min_id(model, unique_index_columns)
+    model.group(unique_index_columns).select("max(id)")
```
moolitayer: method name has min but sql function is max?

Ladas: right, it was changed in last commit, so I need to also rename the function :-)

Ladas:
Ok, so I will need to optimize the deduplication migration. I am testing now on table with bigger amount of data and it seems to be really slow.

@cben btw. there have been new tables from the time I've send this?

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
```diff
@@ -0,0 +1,82 @@
...
+              [:parent_id, :parent_type, :name],
+              :unique => true # FIXME(lsmola) has unused :ems_ref
+    add_index :custom_attributes,
+              [:resource_id, :resource_type, :name, :unique_name, :section, :source],
+              :unique => true,
+              :name   => "index_custom_attributes_parameters_unique_multi_column"
+    add_index :hardwares,
+              [:vm_or_template_id, :host_id, :computer_system_id],
+              :unique => true,
+              :name => "index_hardwares_on_unique_multi_column"
```
kbrock: rubocop wants indentation here.
you indented correctly everywhere else

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,94 @@
...
+  class CustomAttribute < ActiveRecord::Base; end
+  class Hardware < ActiveRecord::Base; end
+  class OperatingSystem < ActiveRecord::Base; end
+
+  class Container < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def cleanup_duplicate_data_delete_all(model, unique_index_columns)
+    # 'ORDER BY id DESC' causes we keep the latest occurrence, ordering ASC would keep the first.
```
kbrock: Think it could read:
```
# 'ORDER BY id DESC' keeps the highest `id`. `ORDER BY id ASC` (more common case) would keep the lowest `id`.
```

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,94 @@
...
+    connection.execute <<-SQL
+      DELETE FROM #{model.table_name}
+        WHERE id IN (SELECT id
+                      FROM (SELECT id, ROW_NUMBER() OVER (partition BY #{unique_index_columns.join(",")} ORDER BY id DESC) AS rnum
+                             FROM #{model.table_name}) t
+                      WHERE t.rnum > 1);
+    SQL
```
kbrock: nicely done. Window functions FTW

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
Ladas: done

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
Ladas: done

Ladas:
@Fryguy @kbrock @cben @lpichler 

Here are the benchmarks, I am just listing the dedup using the window functions, since the old using 'group by' took over 3h for 1 table with 600k rows (I've killed it after, so it could have take much longer)

Benchmark
=========

In each scenario, there is a same amount of rows in each table. The we try it as dedup where each row has 2 duplicates, followed by case with 0 duplicates. The real world case will most likely be somewhere in-between.

CleanUpDuplicatesInContainersTables benchmark
---------------------------------------------------------------------

| Number of rows | Number of uniq rows | Time of processing |
| ------------- | ------------- | ------------- |
|  17.4M(600k per table) | 5.8M(200k per table) | 59.6602s |
|  17.4M(600k per table) | 17.4M(600k per table) | 15.5230s |
|  104.4M(3.6M per table) | 34.8M(1.2M per table) | 814.9881s |
|  104.4M(3.6M per table) | 104.4M(3.6M per table) | 91.4085s |

AddUniqueIndexesToContainersTables benchmark
---------------------------------------------------------------------

| Number of rows | Time of processing |
| ------------- | ------------- | ------------- |
|  5.8M(200k per table) | 8.7078s |
|  17.4M(600k per table) | 19.8361s |
|  34.8M(1.2M per table) | 78.6069s |
|  104.4M(3.6M per table) | 93.7157s |

Each collection has 600k records with 2 duplicates per each (200k uniq), Total 17,4M of records, 5.8M uniq

Tools used
==========

I've used this simple script to generate the data, always starting with ```be rake evm:db:reset```
https://gist.github.com/Ladas/01ab89ae495b4de40c0ebd0dd0fbc9e4

Inserting 104.4M of rows took 3828.352305063s, so it creates around 27.4k records per second. Maybe you can leverage this elsewhere. :-) Right now it inserts just the columns needed for the key and timestamps, it can be easily tweaked to mock all columns, the way foreign keys point to records (with more columns, it will slowdown though)

Conclusion
==========

Benchmarks shows this is more than acceptable, since we can migrate 100 millions of rows under 15 minutes.

Ladas:
@Fryguy @kbrock @cben @lpichler 

Here are the benchmarks, I am just listing the dedup using the window functions, since the old using 'group by' took over 3h for 1 table with 600k rows (I've killed it after, so it could have take much longer)

Benchmark
=========

In each scenario, there is a same amount of rows in each table. The we try it as dedup where each row has 2 duplicates, followed by case with 0 duplicates. The real world case will most likely be somewhere in-between.

CleanUpDuplicatesInContainersTables benchmark
---------------------------------------------------------------------

| Number of rows | Number of uniq rows | Time of processing |
| ------------- | ------------- | ------------- |
|  17.4M(600k per table) | 5.8M(200k per table) | 59.6602s |
|  17.4M(600k per table) | 17.4M(600k per table) | 15.5230s |
|  104.4M(3.6M per table) | 34.8M(1.2M per table) | 814.9881s |
|  104.4M(3.6M per table) | 104.4M(3.6M per table) | 91.4085s |

AddUniqueIndexesToContainersTables benchmark
---------------------------------------------------------------------

| Number of rows | Time of processing |
| ------------- | ------------- | ------------- |
|  5.8M(200k per table) | 8.7078s |
|  17.4M(600k per table) | 19.8361s |
|  34.8M(1.2M per table) | 78.6069s |
|  104.4M(3.6M per table) | 93.7157s |

Each collection has 600k records with 2 duplicates per each (200k uniq), Total 17,4M of records, 5.8M uniq

Tools used
==========

I've used this simple script to generate the data, always starting with ```be rake evm:db:reset```
https://gist.github.com/Ladas/01ab89ae495b4de40c0ebd0dd0fbc9e4

Inserting 104.4M of rows took 3828.352305063s, so it creates around 27.4k records per second. Maybe you can leverage this elsewhere. :-) Right now it inserts just the columns needed for the key and timestamps, it can be easily tweaked to mock all columns, the way foreign keys point to records (with more columns, it will slowdown though)

Conclusion
==========

Benchmarks shows this is more than acceptable, since we can migrate 100 millions of rows under 15 minutes.

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -71,15 +72,16 @@ def cleanup_duplicate_data_delete_all(model, unique_index_columns)
...
     ContainerEnvVar            => %i(container_id name value field_path),
     ContainerLimitItem         => %i(container_limit_id resource item_type),
     ContainerPortConfig        => %i(container_id ems_ref),
-    ContainerQuotaScope        => %i(container_quota_id scope),
     ContainerQuotaItem         => %i(container_quota_id resource),
+    ContainerQuotaScope        => %i(container_quota_id scope),
     ContainerServicePortConfig => %i(container_service_id name),
     ContainerTemplateParameter => %i(container_template_id name),
-    ContainerVolume            => %i(parent_id parent_type name),
+    ContainerVolume            => %i(parent_id parent_type ems_ref name),
```
Ladas: @cben We use :ems_ref for PersistentVolume and :name for ContainerVolume, is that wanted?

Ladas:
@Fryguy gah, core says %i-literals should be delimited by ( and )., but here we want [ ] ?

cben:
- Should uniqueness also hold after replication to master region?
  Everything in here has *some* id as part of index, so I think it can't collide. :heavy_check_mark: 

- I think some of these ems_ref are not actual kubernetes ems_refs but composites synthesized in refresh_parser.  I'd like to adopt some uniform approach.
  - One direction is get rid of synthesized ems_ref; index on the individual columns it consisted at.
    I like this because One Obvious Way To Do It, no need to think twice per table.
  - Another is add synthesized ems_ref where it's missing.
    This has benefit if it turns out we forgot a field for uniqueness (eg. https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/35), we'll be able to add it in parser without requiring migration (?).
    - But can we please not call these "ems_ref" but something else like "unique_ref"...

> Fryguy gah, core says %i-literals should be delimited by ( and )., but here we want [ ] ?

Strange, I just used `%i(...)` in https://github.com/ManageIQ/manageiq-schema/pull/151/files and it didn't complain.
IMNSHO `%i[...]`, `%w[...]` way clearer than `%i(...)`, `%w(...)` — they return an array!  I wish we could allow [ ] in core too...

Ladas:
@cben 
 - you got me with the replication. But I think you speak about the old style of replication we were doing x years ago? Now we just deploy Postgre in HA and the replication is handled on the DB side (which can handle constraints well)
- right, I do not prefer one or the other (synthesized :ems_ref, or multiple columns). Both have benefits and drawbacks. And also there are always entities with 1 UUID on the provider side, so those models would always have unique_ref(ems_ref). We should have a discussion about these, I'll put it on providers team list.
- I had to hack rubocop rules, so rubocop -a returns %i() :-) , but yeah I would expect the rubocop rules to be the same

Ladas:
@cben ping, could you do a final review of this, of what needs to change? We are getting into a time where it would be good to merge this and move forward.

Ladas:
@cben @agrare @Fryguy @kbrock also, we have few indexes that can have NULL values inside. In that case postgre doesn't take those as equal.

e.g. with index on :container_image_registries, that is on [:ems_id, :host, :port], we can call ```ContainerImageRegistry.create!(:ems_id => 10, :host => "host", :port => nil)``` multiple times and it will create the duplicates

Solutions:

1)  We can transform the value in index definition as e.g. ```COALESCE("port", 0)```, but the index might not be used for some queries (I need to test that)

2) We can define several indexes, based on what columns are nullable. So in this case we would have a second unique index for [:ems_id, :host] with ```WHERE port IS NULL``` . This will be used for queries, but then for some tables (:custom_attributes, :hardwares, :operating_systems, etc.) we would need several indexes like this defined.

-----

For both, we should start to define NOT NULL constraints for columns that are not nullable.

@carbonin could you also advise  ^ ?

cben:
reviewing.  @zeari please review too.

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
```diff
@@ -0,0 +1,97 @@
...
+    # Nested tables, not having :ems_id and the foreign_key is a part of the unique index
+    ContainerCondition         => %i(container_entity_id container_entity_type name),
+    SecurityContext            => %i(resource_id resource_type),
+    Tagging                    => %i(taggable_id taggable_type tag_id),
+    ComputerSystem             => %i(managed_entity_id managed_entity_type),
+    ContainerEnvVar            => %i(container_id name value field_path),
+    ContainerLimitItem         => %i(container_limit_id resource item_type),
+    ContainerPortConfig        => %i(container_id ems_ref),
+    ContainerQuotaItem         => %i(container_quota_id resource),
+    ContainerQuotaScope        => %i(container_quota_id scope),
```
zeari: @cben I trust this is correct in correlation with the quota changes...

zeari:
For both, we should start to define NOT NULL constraints for columns that are not nullable.

Im ok with something like this in another PR.


LGTM

cben:
Great find about NULLs!
There is also a 3rd solution — have refresh normalize these to something other than null.

I propose to treat this as a 2-step process:
1. we can merge these if we're sure it doesn't break existing behavior.
2. later, after we tighten them, figure out nullable etc., only then we'll start *relying on uniqueness*.

Ladas:
@cben right, I think we will have to do "have refresh normalize these to something other than null."

Then we can have several indexes per table (when different combination is unique for other provider). For the upsert strategy, I can specify 1 index per upsert then.

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
```diff
@@ -0,0 +1,87 @@
...
+    add_index :container_limit_items,
+              %i(container_limit_id resource item_type),
+              :unique => true,
+              :name   => "index_container_limit_items_unique_multi_column"
+    add_index :container_port_configs,
+              %i(container_id ems_ref),
+              :unique => true,
+              :name   => "index_container_port_configs_unique_multi_column"
+    add_index :container_quota_items,
+              %i(container_quota_id resource),
```
cben: We're going to start keeping history via archiving in this table, and unlike usual soft delete when the ems_ref or other unique id went away, these will collide on same container_quota_id resource.
- Should the index be limited to active (delete_on is null) items?
  But you didn't do this in any other table, and I think it's good to keep uniformity.
- Can you add `quota_desired quota_enforced quota_observed` here?

db/migrate/20180109102630_clean_up_duplicates_in_containers_tables.rb
cben: Thanks!
- Quotas are OK, gonna start regular soft delete, ems_ref is not reused.
- Quota scopes are fine, not going to archive.
- Quota items will collide! https://github.com/ManageIQ/manageiq-schema/pull/19/files#r167436691

cben:
To be honest, this got stuck in "I ought to give this some scrunity, but
don't know what to look for" state.  Sorry :-/
I did review this, no objections except quota.
But thanks to Ari's reminder, I'll try to re-review archiving vs uniqueness
in all archived tables.

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
Ladas: ok, adding ``` :where  => "deleted_on IS NULL",```, for quota items, then with the AR scope, the refresh should leave the archived items be

Ladas: hm, the unique index should then also be ``` [:resource, :quota_desired, :quota_enforced, :quota_observed]``` and the same for manager ref definition

Ladas: @cben is any of ```[:resource, :quota_desired, :quota_enforced, :quota_observed]``` nullable? (we should answer the question and add NOT NULL constraints really, to bring some order to the DB)

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
```diff
@@ -0,0 +1,122 @@
...
+              :name   => "index_hardwares_on_host_id"
+
+    remove_index :hardwares, :computer_system_id
+    add_index :hardwares,
+              %i(computer_system_id),
+              :where  => "vm_or_template_id IS NULL AND host_id IS NULL",
+              :unique => true,
+              :name   => "index_hardwares_on_computer_system_id_"
+
+    remove_index :operating_systems, :vm_or_template_id
```
Ladas: @cben so, this is how we will have to model unique indexes, when table are shared and have nullable parts
(we should redesign the DB though, having N:M mapping tables for vm_or_template_id, host_id and computer_system_id, then those tables can have correct NOT NULL constraints and unique indexes  )

Then for other cases, where index is several columns that are nullable, I think we will have to build the unique :ems_ref out of them

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
cben: quota_enforced & quota_observed come from the `status` part => I think currently, quickly after creating the quota, they might be null.  I need to add a test based on transient states recorded with a watch.

We can modify parser, though not immediately sure how.  Putting enforced=0 or observed=0 when this quota is not enforced yet is semantically wrong...  Perhaps we may simply discard quotas that have no status yet.

cben:
hmm, adding the quota values to unique index does NOT solve the same problem as adding them to refresh manager_ref.
- refresh only looks at active records and care when it *changes* — whether it should archive & create new one
- but over time, the value might be repeated! first 7, then 9, then 7 again — so adding value to uniqueness is useless!
  For uniqueness, should either include deleted_on — resource is unique (*) at given moment in time, or just exclude archived rows from uniqueness.

(*) I hope.  Need to validate it :)

db/migrate/20180109102659_add_unique_indexes_to_containers_tables.rb
Ladas: ok, if we can discard them, it would be ok

In the other case ,we would need several partial unique indexes. And would prepare smart batching, so the could would still work I think. But having set of NOT NULL unique indexes would be most effective

PR #21, opened 2017-06-25
subclassing ContainerImage to Openshift::ContainerImage

enoodle:
Adds type and ems_ref columns to ContainerImage and enables STI.
allow annotating only OpenshiftContainerImages
and remove annotation function from container_image

Schema changes split from https://github.com/ManageIQ/manageiq/pull/15386

### DIFF
--- a/db/migrate/20170613144708_add_sti_to_container_image.rb
+++ b/db/migrate/20170613144708_add_sti_to_container_image.rb
@@ -1,20 +1,6 @@
 class AddStiToContainerImage < ActiveRecord::Migration[5.0]
-  class ContainerImage < ActiveRecord::Base
-    self.inheritance_column = :_type_disabled # disable STI
-  end
-
-
-  def up
+  def change
     add_column :container_images, :type, :string
     add_column :container_images, :ems_ref, :string
-    say_with_time("Updating type column for ContainerImage") do
-      ContainerImage.where(:size => nil).update_all(:type => "ContainerImage")
-      ContainerImage.where.not(:size => nil).update_all(:type => "ManageIQ::Providers::Openshift::ContainerManager::ContainerImage")
-    end
-  end
-
-  def down
-    remove_column :container_images, :type
-    remove_column :container_images, :ems_ref
   end
 end
--- a/db/migrate/20170613144708_add_sti_to_container_image.rb
+++ b/db/migrate/20170613144708_add_sti_to_container_image.rb
@@ -1,0 +1,20 @@
+class AddStiToContainerImage < ActiveRecord::Migration[5.0]
+  class ContainerImage < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+
+  def up
+    add_column :container_images, :type, :string
+    add_column :container_images, :ems_ref, :string
+    say_with_time("Updating type column for ContainerImage") do
+      ContainerImage.where(:size => nil).update_all(:type => "ContainerImage")
+      ContainerImage.where.not(:size => nil).update_all(:type => "ManageIQ::Providers::Openshift::ContainerManager::ContainerImage")
+    end
+  end
+
+  def down
+    remove_column :container_images, :type
+    remove_column :container_images, :ems_ref
+  end
+end
--- a/spec/migrations/20170613144708_add_sti_to_container_image_spec.rb
+++ b/spec/migrations/20170613144708_add_sti_to_container_image_spec.rb
@@ -1,0 +1,17 @@
+require_migration
+
+describe AddStiToContainerImage do
+  let(:container_image_stub) { migration_stub(:ContainerImage) }
+
+  migration_context :up do
+    it "setting type correctly" do
+      container_image = container_image_stub.create!
+      openshift_container_image = container_image_stub.create!(:size => 12_345)
+
+      migrate
+
+      expect(container_image.reload).to have_attributes(:type => "ContainerImage")
+      expect(openshift_container_image.reload).to have_attributes(:type => "ManageIQ::Providers::Openshift::ContainerManager::ContainerImage")
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


db/migrate/20170613144708_add_sti_to_container_image.rb
```diff
@@ -0,0 +1,20 @@
+class AddStiToContainerImage < ActiveRecord::Migration[5.0]
+  class ContainerImage < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    add_column :container_images, :type, :string
```
lpichler: Is it possible to create one migration for structural changes and one for data changes? (you can use `change` method in structural migration)

db/migrate/20170613144708_add_sti_to_container_image.rb
```diff
@@ -0,0 +1,20 @@
...
+    add_column :container_images, :ems_ref, :string
+
+    say_with_time("Adding type and ems_ref columns to ContainerImage and setting type") do
+      ContainerImage.update_all(:type => "ManageIQ::Providers::Openshift::ContainerManager::ContainerImage")
```
lpichler: maybe `where.not.size(:size => nil)` ? (second command is for `where.size(:size => nil)`)

db/migrate/20170613144708_add_sti_to_container_image.rb
```diff
@@ -0,0 +1,20 @@
+class AddStiToContainerImage < ActiveRecord::Migration[5.0]
+  class ContainerImage < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    add_column :container_images, :type, :string
+    add_column :container_images, :ems_ref, :string
+
+    say_with_time("Adding type and ems_ref columns to ContainerImage and setting type") do
```
lpichler: only 'setting type'

spec/migrations/20170613144708_add_sti_to_container_image_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_migration
+
+describe AddStiToContainerImage do
+  let(:container_image_stub) { migration_stub(:ContainerImage) }
+
+  migration_context :up do
+    it "up" do
```
lpichler: maybe 'is setting the type to ...'?

db/migrate/20170613144708_add_sti_to_container_image.rb
enoodle: I tried this split but it doesn't work, The `up` method is not called for some reason that is beyond me. Can you explain the benefits of that approach? How exactly should this be done?

lpichler: I would create the first migration for structural change:
```ruby
def change 
  add_column :container_images, :type, :string
end
```
and the other with your data migration.

the reason can up method was not called that your db is probably already migrated.
try rollback rake `db:rollback STEP=5` (5 means 5 migrations back ) 
Does this work for you ?

**Why separate?**
these are two [different things](https://www.quora.com/What-are-the-differences-between-DDL-and-DML). It can be also helpfull for debbuging problems.
You also create more accurate names of migration. something like `AddTypeToContainerImage` and  `SetTypeTo`...

thanks

enoodle:
ping @lpichler @Fryguy @kbrock

Fryguy:
@enoodle Note that if you are adding the ems_ref and then use that in EMS refreshes, it is likely that you will thrash the records.  This is because the next EMS refresh will come along, it won't recognize the previous ems_ref because they are nil, delete all of the images, and recreate them.  If anyone has tags or policies or anything tied to the previous records, they will be lost.  If the expected ems_ref is based on information already in the database, you can avoid this by creating another data migration that populates the ems_ref column.

cc @blomquisg @simon3z

enoodle:
@Fryguy @simon3z The `ems_ref` was not suppose to be something that we are already collecting, but the `uid` field from Openshift. Reading Json's comment I think we will use the digest as the `ems_ref` or remove it / leave it empty. Thoughts?

moolitayer:
@enoodle did you plan to use the new `ems_ref` in inventory refreshes as primary key?
we don't have an `ems_ref` for all of the images, right?

enoodle:
@moolitayer I thought we should limit the use of `ems_ref` as a key for Openshift::ContainerImage but now I understand it is not possible. I now think it was a mistake to add it.

simon3z:
> @moolitayer I thought we should limit the use of ems_ref as a key for Openshift::ContainerImage but now I understand it is not possible. I now think it was a mistake to add it.

@enoodle @moolitayer @Fryguy who from the CM team reviewed and acked this PR to have it merged?

I would also have expected (at least) a discussion on what parameter to use to identify the image (e.g. size), etc.

enoodle:
@simon3z This was first in ManageIQ/manageiq#15386 before the repository split where some things were discussed.
Currently without https://github.com/ManageIQ/manageiq-providers-openshift/pull/23 This migration will cause problems because the new class doesn't exist.

As I see it, We can either use something that we already know for `ems_ref` (The digest for example) or remove, which is what I think we should do, because I understand now that it is not needed. Either way will require another migration.

Fryguy:
@simon3z I must have misread @moolitayer's approval in https://github.com/ManageIQ/manageiq/pull/15386 as a general approval, and since this was split out from that, continued as if it was good.  I was also going through dozens of PR reviews at the time, so I guess I just made a mistake.  If we need to make changes to remove/revert, we just need a new PR.

simon3z:
> @simon3z I must have misread @moolitayer's approval in ManageIQ/manageiq#15386 as a general approval, and since this was split out from that, continued as if it was good. I was also going through dozens of PR reviews at the time, so I guess I just made a mistake. If we need to make changes to remove/revert, we just need a new PR.

Ah sorry @Fryguy I haven't followed the approval to https://github.com/ManageIQ/manageiq/pull/15386
OK

PR #23, opened 2017-06-25
Add options to ems

enoodle:
This is migration of https://github.com/ManageIQ/manageiq/pull/15398

### DIFF
--- a/db/migrate/20170619161514_add_options_to_ext_management_system.rb
+++ b/db/migrate/20170619161514_add_options_to_ext_management_system.rb
@@ -1,0 +1,53 @@
+class AddOptionsToExtManagementSystem < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :custom_attributes,
+             :as         => :resource,
+             :dependent  => :destroy,
+             :class_name => 'AddOptionsToExtManagementSystem::CustomAttribute'
+    serialize :options
+  end
+
+  class CustomAttribute < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    belongs_to :resource, :polymorphic => true
+    serialize :serialized_value
+  end
+
+  def up
+    add_column :ext_management_systems, :options, :text
+    say_with_time("Migrating Kubernetes provider options") do
+      ExtManagementSystem
+        .where(:type => ["ManageIQ::Providers::Kubernetes::ContainerManager",
+                         "ManageIQ::Providers::Openshift::ContainerManager"])
+        .each do |cp|
+        migrated_options = Hash[
+          cp.custom_attributes.where(:section => 'cluster_settings',
+                                     :name    => %w(no_proxy http_proxy https_proxy))
+            .map { |ca| [ca.name.to_sym, ca.value] }]
+        cp.update(:options => {:image_inspector_options => migrated_options})
+      end
+      CustomAttribute.where(:section => 'cluster_settings',
+                            :name    => %w(no_proxy http_proxy https_proxy)).delete_all
+    end
+  end
+
+  def down
+    say_with_time("Migrating Kubernetes provider options") do
+      ExtManagementSystem
+        .where(:type => ["ManageIQ::Providers::Kubernetes::ContainerManager",
+                         "ManageIQ::Providers::Openshift::ContainerManager"])
+        .each do |cp|
+        options = cp.options
+        %w(no_proxy http_proxy https_proxy).each do |opt|
+          if options[:image_inspector_options].keys.include?(opt.to_sym)
+            cp.custom_attributes.create(:section => 'cluster_settings',
+                                        :name    => opt,
+                                        :value   => options[:image_inspector_options][opt.to_sym])
+          end
+        end
+      end
+    end
+    remove_column :ext_management_systems, :options
+  end
+end
--- a/spec/migrations/20170619161514_add_options_to_ext_management_system_spec.rb
+++ b/spec/migrations/20170619161514_add_options_to_ext_management_system_spec.rb
@@ -1,0 +1,37 @@
+require_migration
+
+describe AddOptionsToExtManagementSystem do
+  let(:ems_stub)   { migration_stub(:ExtManagementSystem) }
+  let(:ca_stub)    { migration_stub(:CustomAttribute) }
+  let(:proxy_name) { "my_proxy" }
+
+  migration_context :up do
+    it "migrates proxy options form CustomAttribute" do
+      kcp = ems_stub.create!(:type => "ManageIQ::Providers::Openshift::ContainerManager")
+      kcp.custom_attributes.create!(:name    => "http_proxy",
+                                    :section => "cluster_settings",
+                                    :value   => proxy_name)
+
+      migrate
+
+      expect(kcp.reload).to have_attributes(
+        :options => {:image_inspector_options => {:http_proxy => proxy_name}}
+      )
+      expect(ca_stub.where(:section => "cluster_settings").count).to eq(0)
+    end
+  end
+
+  migration_context :down do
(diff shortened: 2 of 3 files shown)


enoodle:
@simon3z @moolitayer @Fryguy Please take a look

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+  class CustomAttribute < ActiveRecord::Base
+    serialize :serialized_value
+  end
+
+  def up
+    add_column :ext_management_systems, :options, :text
+
+    say_with_time("Migrating Kubernetes provider options") do
+      ExtManagementSystem.where(:type => ["ManageIQ::Providers::Kubernetes::ContainerManager",
```
lpichler: I think we can remove N+1 and do something like:
```ruby
ExtManagementSystem.where(...).includes(:custom_attributes).where(:custom_attributes => {:section=> 'cluster_settings',... }.each do |ems|
options = ems.custom_attributes.map { |ca| Hash[*["image_inspector_#{ca.name}", ca.value]] }
ems.update(:options => options)
end
``` 
Does it cover your intention?

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -0,0 +1,34 @@
...
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    serialize :options
+  end
+  class CustomAttribute < ActiveRecord::Base
+    serialize :serialized_value
+  end
+
+  def up
+    add_column :ext_management_systems, :options, :text
```
moolitayer: `JSON` ?

https://www.postgresql.org/docs/9.4/static/datatype-json.html

enoodle: We are serializing this field. Does json work well with that? We are trying to mimic https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20170206210552_add_options_column_to_authentication.rb

moolitayer: I get that, I think the code should work the same except for dropping the serialize and some other code in the tests.
That way we get native JSON support (for example query only the `http_proxy` out of options) in the db layer.

@durandom wdyt?

durandom: we have `jsonb` in our migrations, e.g. [showback](https://github.com/ManageIQ/manageiq-schema/blob/27582ecff95dfcb740a8e3d22d2e7c3e03ea46fe/db/migrate/20170519153350_add_showback_models.rb#L35) lately

maybe look at their ActiveRecord model about serializing. I have never looked at how we do it

Fryguy: Downside to JSON is that you don't get time objects.  Will you need those?

enoodle: Maybe we will want to have time based optoins? @simon3z

moolitayer: OK I understand that gap in data types. Since this one is meant to be small
(configuration) I think clear serialization is more important then efficiency.
So I change my mind :+1: for `:text`

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -19,7 +19,6 @@ def up
       ExtManagementSystem
         .where(:type => ["ManageIQ::Providers::Kubernetes::ContainerManager",
                          "ManageIQ::Providers::Openshift::ContainerManager"])
-        .includes(:custom_attributes).where(:custom_attributes => {:section => "cluster_settings"})
```
enoodle: I wonder why this line causes this error: `PG::UndefinedColumn: ERROR:  column ext_management_systems.security_protocol does not exist` , Can anyone advise? cc @lpichler

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -0,0 +1,37 @@
...
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :custom_attributes,
+             :as         => :resource,
+             :dependent  => :destroy,
+             :class_name => 'AddOptionsToExtManagementSystem::CustomAttribute'
+    serialize :options
+  end
+
+  class CustomAttribute < ActiveRecord::Base
```
moolitayer: We now put an inheritance_column in every mock class to make it future proof

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -0,0 +1,37 @@
...
+    say_with_time("Migrating Kubernetes provider options") do
+      ExtManagementSystem
+        .where(:type => ["ManageIQ::Providers::Kubernetes::ContainerManager",
+                         "ManageIQ::Providers::Openshift::ContainerManager"])
+        .each do |cp|
+        migrated_options = Hash[
+          cp.custom_attributes.where(:section => 'cluster_settings',
+                                     :name    => %w(no_proxy http_proxy https_proxy))
+            .map { |ca| ["image_inspector_#{ca.name}".to_sym, ca.value] }]
+        cp.update(:options => migrated_options)
```
moolitayer: the custom attribute names are cluster_settings/no_proxy  cluster_settings/http_proxy etc and the new one is image_inspector_no_proxy ?
I think that would be better expressed with nesting.

enoodle: :+1:

db/migrate/20170619161514_add_options_to_ext_management_system.rb
simon3z: > Maybe we will want to have time based optoins? @simon3z

+1 for the ability to have time objects

db/migrate/20170619161514_add_options_to_ext_management_system.rb
```diff
@@ -0,0 +1,38 @@
...
+            .map { |ca| [ca.name.to_sym, ca.value] }]
+        cp.update(:options => {:image_inspector_options => migrated_options})
+      end
+      CustomAttribute.where(:section => 'cluster_settings',
+                            :name    => %w(no_proxy http_proxy https_proxy)).delete_all
+    end
+  end
+
+  def down
+    remove_column :ext_management_systems, :options
```
moolitayer: `up` & `down` will remove the setting...
I'm not sure how stringent we need to be here

spec/migrations/20170619161514_add_options_to_ext_management_system_spec.rb
```diff
@@ -0,0 +1,23 @@
+require_migration
+
+describe AddOptionsToExtManagementSystem do
+  let(:ems_stub)   { migration_stub(:ExtManagementSystem) }
+  let(:ca_stub)    { migration_stub(:CustomAttribute) }
+  let(:proxy_name) { "my_proxy" }
+
+  migration_context :up do
+    it "up" do
+      kcp = ems_stub.create!(:type => "ManageIQ::Providers::Openshift::ContainerManager", :id => 1)
```
moolitayer: We should avoid ids in tests - that violates region constraints

db/migrate/20170619161514_add_options_to_ext_management_system.rb
enoodle: I have added `down` migration.

enoodle:
ping @lpichler @Fryguy @durandom PTAL

simon3z:
@miq-bot assign Fryguy

enoodle:
ping @Fryguy

moolitayer:
cc @Loicavenel

PR #24, opened 2017-06-25
migrations for merging container_definition and container

zeari:
part of https://github.com/ManageIQ/manageiq/pull/15393


moolitayer:
missing `db/schema.yml` changes

moolitayer:
Looks good, one thing i'm concerned about is iterating over all containers. That would amount to tens of thousands of queries in large environments. Take a look at:
https://github.com/ManageIQ/manageiq-schema/blob/64af9f368dc352686f2908dec6ca9ab4b66d44ec/db/migrate/20170115140217_update_ems_in_miq_alert_status.rb

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,159 @@
...
+    add_column :containers, :privileged,         :boolean
+    add_column :containers, :run_as_user,        :bigint
+    add_column :containers, :run_as_non_root,    :boolean
+    add_column :containers, :capabilities_add,   :string
+    add_column :containers, :capabilities_drop,  :string
+    add_column :containers, :command,            :text
+
+    say_with_time("Copying over columns from container_definition to container") do
+      Container.all.each do |container|
+        container_definition = ContainerDefinition.find(container.container_definition_id)
```
cben: can this ever fail?

zeari: I dont think so. in order to have containers at all the inventory refresh has to go through and save provider->pod->container_def->container.

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,159 @@
+class UnifyContainerDefinitionAndContainer < ActiveRecord::Migration[5.0]
+  class ContainerDefinition < ActiveRecord::Base
+    has_many :container_port_configs
```
lpichler: @zeari we need to add `class_name` for all relations:
`:class_name => 'UnifyContainerDefinitionAndContainer::ContainerPortConfig'`
because without it, it will not take models defined in migrations.

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,159 @@
...
+    self.inheritance_column = :_type_disabled
+  end
+
+  class SecurityContext < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class ContainerEnvVar < ActiveRecord::Base
+    belongs_to :container_definition
+    self.inheritance_column = :_type_disabled
```
lpichler: this in only for tables with type column and there is no ContainerEnvVar#type, same for ContainerDefinition, SecurityContext and ContainerPortConfig.

lpichler:
is it possible to separate structural and data migration to standalone migrations ?

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,159 @@
...
+    add_column :containers, :container_group_id, :bigint
+    add_column :containers, :privileged,         :boolean
+    add_column :containers, :run_as_user,        :bigint
+    add_column :containers, :run_as_non_root,    :boolean
+    add_column :containers, :capabilities_add,   :string
+    add_column :containers, :capabilities_drop,  :string
+    add_column :containers, :command,            :text
+
+    say_with_time("Copying over columns from container_definition to container") do
+      Container.all.each do |container|
```
lpichler: I think we can save some N+1 queries by this and use batch processing by with `find_each`:

```ruby
Container.includes(:container_definition => [:container_port_configs, :container_env_vars, :security_contexts]).find_each do |container|
  container.update!(  :image => container_definition.image, ...)
  container.container_definition.container_port_configs.update_all(...
  container.container_definition.container_env_vars.update_all(...
  container.container_definition.security_context.update
end
```

db/migrate/20170529142557_unify_container_definition_and_container.rb
moolitayer: We were asked to put this in all models to prevent problems if we ever add the column in the future (I understand that is the current policy)

zeari:
> is it possible to separate structural and data migration to standalone migrations ?

Im not sure that would turn out so nice. we need to do structural changes before and after copying the data

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,134 @@
...
+    say_with_time("Copying over columns from container_definition to container") do
+      %w(image image_pull_policy memory cpu_cores container_group_id privileged
+         run_as_user run_as_non_root capabilities_add capabilities_drop command).each do |column|
+        join_sql = definitions.project(definitions[column.to_sym])
+                              .where(definitions[:id].eq(containers[:container_definition_id])).to_sql
+        Container.update_all("#{column} = (#{join_sql})")
+      end
+    end
+
+    Container.includes(:container_definition => [:container_port_configs, :container_env_vars, :security_context]).find_each do |container|
```
kbrock: starting on the other side (e.g. `ContainerEnvVars`) would work better.
In general, removing `belongs_to`

our suggestion from gitter:

```sql
update security_contexts 
set resource_id = (select containers.id from containers,container_definitions 
  where containers.container_definition_id = container_definitions.id 
  and container_definition_id = security_contexts.resource_id 
  and resource_type = 'ContainerDefinition')
```

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,150 @@
...
+          :cpu_cores          => container.cpu_cores,
+          :container_group_id => container.container_group_id,
+          :privileged         => container.privileged,
+          :run_as_user        => container.run_as_user,
+          :run_as_non_root    => container.run_as_non_root,
+          :capabilities_add   => container.capabilities_add,
+          :capabilities_drop  => container.capabilities_drop,
+          :command            => container.command
+        )
+        container.update!(:container_definition_id => container_def.id)
```
zeari: @kbrock what do you think is a good way to bulk insert into container_definitions? 
overall we need to do something like
 ```SQL
INSERT INTO container_definitions ( ems_id, name, image.... )
SELECT  ems_id, name, image....
FROM    containers
```

and then update `container_definition_id` in `containers`

Comment (zeari) [inline db/migrate/20170529142557_unify_container_definition_and_container.rb]: we might be able to do something like
```SQL
add_column :container_definitions, :container_id, :bigint
    

insert into container_definitions (container_id, columns...)
values (select (id, columns...) from containers)

update containers set container_definition_id = (select container_definition.id from container_definitions, containers)
                      where container.id = container_definitions.container_id


remove_column :container_definitions, :container_id
```



or maybe leverage `RETURNING` somehow - 
```SQL 
INSERT INTO container_definitions VALUES (....) RETURNING (id,....)
```

kbrock: @zeari Oh, I had thought that the `container_definitions` already existed for the `containers`.

The pure sql way is:

```sql
insert into container_Definitions (container_id, columns...)
  select id, columns..
  from conatiners
```

Not sure the MIQ way of doing that. we tend to be a little sql shy

simon3z:
@zeari @cben IIUC we need this PR to be reviewed/approved before we can move on with https://github.com/ManageIQ/manageiq/pull/15393 and https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/42

@cben @moolitayer @agrare can you review/approve?

moolitayer:
@zeari can we have tags or other relevant relations associated with a removed `container_definitions` ? 
(anything that has a generic resource relation to `container_definitions` )

zeari:
> @zeari can we have tags or other relevant relations associated with a removed container_definitions ?
> (anything that has a generic resource relation to container_definitions )

Looking at `container_definition.rb`, i cant say that we do. Metrics, events and such belong to container.

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+    add_column :containers, :command,            :text
+
+    containers = Arel::Table.new(:containers)
+    definitions = Arel::Table.new(:container_definitions)
+    say_with_time("Copying over columns from container_definition to container") do
+      %w(image image_pull_policy memory cpu_cores container_group_id privileged
+         run_as_user run_as_non_root capabilities_add capabilities_drop command).each do |column|
+        join_sql = definitions.project(definitions[column.to_sym])
+                              .where(definitions[:id].eq(containers[:container_definition_id])).to_sql
+        Container.update_all("#{column} = (#{join_sql})")
```
moolitayer: :+1:

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+      t.string     :capabilities_add
+      t.string     :capabilities_drop
+      t.text       :command
+    end
+
+    add_column :containers, :container_definition_id, :bigint
+
+    say_with_time("splitting columns from container into container_definition") do
+      ContainerDefinition.transaction do
+        Container.all.each do |container|
```
kbrock: This is going to be # containers x 2 queries

kbrock: maybe temporary column `ContainerDefinition#container_id` would make sense.

That way, this could be done in 2 queries.

zeari: sure, but this means we go with pure sql on this. (either that or something similar like https://gist.github.com/janko-m/2b2cea3e8e21d9232fb9 which is similar)

I did try to mitigate this with `ContainerDefinition.transaction` but it still means the DB goes through a lot of `INSERT INTO container_definitions VALUES (......)`. 

So pure sql is all I can think of doing here.

simon3z:
@cben @moolitayer this PR is missing your final review/approval.

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    say_with_time("switch container_definition_id with container_id for container_port_configs") do
+      env_vars = Arel::Table.new(:container_env_vars)
+      join_sql = containers.project(containers[:id])
+                           .where(containers[:container_definition_id].eq(env_vars[:container_definition_id])).to_sql
+      ContainerEnvVar.update_all("container_definition_id = (#{join_sql})")
+    end
+
+    say_with_time("switch container_definition_id with container_id for security_contexts") do
```
cben: swtich resource_id with container_id, resource_type to 'Container' for security_contexts

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+    rename_column :container_env_vars, :container_definition_id, :container_id
+
+    remove_column :containers, :container_definition_id
+    drop_table :container_definitions
+  end
+
+  def down
+    create_table :container_definitions do |t|
+      t.belongs_to :ems, :type => :bigint
+      t.string     :ems_ref
```
cben: definition ems_ref is lost in `up`.  Would copying it back FROM containers restore old situation?
parser synthesizes `"#{pod_id}_#{container_def.name}_#{container_def.image}"`,
for containers it synthesizes `"#{pod_id}_#{container.name}_#{container.image}"`.
- [x] pod_id is same `pod.metadata.uid` for both.
- [x] name should be same, it's what parser cross-ref'd containers to container_definitions.
- [ ] is image necessarily same?  cc @enoodle 

(I don't know if 100% reversible migration is a blocker here, but want to understands things first.)

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+      t.bigint     :old_ems_id
+      t.timestamp  :deleted_on
+      t.string     :name
```
cben: see above, copying name back should be OK

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+
+    remove_column :containers, :container_definition_id
+    drop_table :container_definitions
+  end
+
+  def down
+    create_table :container_definitions do |t|
+      t.belongs_to :ems, :type => :bigint
+      t.string     :ems_ref
+      t.bigint     :old_ems_id
```
cben: did containers and definitions always have same ems_id, old_ems_id?  were they archived in sync?
(trying to understand if copying back is safe)

db/migrate/20170529142557_unify_container_definition_and_container.rb
cben: Can you add comment that above columns were lost in `up` and are being reconstructed (by copying),
while following columns are *moved* back from containers table.

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+      t.string     :capabilities_add
+      t.string     :capabilities_drop
+      t.text       :command
+
+      t.bigint     :container_id # temp column
+    end
+
+    add_column :containers, :container_definition_id, :bigint
+
+    # populate container_definitions. use container_id to keep relation to containers
```
cben: consider turning this comment to a say_with_time
also for UPDATE below

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    say_with_time("switch container_definition_id with container_id for container_port_configs") do
+      env_vars = Arel::Table.new(:container_env_vars)
+      join_sql = containers.project(containers[:container_definition_id])
+                           .where(containers[:id].eq(env_vars[:container_id])).to_sql
+      ContainerEnvVar.update_all("container_id = (#{join_sql})")
+    end
+
+    say_with_time("switch container_definition_id with container_id for security_contexts") do
```
cben: again, resource_id and resource_type

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+
+    containers = Arel::Table.new(:containers)
+    say_with_time("switch container_definition_id with container_id for container_port_configs") do
+      port_configs = Arel::Table.new(:container_port_configs)
+      join_sql = containers.project(containers[:container_definition_id])
+                           .where(containers[:id].eq(port_configs[:container_id])).to_sql
+      ContainerPortConfig.update_all("container_id = (#{join_sql})")
+    end
+
+    say_with_time("switch container_definition_id with container_id for container_port_configs") do
```
cben: for container_env_vars

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+    # use container_id to join tables and update container_definition_id in containers
+    update_statement = "UPDATE containers
+                        SET container_definition_id = (SELECT container_definitions.id
+                                                       FROM container_definitions
+                                                       WHERE containers.id = container_definitions.container_id)"
+
+    ActiveRecord::Base.connection.execute(update_statement)
+
+    # finally, remove the temp column
+    remove_column :container_definitions, :container_id
```
cben: idea: if you move this remove to the end, can you simplify the 3 "switch container_definition_id with container_id" below to not need a join?
personally I'm not so much concerned with performance as simplicity ;-)

if yes, perhaps it's worth creating a temp column even in `up`?

just an idea, feel free to keep as is.

spec/migrations/20170529142557_unify_container_definition_and_container_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      :command            => "command"
+    }
+  end
+
+  let(:container_hash) do
+    {
+      :name       => "mycontainer",
+      :ems_ref    => "123",
+      :ems_id     => 1,
+      :old_ems_id => 1,
```
cben: Please take realistic data from real DB.  (that's not how ems_ref looks, not sure ems_id == old_ems_id == 1 is possible etc...)
And since I have some worries about the overlapping columns like `ems_ref`, consider spelling out full data for both AND the expected merge, without assuming `definition_hash.merge(container_hash))`.

[have I mentioned migrations scare me, perhaps unreasonably?]

db/migrate/20170529142557_unify_container_definition_and_container.rb
enoodle: Yes, If I understand correctly the contianer definition gets its image thought the container.

db/migrate/20170529142557_unify_container_definition_and_container.rb
zeari: Theyre usually always archived in sync.  
```Ruby 
# from container_definition.rb
def disconnect_inv
   return if ems_id.nil?
   _log.info "Disconnecting Container definition [#{name}] id [#{id}]"
    self.container.try(:disconnect_inv)
    self.deleted_on = Time.now.utc
    save
  end
```

zeari:
@kbrock I made changes and i think this should be good. can you have another look?

@cben i fixed most of what you asked.

simon3z:
@cben please review/approve, thanks!

simon3z:
> LGTM but one last question: does this handle container_definition that doesn't have a container?
> IMHO worth adding a test for that.

Thanks @cben I have the feeling it's an interesting case to test, @zeari what do you think?

zeari:
> LGTM but one last question: does this handle container_definition that doesn't have a container?
> IMHO worth adding a test for that.

@simon3z @cben 
Added test. im ignoring container_definitions without containers because i dont think theyre meaningful:
* we dont show them anywhere ui/reports
* if they do get a container, we will catch it in the next refresh
* nothing important is attached to them - metrics\events

zeari:
@kbrock i think this is ready

db/migrate/20170529142557_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,169 @@
...
+
+    say_with_time("switch resource_id with container_id, resource_type to 'Container' for security_contexts") do
+      security_contexts = Arel::Table.new(:security_contexts)
+      join_sql = containers.project(containers[:id])
+                           .where(containers[:container_definition_id].eq(security_contexts[:resource_id])
+                           .and(security_contexts[:resource_type].eq(Arel::Nodes::Quoted.new('ContainerDefinition')))).to_sql
+      SecurityContext.where(:resource_type => 'ContainerDefinition').update_all("resource_type = 'Container', resource_id = (#{join_sql})")
+    end
+
+    MiqQueue.where(:method_name => "purge_timer", :class_name => 'ContainerDefinition').destroy_all
```
zeari: @cben @simon3z Destroying the purge jobs.

simon3z:
> Added test. im ignoring container_definitions without containers because i dont think theyre meaningful:

@zeari ok as long as we don't crash in case there is a `container_definitions` without `containers`.

moolitayer:
@zeari I think there are some relevant rubocop warnings

db/schema.yml
```diff
@@ -805,7 +787,7 @@ container_port_configs:
 - port
 - host_port
 - protocol
-- container_definition_id
+- container_id
```
Fryguy: @carbonin Will this cause a problem since it's in the middle of the pack?  I think it's ok, just want to verify.

carbonin: Yeah, as long as it was implemented as a rename (not a delete and add).

carbonin: "Yeah" as in it's okay.... just to clarify 👍

zeari:
ping @Fryguy @carbonin Can we merge this one and https://github.com/ManageIQ/manageiq/pull/15393 ?

moolitayer:
cc @Loicavenel

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,174 @@
+class UnifyContainerDefinitionAndContainer < ActiveRecord::Migration[5.0]
+  class ContainerDefinition < ActiveRecord::Base
+    has_many :container_port_configs, :class_name => 'UnifyContainerDefinitionAndContainer::ContainerPortConfig'
+    has_many :container_env_vars, :class_name => 'UnifyContainerDefinitionAndContainer::ContainerEnvVar'
+    has_one :security_context, :as => :resource
```
Fryguy: This doesn't look like it's being used, so can we remove it?  I'm asking because polymorphic types end up causing issues when accessed through has_many/belongs_to, but luckily you are using resource_type/resource_id directly in the migrations below.

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,174 @@
...
+    has_one :container, :class_name => 'UnifyContainerDefinitionAndContainer::Container'
+  end
+
+  class Container < ActiveRecord::Base
+    belongs_to :container_definition, :class_name => 'UnifyContainerDefinitionAndContainer::ContainerDefinition'
+  end
+
+  class SecurityContext < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    belongs_to :resource, :polymorphic => true
```
Fryguy: Same comment as https://github.com/ManageIQ/manageiq-schema/pull/24/files#r129430898 (this is the other half)

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,174 @@
...
+      t.boolean    :privileged
+      t.bigint     :run_as_user
+      t.boolean    :run_as_non_root
+      t.string     :capabilities_add
+      t.string     :capabilities_drop
+      t.text       :command
+      t.bigint     :container_id                # temp column
+    end
+
+    add_index :container_definitions, :deleted_on
```
Fryguy: You probably want to add the index *after* you've done the data migration, otherwise each insert needs to update the index individually making it slower overall.

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -0,0 +1,174 @@
...
+
+    add_column :containers, :container_definition_id, :bigint
+
+    say_with_time("populate container_definitions. use container_id to keep relation to containers") do
+      insert_statement = "INSERT INTO container_definitions (container_id, ems_id, ems_ref, old_ems_id, deleted_on, name, image,
```
Fryguy: Prefer this format for building SQL

```ruby
    insert_statement = <<-SQL
INSERT INTO container_defintions...
SELECT id, ...
FROM containers
SQL
```

The reason is that a) it keeps the SQL on it's own lines, and b) some IDEs will see Ruby <<-SQL block and actually color the SQL statement with SQL syntax coloring.

Fryguy:
LGTM, so I am going to merge as is to unstick the main PR.  @zeari Can you create a follow up PR that cleans up the various things I mentioned above?

PR #27, opened 2017-06-29
This attribute is no longer used.

kbrock:
To be honest I can't really find when it was ever used.
I was able to track down the individual commit that introduced it, but did not find code that used it

moved from https://github.com/ManageIQ/manageiq/pull/15424

### DIFF
--- a/db/migrate/20170621204151_drop_miq_queue_for_user.rb
+++ b/db/migrate/20170621204151_drop_miq_queue_for_user.rb
@@ -1,0 +1,11 @@
+class DropMiqQueueForUser < ActiveRecord::Migration[5.0]
+  def up
+    remove_column :miq_queue, :for_user
+    remove_column :miq_queue, :for_user_id
+  end
+
+  def down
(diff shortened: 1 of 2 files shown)


Fryguy:
@kbrock What is the commit that introduced it?

kbrock:
@Fryguy it was introduced but looking through the code, never found these columns added to the actual code

Never added `MiqQueue#for_user` and never passed in `:for_user` / `:for_user_id` into `MiqQueue`.

```
commit 225601789df27209da6ecec6ee78d1159510ad74
Author: Jason Frey <jfrey@redhat.com>
Date:   Fri Feb 25 19:51:54 2011 +0000

    Added for_user / for_user_id columns to miq_queue.
    BugzID:11419
    
    git-svn-id: http://miq-ubuntusub.manageiq.com/svn/svnrepos/Manageiq/trunk@26765 3c68ef56-dcc3-11dc-9475-a42b84ecc76f
```

PR #28, opened 2017-07-03
Rename applies_to_exp to visibility_expression

lpichler:
`applies_to_exp` is unused and we can use it for storing
MiqExpression which can say whether the button will be
displayed or not. 
I think that it is better-named regard to feature but I am open for other ideas how to name it.

## Links 
https://www.pivotaltracker.com/n/projects/1608513/stories/147780727
cc @gtanzillo 

@miq-bot assign @Fryguy 

@miq-bot  add_label gapridashvili

### DIFF
--- a/db/migrate/20170713083836_rename_applies_to_visibility_expression.rb
+++ b/db/migrate/20170713083836_rename_applies_to_visibility_expression.rb
@@ -1,0 +1,5 @@
+class RenameAppliesToVisibilityExpression < ActiveRecord::Migration[5.0]
+  def change
+    rename_column :custom_buttons, :applies_to_exp, :visibility_expression
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -985,7 +985,7 @@ custom_buttons:
 - guid
 - description
 - applies_to_class
-- applies_to_exp
+- visibility_expression
 - options
 - userid
 - wait_for_complete


db/migrate/20170703170753_rename_applies_to_exp_to_miq_expression_visible.rb
```diff
@@ -0,0 +1,5 @@
+class RenameAppliesToExpToMiqExpressionVisible < ActiveRecord::Migration[5.0]
+  def change
+    rename_column :custom_buttons, :applies_to_exp, :miq_expression_visible
```
Fryguy: `visible` in an adjective, so this reads funny to me.  I'm thinking `visibility_expression` or `visibility_miq_expression`?  Thoughts?  Similar discussion in #29

Fryguy: @imtayadeway Thoughts?

imtayadeway: As in the other PR, I think `visibility_<thing>`, where thing is one of the common words we use to designate an expression is best 👍

lpichler:
I used `visibility_expression` thanks @Fryguy and @imtayadeway  :)

Fryguy:
@lpichler Is there a corresponding PR on manageiq that uses this column that needs to be merged at the same time?

PR #34, opened 2017-07-07
Add stored_data column to showback_event model

aljesusg:
We need to add a stored_data column to save the data of the event when the charge changes. @sergio-ocon 
img_2017-07-07 18 06 58


sergio-ocon:
Takes into account two use cases:
- Rate change at the middle of the cycle
- Several pools that are part of the same invoice

I.e. (follow Amazon changing rates)

sergio-ocon:
@miq-bot assign @Fryguy

aljesusg:
Thanks @Fryguy

PR #38, opened 2017-07-09
revert the addition of ems_ref to container image

enoodle:
Following the discussion in https://github.com/ManageIQ/manageiq-schema/pull/21 we should revert the addition of `ems_ref` column to container images.

### DIFF
--- a/db/migrate/20170710080149_remove_ems_ref_from_container_image.rb
+++ b/db/migrate/20170710080149_remove_ems_ref_from_container_image.rb
@@ -1,0 +1,5 @@
+class RemoveEmsRefFromContainerImage < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :container_images, :ems_ref, :string
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -750,7 +750,6 @@ container_images:
 - old_ems_id
 - deleted_on
 - type
-- ems_ref
 container_label_tag_mappings:
 - id
 - labeled_resource_type


enoodle:
ping @Fryguy @simon3z @moolitayer

db/migrate/20170613144708_add_sti_to_container_image.rb
```diff
@@ -1,6 +1,5 @@
 class AddStiToContainerImage < ActiveRecord::Migration[5.0]
   def change
     add_column :container_images, :type, :string
-    add_column :container_images, :ems_ref, :string
```
moolitayer: Please add a new migration @enoodle. 
That way developers that already ran the migration will remove the new column

enoodle:
@moolitayer I created a new migration like you asked. PTAL

simon3z:
@enoodle @moolitayer now that's in, would it hurt to just keep it instead (not using it). Just in case we'll find later that we need it?

enoodle:
@simon3z If we leave it always empty it will be ok.

Fryguy:
>  If we leave it always empty it will be ok.

I'm not so sure about that.  ems_ref is expected to be unique.  If you don't fill it in with something, then all images will be the same reference which will likely confuse the save_inventory / graph_persistor.

moolitayer:
> I'm not so sure about that. ems_ref is expected to be unique. If you don't fill it in with something, then all images will be the same reference which will likely confuse the save_inventory / graph_persistor.

We are using it in save currently (and of course we should not)

enoodle:
So it seems like we want to merge this one to avoid problems in the future @simon3z @Fryguy

moolitayer:
> We are using it in save currently (and of course we should not)

Oh my, terrible scribal error, we are NOT using it, sorry @enoodle
What I meant to say is that we should be safe leaving it in AFAIK.

the save_inventory_container code: https://github.com/ManageIQ/manageiq/blob/8ad46ceb2e51ba4fc9a0ba9f8cf6c1c228da675e/app/models/ems_refresh/save_inventory_container.rb#L222

simon3z:
LGTM :+1: cc @Fryguy

PR #40, opened 2017-07-11
Fixed possible NumericValueOutOfRange error on hardwares.memory_mb column

yrudman:
Custom report was throwing error when trying to convert value from`memory_mb` to bytes

https://bugzilla.redhat.com/show_bug.cgi?id=1464154

@miq-bot add-label bug

\cc @gtanzillo

### DIFF
--- a/db/migrate/20170711210740_change_memory_mb_in_hardwares_table_to_bigint.rb
+++ b/db/migrate/20170711210740_change_memory_mb_in_hardwares_table_to_bigint.rb
@@ -1,0 +1,9 @@
+class ChangeMemoryMbInHardwaresTableToBigint < ActiveRecord::Migration[5.0]
+  def up
+    change_column :hardwares, :memory_mb, :bigint
+  end
+
+  def down
+    change_column :hardwares, :memory_mb, :integer
+  end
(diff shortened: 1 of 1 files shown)


yrudman:
@miq-bot add-label bug

yrudman:
there is alternative solution for reported bug in  https://github.com/ManageIQ/manageiq/pull/15554 - Cast virtual attribute `Hardware#ram_size_in_bytes` to bigint.

Fryguy:
This *feels* like the wrong fix for the problem.  While I don't mind changing this, the current columns holds the value in MB, so the max value it can hold is 2147483647 megabytes => 2.147483647 petabytes, which is way more memory that any computer system I've ever heard of.

I think your alternative solution handles it better.

Fryguy:
Closing in favor of  ManageIQ/manageiq#15554

yrudman:
error thrown when trying to execute report and show virtual column defined as ```memory_mb * 1.megabyte``` ( = ```memory_mb * 1048576```) and max value it can hold before out of range error  is 2147483647 / 1048576 = 2048 MB

PR #41, opened 2017-07-13
Add DialogFieldAssociation table

eclarizio:
This is in preparation for the targeted auto-refresh enhancement. We will need dialog fields to be associated to one another.

[Here is the corresponding main repo PR](https://github.com/ManageIQ/manageiq/pull/15566)

https://www.pivotaltracker.com/story/show/148013179

@gmcculloug I don't know if you have merge rights on this repo but I figured I'd tag you anyway.

### DIFF
--- a/db/migrate/20170713144355_create_dialog_field_associations_table.rb
+++ b/db/migrate/20170713144355_create_dialog_field_associations_table.rb
@@ -1,0 +1,8 @@
+class CreateDialogFieldAssociationsTable < ActiveRecord::Migration[5.0]
+  def change
+    create_table :dialog_field_associations do |t|
+      t.bigint :trigger_id
+      t.bigint :respond_id
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20170713144355_create_dialog_field_associations_table.rb
```diff
@@ -0,0 +1,8 @@
+class CreateDialogFieldAssociationsTable < ActiveRecord::Migration[5.0]
+  def change
+    create_table :dialog_field_associations, :id => false do |t|
+      t.integer "dialog_field_trigger_id", :null => false
+      t.integer "dialog_field_respond_id", :null => false
```
Fryguy: - Please use symbols for consistency.
- Please remove the :null => false.
- All tables must have an id for replication purposes, so remove the :id => false

eclarizio:
@Fryguy Fixed, thanks.

eclarizio:
@Fryguy I'm assuming the Rubocop issues are ignorable considering all the rest of the migrations are written similarly to how I wrote them, but please let me know if I need to fix those issues to get this merged. Thanks.

PR #42, opened 2017-07-17
Add Resource to OpenSCAP results

aufi:
Adding resource_id and resource_type fields to openscap_results table
to allow assigning OpenSCAPResult to OpenStack VMs and later also
other inventory.

Existing container_id association is not touched in this PR, but it could
be migrated to the polymorphic resource association in future.

### DIFF
--- a/db/migrate/20170717084208_add_resource_to_open_scap_result.rb
+++ b/db/migrate/20170717084208_add_resource_to_open_scap_result.rb
@@ -1,0 +1,6 @@
+class AddResourceToOpenScapResult < ActiveRecord::Migration[5.0]
+  def change
+    add_column :openscap_results, :resource_id,   :bigint
+    add_column :openscap_results, :resource_type, :string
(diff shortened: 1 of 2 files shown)


db/migrate/20170717084208_add_subject_to_open_scap_result.rb
```diff
@@ -0,0 +1,6 @@
+class AddSubjectToOpenScapResult < ActiveRecord::Migration[5.0]
+  def change
+    add_column :openscap_results, :subject_id,   :bigint
+    add_column :openscap_results, :subject_type, :string
```
Fryguy: If this is a polymorphic reference, prefer :resource_id, :resource_type for consistency with other models.

aufi: @Fryguy Thanks for comment and info behind, updated.

PR #44, opened 2017-07-19
Add numeric columns to Container Quota Items

zakiva:
Related PRs: 
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/69
https://github.com/ManageIQ/manageiq/pull/15639

### DIFF
--- a/db/migrate/20170827091406_change_container_quota_items_columns_to_float.rb
+++ b/db/migrate/20170827091406_change_container_quota_items_columns_to_float.rb
@@ -1,0 +1,64 @@
+class ChangeContainerQuotaItemsColumnsToFloat < ActiveRecord::Migration[5.0]
+  class ContainerQuotaItem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    add_column :container_quota_items, :quota_desired_float, :float
+    add_column :container_quota_items, :quota_enforced_float, :float
+    add_column :container_quota_items, :quota_observed_float, :float
+
+    ContainerQuotaItem.find_each do |item|
+      item.quota_desired_float = numeric_quantity(item.quota_desired)
+      item.quota_enforced_float = numeric_quantity(item.quota_enforced)
+      item.quota_observed_float = numeric_quantity(item.quota_observed)
+      item.save
+    end
+
+    remove_column :container_quota_items, :quota_desired
+    remove_column :container_quota_items, :quota_enforced
+    remove_column :container_quota_items, :quota_observed
+
+    rename_column :container_quota_items, :quota_desired_float, :quota_desired
+    rename_column :container_quota_items, :quota_enforced_float, :quota_enforced
+    rename_column :container_quota_items, :quota_observed_float, :quota_observed
+  end
+
+  def down
+    add_column :container_quota_items, :quota_desired_string, :string
+    add_column :container_quota_items, :quota_enforced_string, :string
+    add_column :container_quota_items, :quota_observed_string, :string
+
+    ContainerQuotaItem.find_each do |item|
+      item.quota_desired_string = string_quantity(item.quota_desired)
+      item.quota_enforced_string = string_quantity(item.quota_enforced)
+      item.quota_observed_string = string_quantity(item.quota_observed)
+      item.save
+    end
+
+    remove_column :container_quota_items, :quota_desired
+    remove_column :container_quota_items, :quota_enforced
+    remove_column :container_quota_items, :quota_observed
+
+    rename_column :container_quota_items, :quota_desired_string, :quota_desired
+    rename_column :container_quota_items, :quota_enforced_string, :quota_enforced
+    rename_column :container_quota_items, :quota_observed_string, :quota_observed
+  end
+
+  def numeric_quantity(resource) # parse a string with a suffix into a int\float
+    return nil if resource.nil?
+
+    begin
+      resource.iec_60027_2_to_i
+    rescue
+      resource.decimal_si_to_f
+    end
+  end
+
+  def string_quantity(resource)
+    return nil if resource.nil?
+
+    val = resource == resource.to_i ? resource.to_i : resource
+    val.to_s
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -802,10 +802,10 @@ container_projects:
 container_quota_items:
 - id
 - resource
+- container_quota_id
 - quota_desired
 - quota_enforced
 - quota_observed
-- container_quota_id
 container_quotas:
 - id
 - name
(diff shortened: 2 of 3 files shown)


zakiva:
@simon3z @zeari

zakiva:
@moolitayer Please review

zakiva:
@miq-bot rm_label wip

db/migrate/20170717140915_add_numeric_columns_to_container_quota_items.rb
```diff
@@ -0,0 +1,44 @@
+class AddNumericColumnsToContainerQuotaItems < ActiveRecord::Migration[5.0]
+  class ContainerQuotaItem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    add_column :container_quota_items, :quota_desired_float,  :float
+    add_column :container_quota_items, :quota_desired_int,    :bigint
```
Fryguy: Why have both an int *and* a float?

zeari: @Fryguy We are trying to persist the `quantity` type from kubernetes which is a string that represents some type of number https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L41.
We are trying to ease the overhead on reporting when aggregating or averaging these values by saving them directly as bigint\float. This way instead of parsing strings on the fly when generating reports we already have their numeric value. Since we dont know if the value will be a float or an int we add both columns and fill the appropriate one when parsing the string.
Do you think we can find a better trick here?

cc @simon3z

cben: Sorry, I'm unconvinced.
- I could see an argument for not having any floats, only bigints, to avoid any rounding.
  But since we allow floats in *some* cases, why not use only floats?
- These come from quotas ultimately set by kubernetes users, who can express same quantity as `2048000m` cpu cores or `2Ki` cores.  The first would parse as 2048.0 float, the second as 2048 int, but I don't see why keep the distinction to any code that will deal with these columns...
- AFAIK `m`illi is the only sub-integer unit in this use case, so if we really wanted no rounding we could store everything in 1000x scaled fixed point.  Ugly but better than 2 columns IMHO.

P.S. this may be out of scope, but I wish we stored units somewhere, and not just inferred them from `resource`.  does quota_enforced_int == 1000 mean 1kBytes ? 1kdollars ? 1ksievert ?

simon3z: > * These come from quotas ultimately set by kubernetes users, who can express same quantity as 2048000m cpu cores or 2Ki cores. The first would parse as 2048.0 float, the second as 2048 int, but I don't see why keep the distinction to any code that will deal with these columns...

It doesn't matter how to the user expresses them, of course Kubernetes is not keeping the user's string, it's transforming it in a `Quantity` which will be then returned with the most appropriate formatting.

> * AFAIK milli is the only sub-integer unit in this use case, so if we really wanted no rounding we could store everything in 1000x scaled fixed point. Ugly but better than 2 columns IMHO.

OK for using floats. I think that the rounding takes place somewhere around ~`2**50` (please verify this).
It would be nice to have this check as a test, e.g. store a large integer (something like `2**50`) and check that reading the resulting float is still `==` to the original integer.

Fryguy: > P.S. this may be out of scope, but I wish we stored units somewhere, and not just inferred them from resource

In most every other column where we care about this we've chosen a base unit, so there is no need for storing a unit.  For example, for disk sizes we stores bytes always, and if the code receives something else it just converts to bytes.  This is very important for presentation purposes, because they can be assured that whatever they get back from the database is the same unit, and then they can apply consistent formatting rules during presentation (e.g. showing as megabytes or whatever).  I recommend doing the same here.

db/migrate/20170827091406_change_container_quota_items_columns_to_float.rb
```diff
@@ -0,0 +1,36 @@
...
+      item.save
+    end
+
+    remove_column :container_quota_items, :quota_desired
+    remove_column :container_quota_items, :quota_enforced
+    remove_column :container_quota_items, :quota_observed
+
+    rename_column :container_quota_items, :quota_desired_float, :quota_desired
+    rename_column :container_quota_items, :quota_enforced_float, :quota_enforced
+    rename_column :container_quota_items, :quota_observed_float, :quota_observed
```
zakiva: I didn't find a way to rename the columns and convert the values without adding the new columns first and then removing the old ones.

zakiva:
@cben @simon3z @Fryguy Changed the migration to use only float columns, PTAL.

db/migrate/20170717140915_add_numeric_columns_to_container_quota_items.rb
cben: IIUC in this table we have mixed *quantities* which can't have one unit: a quota where `resource=="cpu"` is in cores, a quota where `resource=="memory"` (?) is in bytes...

Right, it'd be wasteful to actually *store* a unit in every row.  I just think it's a question the *model* should be able answer (based on the `resource`), rather than having this logic elsewhere.

db/migrate/20170827091406_change_container_quota_items_columns_to_float.rb
```diff
@@ -0,0 +1,36 @@
...
+  class ContainerQuotaItem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    add_column :container_quota_items, :quota_desired_float, :float
+    add_column :container_quota_items, :quota_enforced_float, :float
+    add_column :container_quota_items, :quota_observed_float, :float
+
+    ContainerQuotaItem.all.each do |item|
```
cben: can you use `find_each` to batch this instead of loading `all` at once?
otherwise, LGTM.

zakiva: done :+1:

Fryguy:
Should there be a down migration?

zakiva:
> Should there be a down migration?

Well, I would say that for the down direction we can just convert the float value directly to a string i.e. without the original string suffix (using `to_s` and maybe rounding for integers). @zeari What do you think?

zeari:
> Well, I would say that for the down direction we can just convert the float value directly to a string i.e. without the original string suffix (using to_s and maybe rounding for integers). @zeari What do you think?

Thats should be ok. The only place we use this info is in the summary pages for projects so nothing would break. It would look a little odd but still understandable.

zakiva:
@Fryguy Added the down direction, PTAL

Fryguy:
LGTM :+1:  Will wait on merge until once @simon3z gives the +1 on the dependent PRs.

zakiva:
> LGTM  Will wait on merge until once @simon3z gives the +1 on the dependent PRs.

@Fryguy We have @simon3z's approval on the related PRs, we should be good to go here too.

moolitayer:
@Fryguy @chessbyte ready for merge per https://github.com/ManageIQ/manageiq-schema/pull/44#issuecomment-327526788

zakiva:
ping @Fryguy

PR #47, opened 2017-08-01
Add expression to entitlements

imtayadeway:
In support of https://www.pivotaltracker.com/story/show/148794877

@miq-bot add-label enhancement, core
@miq-bot assign @gtanzillo


Fryguy:
I'm wondering if this should be named `filter_expression` to make it clearer what the intent of the expression is.  `expression` by itself seems too open ended.

PR #48, opened 2017-08-09
single sequence for all metrics sub tables

kbrock:
**Before:**
Every `metrics_**` table had a different sequence.
So inserting directly into a subtable would result in overlapping `id` values.
We currently insert into the base `metrics` table, so this is not a problem.

```
 Column |  Type  |                           Modifiers                            
--------+--------+---------------------------------------------------------------
 id     | bigint | not null default nextval('metric_rollups_01_id_seq'::regclass)
```

**After:**
Every`metrics` subtable uses the same sequence.
So inserting into any of the metrics tables will get a unique id.

```
 Column |  Type  |                           Modifiers                            
--------+--------+---------------------------------------------------------------
 id     | bigint | not null default nextval('metric_rollups_id_seq'::regclass)
```

Rollback does bring back to the previous value

```
 Column |  Type  |                           Modifiers                            
--------+--------+---------------------------------------------------------------
 id     | bigint | not null default nextval('metric_rollups_01_id_seq'::regclass)
```

This change allows us to insert directly into the subtable, which is 10x faster than inserting into the common parent table.

/cc @chrisarcand @Fryguy

### DIFF
--- a/db/migrate/20170809191203_metrics_sequences.rb
+++ b/db/migrate/20170809191203_metrics_sequences.rb
@@ -1,0 +1,29 @@
+class MetricsSequences < ActiveRecord::Migration[5.0]
+  include MigrationHelper
+  def up
+    change_sequences("metrics", 0..23, false)
+    change_sequences("metric_rollups", 1..12, false)
+  end
+
+  def down
+    change_sequences("metrics", 0..23, true)
+    change_sequences("metric_rollups", 1..12, true)
+  end
+
+  # @param table_name [String] base table name to inherit from
+  # @param range [Range] range of subtables
+  # @param unique [Boolean] true if each table gets a unique sequence,
+  #                         false if they use the base table's sequence
+  def change_sequences(table_name, range, unique = true)
+    range.each do |n|
+      s = subtable_name(table_name, n)
+      execute("CREATE SEQUENCE #{s}_id_seq") if unique
+      change_column_default(s, :id, -> { "nextval('#{unique ? s : table_name}_id_seq')" })
+      execute("DROP SEQUENCE #{s}_id_seq") unless unique
+    end
+  end
+
+  def subtable_name(inherit_from, index)
+    "#{inherit_from}_#{index.to_s.rjust(2, '0')}"
+  end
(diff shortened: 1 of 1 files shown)


chrisarcand:
Seems ok to me.

Fryguy:
Can you also drop all of the individual sequences? `metrics_##_id_seq` and `metrics_rollups_##_id_seq`

Also, since this PR doesn't change the insertion into the main table, I'm assuming that still works as well?

chessbyte:
@kbrock @Fryguy what is the status of this PR?

kbrock:
@chessbyte thanks. I had thought this was merged already - will rebase and push again

Fryguy:
@kbrock Can you address https://github.com/ManageIQ/manageiq-schema/pull/48#issuecomment-321369724 ?

kbrock:
ugh. thanks @Fryguy 

1. added dropping of the sequences. I went into the db and they no longer show up.
2. yes inserts into the primary table work as it did before

db/migrate/20170809191203_metrics_sequences.rb
```diff
@@ -0,0 +1,29 @@
+class MetricsSequences < ActiveRecord::Migration[5.0]
+  include MigrationHelper
+  def up
+    change_sequences("metrics", 0..23, false)
+    change_sequences("metric_rollups", 1..12, false)
+  end
+
+  def down
+    change_sequences("metrics", 0..23, true)
+    change_sequences("metric_rollups", 1..12, true)
```
Fryguy: I just realized that on the down migration I *think* you also have to update the nextval to the highest id in the table or you'll get duplicates.

kbrock: In the old code, we can not create a record in the old tables.
The sequences will overlap

Fryguy: Ahhhh I see.  The only other thing we might want to do is ensure we set the value to the "min value for the region" as opposed to 1.  If that happens automatically, then we are good.

PR #49, opened 2017-08-15
Add hash_expression to MiqAlert

zakiva:
Related PR: https://github.com/ManageIQ/manageiq/pull/15315
* Adding `hash_expression` column to `miq_alerts` table. 
* Renaming `expression` column to `miq_expression`.

### DIFF
--- a/db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
+++ b/db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
@@ -1,0 +1,30 @@
+class AddHashExpressionToMiqAlert < ActiveRecord::Migration[5.0]
+  class MiqAlert < ActiveRecord::Base
+  end
+
+  def up
+    add_column :miq_alerts, :hash_expression, :text
+
+    say_with_time('Add hash expression to MiqAlert for existing alerts') do
+      MiqAlert.where("expression NOT LIKE ?", "--- !ruby/object:MiqExpression%").each do |alert|
+        alert.hash_expression = alert.expression
+        alert.expression = nil
+        alert.save
+      end
+    end
+
+    rename_column :miq_alerts, :expression, :miq_expression
+  end
+
+  def down
+    say_with_time('Remove hash expression from MiqAlert for existing alerts') do
+      MiqAlert.where.not(:hash_expression => nil).each do |alert|
+        alert.miq_expression = alert.hash_expression
+        alert.save
+      end
+    end
+
+    remove_column :miq_alerts, :hash_expression
+    rename_column :miq_alerts, :miq_expression, :expression
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -4881,10 +4881,11 @@ miq_alerts:
 - updated_on
 - options
 - db
-- expression
+- miq_expression
 - responds_to_events
 - enabled
 - read_only
+- hash_expression
 miq_approvals:
 - id
 - description
--- a/spec/migrations/20170703094235_add_hash_expression_to_miq_alert_spec.rb
+++ b/spec/migrations/20170703094235_add_hash_expression_to_miq_alert_spec.rb
@@ -1,0 +1,69 @@
+require_migration
+
+describe AddHashExpressionToMiqAlert do
+  migration_context :up do
+    let(:miq_alert_stub) { migration_stub(:MiqAlert) }
+
+    it 'alert with hash expression has nil miq_expression and hash_expression is set' do
+      hash_expression = '---\n:mode: internal\n'
+      alert = miq_alert_stub.create!(:description => 'Test Alert', :expression => hash_expression)
+
+      migrate
+
+      alert.reload
+      expect(alert.hash_expression).to eq(hash_expression)
+      expect(alert.miq_expression).to be_nil
+      expect { alert.expression }.to raise_error(NoMethodError)
+    end
+
+    it 'alert with non hash expression has nil hash_expression and miq_expression is set' do
+      expression = '--- !ruby/object:MiqExpression
+      exp:
+        "=":
+          field: Vm-platform
+          value: windows
+      '
+      alert = miq_alert_stub.create!(:description => 'Test Alert', :expression => expression)
+
+      migrate
+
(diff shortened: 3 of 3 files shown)


zakiva:
@simon3z @moolitayer @imtayadeway Please review

Fryguy:
This PR looks good as is, but I want to wait on a coordinated merge with https://github.com/ManageIQ/manageiq/pull/15315, in case any changes are made there.

db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
```diff
@@ -0,0 +1,35 @@
+class AddHashExpressionToMiqAlert < ActiveRecord::Migration[5.0]
+  class MiqAlert < ActiveRecord::Base
+    serialize :expression
+    serialize :miq_expression
```
Fryguy: Actually I just realized this is a problem.  If the MiqExpression contains any object references, then those will be materialized on deserialization.  I'm not sure what's actually in an MiqExpression, but if it's a class name, and in the future that class no longer exists, then this migration will fail on db:migrate because it will be trying to load that actual class from app/models via autoload.

Fryguy: cc @imtayadeway @gtanzillo @jrafanie Keep me honest here.  I know in other migrations I've avoided deserializing MiqExpressions because of this exact reason.

db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
```diff
@@ -0,0 +1,35 @@
...
+    add_column :miq_alerts, :hash_expression, :text
+
+    say_with_time('Add hash expression to MiqAlert for existing alerts') do
+      MiqAlert.all.each do |alert|
+        next unless alert.expression.kind_of?(Hash)
+        alert.hash_expression = alert.expression
+        alert.expression = nil
+        alert.save
+      end
+    end
```
Fryguy: Reading this PR carefully, it seems you only care about when the type of the expression is a hash, so to avoid deserializing MiqExpression, you can use string comparisons on the up, and then just transfer the column normally.

- Remove the serialize lines from the MiqAlert stub
- Use string comparisons looking for MiqExpression types in the columns
  ```ruby
  say_with_time('Add hash expression to MiqAlert for existing alerts') do
    MiqAlert.all.each do |alert|
      next if alert.expression.start_with?("--- !ruby/object:MiqExpression")
      alert.hash_expression = alert.expression
      alert.expression = nil
      alert.save
    end
  end
  ```
- The down can actually stay the same.

---

This has the side effect of making the PR faster because you don't go through `deserialization -> type check -> serialize`, but instead just do `string check -> copy`

---

Taking this a step further, since you know the anti-signature, you can do the filtering at the query level with something like

```ruby
MiqAlert.where("expression NOT LIKE ?", "--- !ruby/object:MiqExpression%").each do |alert|
  alert.hash_expression = alert.expression
  alert.expression = nil
  alert.save
end
```

db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
gtanzillo: @Fryguy Afaik, there are not object references in `MiqExpression` instances. All of the operators are strings and all the operands are also strings or numerics. I also defer to @imtayadeway and @jrafanie to keep me honest. But, even so, would it be safer to treat the `expression` column as a string for the migration to avoid any deserialization altogether?

db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
zakiva: @Fryguy I made the changes, thanks!

db/migrate/20170703094235_add_hash_expression_to_miq_alert.rb
```diff
@@ -0,0 +1,30 @@
+class AddHashExpressionToMiqAlert < ActiveRecord::Migration[5.0]
+  class MiqAlert < ActiveRecord::Base
+  end
+
+  def up
+    add_column :miq_alerts, :hash_expression, :text
```
imtayadeway: Just wondering if this is the best name, because "expression" makes me think of `MiqExpression`, which are also serialized as a hash. Are they called "hash expressions" on the provider side?

Fryguy: Yeah, I wasn't sure about the name either, but since I really couldn't think of anything better, I passed on it.  Do you have any suggestions?

imtayadeway: I don't, which I know makes this comment not much help ;) But if I understand this correctly (I may not), I thought this "expression" was used in the API call to the provider, so would suggest keeping it uniform with whatever they call it?

zakiva: I couldn't come up with a better name either :\ @imtayadeway I'm not sure I understand what do you mean by the "provider side", we plan to use this field in API calls from an ansible module when creating new policies.

imtayadeway: @zakiva I probably got some details wrong then :) If it's only meaningful in the context of ansible when creating new policies....possible `policy_options` or `policy_conditions`? @Fryguy ?

Fryguy: ¯\\\_(ツ)_/¯  I'll let you two duke it out 😆

imtayadeway: @zakiva if neither of those seem appropriate I'm OK with leaving this as is 👍

moolitayer: I think this is something other then a policy_x , this is a related to a `miq_alert` not `miq_policy`?
:+1: for leaving as is

zakiva: Thanks, keeping it as is.

PR #51, opened 2017-08-16
Set ActiveRecord::Migrator.migrations_paths correctly

carbonin:
Not having this path was causing `ActiveRecord::Migrator.needs_migration?` to return false always.

We use this in the `evm:deployment_status` task in manageiq to determine if we should run the migrations in the container upgrade scenario.

Without this change upgrades in containers will not run.

@miq-bot assign @Fryguy 
@miq-bot add_label bug


carbonin:
@simaishi This is problem 2 from the build (#52 being the first issue :wink: )

PR #54, opened 2017-08-17
Move rake tasks

bdunne:
I don't see any reason to expose `:write_schema` to anything outside of this repo.  `:check_schema` I'm on the fence about.

@carbonin @Fryguy What do you guys think?

@mkanoor pointed out that the test failure message was incorrect.  It was exposing `app:write_schema`


Fryguy:
Oh right...on the ManageIQ side it uses the class itself and not the rake task 👍

PR #56, opened 2017-08-26
Remove duplicated instance_types (added by user) from settings_changes

jameswnl:
After user makes use of https://github.com/ManageIQ/manageiq-providers-amazon/pull/289 to add `instance_types`, in next releasing cycle, MIQ also add these into our code, these entries in `settings_changes` will become redundant. This migration is to remove them.

### DIFF
--- a/db/migrate/20170825140348_remove_amazon_custom_instance_types.rb
+++ b/db/migrate/20170825140348_remove_amazon_custom_instance_types.rb
@@ -1,0 +1,15 @@
+class RemoveAmazonCustomInstanceTypes < ActiveRecord::Migration[5.0]
+  class SettingsChange < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time('Remove /ems/ems_amazon/additional_instance_types/* defined also in instance_types.rb') do
+      SettingsChange.where('key LIKE ?', '/ems/ems_amazon/additional_instance_types/%/name').each do |s|
+        *_rest, instance_type, _name = s.key.split('/')
+        if InstanceTypes::AVAILABLE_TYPES.include?(instance_type)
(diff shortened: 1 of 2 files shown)


jameswnl:
@miq-bot add_labels wip, enhancement

jameswnl:
@miq-bot remove_label wip

Fryguy:
I don't understand...won't this happen constantly?  That is, after this migration goes and runs, and all is well, then they add a new instance_type, and some time after that it becomes part of the product...are you going to create *another* migration like this?

Why not just leave it as is...is it harming anything?

cc @blomquisg

blomquisg:
> Why not just leave it as is...is it harming anything?

Right.  *If* we need to dedup these, then it would probably have to be done on in the Scheduler and not via a one-off migration.

@jameswnl can you check on whether having a duplicated `additional_instance_type` will cause any problems.  If it doesn't, I say just leave it.  Then, we can look at a dedup Scheduler script later if it's needed.

Fryguy:
I guess it could create duplicates because it might merge in what the user has with the original.  But even so, it seems we should just `.uniq` it in the InstanceTypes code.

jameswnl:
Duplicates are already taken care in the `merge` operation at https://github.com/ManageIQ/manageiq-providers-amazon/pull/289/files#diff-6239e0700a502ffce9e787dd8034668aR1893 and so won't be a problem.

This PR is to address @gtanzillo and @bronaghs 's concern that these database records would be accumulated over time. 

(A separate discussion with @gtanzillo and @jrafanie was to consolidate these to a single place. e.g. the `setting_changes` table, but that's a separate topic)

jameswnl:
Since this is not a concern, I'm closing it.

PR #58, opened 2017-08-29
Migrate the roles that we live migrate at startup

Fryguy:
Depends on https://github.com/ManageIQ/manageiq/pull/15905 which removes the "live" migration at startup.

@jrafanie Please review.


jrafanie:
LGTM

PR #59, opened 2017-09-01
Use subqueries to avoid n + 1 in migration UnifyContainerDefinitionAndContainer

lpichler:
I just got euwe DB with these counts of objetcts used in migration:
```
Container.count => 19811
ContainerDefinition.count => 20540
SecurityContext.count => 20540
ContainerEnvVar.count => 394082
ContainerPortConfig.count => 5503
```
and this migration was running for 1 hour and it didn`t ends.
after this changes it is about few seconds.

it is beacause there is N+1 in progress thanks to such queries:
```ruby
     containers = Arel::Table.new(:containers)
      port_configs = Arel::Table.new(:container_port_configs)
      join_sql = containers.project(containers[:id])
                           .where(containers[:container_definition_id].eq(port_configs[:container_definition_id])).to_sql
      ContainerPortConfig.update_all("container_definition_id = (#{join_sql})")
```
in sql:
```SQL
UPDATE container_port_configs 
SET container_definition_id = 
  (SELECT "containers"."id" FROM "containers" 
   WHERE "containers"."container_definition_id" = "container_port_configs"."container_definition_id"
 )
```
so I am using postgres subqueries, which are performed one time in the query:
```SQL
UPDATE container_port_configs
SET container_definition_id = subquery.id
FROM (SELECT id, container_definition_id FROM containers) AS subquery
WHERE subquery.container_definition_id = container_port_configs.container_definition_id
```

All related queries are changed by this way in up direction. Some specs are missing in down direction so I am leaving it for other PR, I don't want to broke anything.

@miq-bot assign @Fryguy
please review @zeari 
cc  @gtanzillo @kbrock 

thanks @Ladas

### DIFF
--- a/db/migrate/20170718111834_unify_container_definition_and_container.rb
+++ b/db/migrate/20170718111834_unify_container_definition_and_container.rb
@@ -40,9 +40,8 @@ class UnifyContainerDefinitionAndContainer < ActiveRecord::Migration[5.0]
     add_column :containers, :capabilities_drop,  :string
     add_column :containers, :command,            :text
 
-    containers = Arel::Table.new(:containers)
-    definitions = Arel::Table.new(:container_definitions)
     db_connection = ActiveRecord::Base.connection
+
     say_with_time("Copying over columns from container_definition to container") do
       %w(image image_pull_policy memory cpu_cores container_group_id privileged
          run_as_user run_as_non_root capabilities_add capabilities_drop command).each do |column|
--- a/db/migrate/20170718111834_unify_container_definition_and_container.rb
+++ b/db/migrate/20170718111834_unify_container_definition_and_container.rb
@@ -40,60 +40,46 @@ class UnifyContainerDefinitionAndContainer < ActiveRecord::Migration[5.0]
     add_column :containers, :capabilities_drop,  :string
     add_column :containers, :command,            :text
 
-    db_connection = ActiveRecord::Base.connection
-
     say_with_time("Copying over columns from container_definition to container") do
       %w(image image_pull_policy memory cpu_cores container_group_id privileged
          run_as_user run_as_non_root capabilities_add capabilities_drop command).each do |column|
-        subquery = <<-SQL
-          SELECT id, #{column} FROM container_definitions
-        SQL
-
-        db_connection.execute <<-SQL
+        connection.execute <<-SQL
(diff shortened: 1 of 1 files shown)


db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -40,37 +40,54 @@ def up
     add_column :containers, :capabilities_drop,  :string
     add_column :containers, :command,            :text
 
-    containers = Arel::Table.new(:containers)
-    definitions = Arel::Table.new(:container_definitions)
+    db_connection = ActiveRecord::Base.connection
```
Fryguy: `connection` is a method directly available on migrations, so this should not be needed.

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -40,37 +40,54 @@ def up
...
+        db_connection.execute <<-SQL
+          UPDATE containers SET #{column} = subquery.#{column}
+          FROM (#{subquery}) AS subquery
+          WHERE subquery.id = containers.container_definition_id
+        SQL
       end
     end
 
+    subquery_containers = <<-SQL
+        SELECT id, container_definition_id FROM containers
```
Fryguy: 2-space indent

db/migrate/20170718111834_unify_container_definition_and_container.rb
```diff
@@ -40,37 +40,54 @@ def up
...
-        join_sql = definitions.project(definitions[column.to_sym])
-                              .where(definitions[:id].eq(containers[:container_definition_id])).to_sql
-        Container.update_all("#{column} = (#{join_sql})")
+        subquery = <<-SQL
+          SELECT id, #{column} FROM container_definitions
+        SQL
+
+        db_connection.execute <<-SQL
+          UPDATE containers SET #{column} = subquery.#{column}
+          FROM (#{subquery}) AS subquery
```
kbrock: Would it make sense to just put `subquery` inline here?

I'm not sure how we typically use raw SQL.

I guess it does remove a little duplication - which ever you feel is best.

Personally, I find the sql you have written to be easy to read and much more readable than anything arel would give us.

lpichler:
thanks for your comments, updated 👍

PR #60, opened 2017-09-04
Add container_project_id to persistent_volume_claims

zeari:
needed for https://github.com/ManageIQ/manageiq/pull/15932 so we can add a `belongs_to :container_projects` relation to  persistent_volume_claims

@zakiva Please review
cc @simon3z @cben @moolitayer


moolitayer:
@zeari we will need a migration for existing records right?

zeari:
> @zeari we will need a migration for existing records right?

the relationship should be filled in the next inventory refresh

PR #61, opened 2017-09-06
Add schema support for LXCA config patterns

skovic:
This PR adds schema support for Lenovo XClarity Administrator (LXCA) configuration patterns which can be thought of as templates describing the configuration of physical systems. These "templates" are used by LXCA to configure hardware such as servers, switches, and storage.

### DIFF
--- a/db/migrate/20170905210626_create_configuration_templates.rb
+++ b/db/migrate/20170905210626_create_configuration_templates.rb
@@ -2,7 +2,7 @@ class CreateConfigurationTemplates < ActiveRecord::Migration[5.0]
   def change
     create_table :configuration_templates do |t|
       t.bigint :ems_id
-      t.string :external_id
+      t.string :ems_ref
       t.string :name
       t.string :description
       t.boolean :user_defined
--- a/db/migrate/20170905210626_create_configuration_templates.rb
+++ b/db/migrate/20170905210626_create_configuration_templates.rb
@@ -1,0 +1,12 @@
+class CreateConfigurationTemplates < ActiveRecord::Migration[5.0]
+  def change
+    create_table :configuration_templates do |t|
+      t.bigint :ems_id
+      t.string :external_id
+      t.string :name
+      t.string :description
+      t.boolean :user_defined
+      t.boolean :in_use
+    end
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -572,7 +572,7 @@ configuration_tags_configured_systems:
 configuration_templates:
 - id
 - ems_id
-- external_id
+- ems_ref
 - name
 - description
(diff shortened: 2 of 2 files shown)


skovic:
@miq-bot add_label wip

db/migrate/20170905210626_create_configuration_templates.rb
```diff
@@ -0,0 +1,12 @@
+class CreateConfigurationTemplates < ActiveRecord::Migration[5.0]
+  def change
+    create_table :configuration_templates do |t|
+      t.bigint :ems_id
+      t.string :external_id
```
skovic: The _external_id_ field is used to represent the unique ID that's generated by LXCA for each configuration pattern.

Fryguy: fields ending in `_id` are reserved for ActiveRecord relations.   Since this is a unique id, let's use ems_ref so that it matches all of the other tables, since that is the "native unique id"

db/migrate/20170905210626_create_configuration_templates.rb
```diff
@@ -0,0 +1,12 @@
+class CreateConfigurationTemplates < ActiveRecord::Migration[5.0]
```
Fryguy: @bdunne Can you review here as well.  I'm ok with configuration_templates, but could that get confusing with configuration manager based classes?

db/migrate/20170905210626_create_configuration_templates.rb
skovic: @Fryguy So are you suggesting to change _external_id_ to _ems_ref_? The _external_id_ field is meant to represent the unique ID that is generated by LXCA for each config pattern. We are storing this in the DB, so when the user selects a particular pattern, we can retrieve and send the correct ID to the LXCA REST interface to execute the pattern. Just making sure I understand because I notice that _ems_ref_ is used in other tables and thought it was for MIQ-specific purposes. Thanks

Fryguy: Correct.  ems_ref is the *native* unique identifier, so that sounds exactly like what you are describing.

skovic:
@miq-bot remove_label wip

db/migrate/20170905210626_create_configuration_templates.rb
bdunne: I agree that it will be confusing, but I don't have any better names for it right now.

skovic:
@Fryguy @bdunne Can we proceed with this PR or do you want us to come up with another name? I am also unsure about what would be a better name. Thanks

Fryguy:
Yeah, we can merge this.

PR #67, opened 2017-09-18
Fix issue with showback_rate belongs and type of the data in steps let INFINITY

aljesusg:
- Change showback_rates to showback_rate  => Error
- Change type of steps to float for let INFINITY value

### DIFF
--- a/db/migrate/20170918190738_issue_shorback_tiers.rb
+++ b/db/migrate/20170918190738_issue_shorback_tiers.rb
@@ -1,0 +1,13 @@
+class IssueShorbackTiers < ActiveRecord::Migration[5.0]
+  def up
+    rename_column    :showback_tiers, :showback_rates_id, :showback_rate_id
+    change_column    :showback_tiers, :tier_start_value,  :float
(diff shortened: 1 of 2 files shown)


aljesusg:
@miq-bot add_label bug

sergio-ocon:
@gtanzillo @Fryguy 
I am afraid we didn't see this coming. Could you please merge it?

db/migrate/20170918190738_issue_shorback_tiers.rb
```diff
@@ -0,0 +1,13 @@
+class IssueShorbackTiers < ActiveRecord::Migration[5.0]
+  def up
+    rename_column    :showback_tiers, :showback_rates_id, :showback_rate_id
+    change_column    :showback_tiers, :tier_start_value,  :float
+    change_column    :showback_tiers, :tier_end_value,    :float
```
jrafanie: To be clear, the current chargeback uses floats and we mistakingly are using bigint here.  We only really want to use Float for Float::INFINITY.

PR #68, opened 2017-09-19
Fix db:migrate:reset by setting DatabaseTasks.migrations_paths

NickLaMuro:
The task `db:migrate:reset` is something that is included in rails, and is effectively a `rake db:drop db:create db:migrate`.

Explanation
-----------

The definition of `rake db:drop` however looks like this:

```ruby
 # From active_record/railties/databases.rake

task :load_config do
  puts Rails.application.__id__
  ActiveRecord::Base.configurations       = ActiveRecord::Tasks::DatabaseTasks.database_configuration || {}
  ActiveRecord::Migrator.migrations_paths = ActiveRecord::Tasks::DatabaseTasks.migrations_paths
end

 # ...

desc 'Drops the database from DATABASE_URL ....'
task :drop => [:load_config, :check_protected_environments] do
  puts Rails.application.__id__
  db_namespace["drop:_unsafe"].invoke
end
```

Importantly, the `db:drop` command doesn't require the Rails application to be initialized, thus the `.migrations_paths` module method of `ActiveRecord::Tasks::DatabaseTasks` is called prior to the rails environment being loaded, and that method is memoized.  This means that the defaults for rails are what are set for the `.migrations_paths` of `ActiveRecord::Tasks::DatabaseTasks`, and not what we configure in the `manageiq-schema` initializer.

This fix makes sure that we redefine this attribute when the `:environment` task is called.

Alternative solution
--------------------

What might make more sense is to instead have this path changing to a method, and have it be called out to in `config/application.rb` in the main app.  This would then prevent the need for the line:

```ruby
ActiveRecord::Migrator.migrations_paths = app.config.paths["db/migrate"]
```

from being necessary.  This might not be possible, so for this fix, I just did what was already in place, but figured I would bring this up as a discussion topic.

### DIFF
--- a/lib/manageiq/schema/engine.rb
+++ b/lib/manageiq/schema/engine.rb
@@ -9,6 +9,7 @@ module ManageIQ
             app.config.paths["db/migrate"] << expanded_path
           end
           ActiveRecord::Migrator.migrations_paths = app.config.paths["db/migrate"]
+          ActiveRecord::Tasks::DatabaseTasks.migrations_paths = app.config.paths["db/migrate"].to_a
         end
       end
     end


NickLaMuro:
@miq-bot assign @Fryguy 

cc @carbonin  (since you did the fix that was above this one)

carbonin:
This seems okay.

As for the "Alternative Solution", I'm all for making this simpler as none of this seems like an obvious thing to have to do, but we should probably get this in first and make an issue to discuss why these are needed and possible solutions outside the scope of fixing this particular issue.

NickLaMuro:
> As for the "Alternative Solution", I'm all for making this simpler as none of this seems like an obvious thing to have to do, but we should probably get this in first and make an issue to discuss why these are needed and possible solutions outside the scope of fixing this particular issue.

@carbonin sorry I didn't to respond to this before "#yolomerge magee" (aka @Fryguy ), but yes, I completely agree, and I can see about handling this via the "Alternative Solution" route in some subsequent PRs when I get a moment.

P.S.  Mostly commenting here because I wanted an excuse to call @Fryguy "#yolomerge magee"... for teh lolz

PR #69, opened 2017-09-19
Add parent/child support for lans, as well as subnets

djberg96:
This adds a `parent_id` column to the `lans` table which will be used to support VM networks for SCVMM and possibly other providers. A VM network will be represented as a child of a LAN.

It also adds a `vm_subnets` table with a FK to the `lans` table. The plan is for the AR model to scope the linkage.

### DIFF
--- a/db/migrate/20170919211256_add_parent_id_to_lans.rb
+++ b/db/migrate/20170919211256_add_parent_id_to_lans.rb
@@ -1,0 +1,5 @@
+class AddParentIdToLans < ActiveRecord::Migration[5.0]
+  def change
+    add_column :lans, :parent_id, :bigint
+  end
+end
--- a/db/migrate/20170919211745_create_subnets.rb
+++ b/db/migrate/20170919211745_create_subnets.rb
@@ -1,0 +1,11 @@
+class CreateSubnets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :subnets do |t|
+      t.string :ems_ref
+      t.string :name
+      t.string :cidr
+      t.string :type
+      t.bigint :lan_id
+    end
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -1617,6 +1617,7 @@ lans:
 - computed_allow_promiscuous
 - computed_forged_transmits
 - computed_mac_changes
+- parent_id
 ldap_domains:
 - id
 - name
@@ -6099,6 +6100,13 @@ storages_vms_and_templates:
 - storage_id
 - vm_or_template_id
 - id
+subnets:
+- id
+- ems_ref
+- name
+- cidr
+- type
+- lan_id
 switches:
 - id
 - name


djberg96:
@bronaghs @Ladas @agrare Thoughts on names, columns, etc?

db/migrate/20170919211745_create_vm_subnets.rb
```diff
@@ -0,0 +1,11 @@
+class CreateVmSubnets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :vm_subnets do |t|
```
Ladas: @agrare should we reuse our cloud_subnet model? Ignoring the bad naming for now...

Fryguy:
We are in the process of modeling a new (alternate?) network model with respect to physical networking, so once we get through that, I'd like to discuss in more detail to see where this and cloud networking fits in.

Fryguy:
cc @rodneyhbrown7

db/migrate/20170919211745_create_vm_subnets.rb
djberg96: @agrare @Ladas I think that would result in more confusion, especially as that table seems to expect relationships to other cloud_xxx tables.

Ladas: right, so best to get together with @Fryguy, since there seems to be some architecture shift. Not sure if we should use different Model for very similar things (AFAIK subnet is just subnet)

bronaghs:
@Fryguy - 
Re. your comment [above](https://github.com/ManageIQ/manageiq-schema/pull/69#issuecomment-330862781). Can Dan proceed with this SCVMM work as designed or should we put it on hold to determine the impact the Lenovo networking work might have on it.
Note - these SCVMM networking enhancements are slated for 4.6 - feature freeze is sprint 72.

agrare:
Lets confirm with @Fryguy but I think we should make this a generic subnet model that can be used by physical when they get there and make it STI-able so we can subclass it as needed.

When we get further into a generic set of network models we can look at migrating the `cloud_*` network models over.

blomquisg:
Ah, I read [@Fryguy's comment](https://github.com/ManageIQ/manageiq-schema/pull/69#issuecomment-330862781) to mean, *after* this gets merged for the "G" release, let's revisit this and see what refactoring needs to take place.  Let's definitely get on the same page for this.

djberg96:
Ok, I've renamed the table to just "subnets" for now.

djberg96:
Removed the WIP label. @agrare @Fryguy Unless you want something else, I think this is all we need on the schema side for now.

Fryguy:
@bdunne Thoughts?

agrare:
How about adding a `cidr` field?  Maybe `type` so we can subclass it later?

db/migrate/20170919211745_create_subnets.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSubnets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :subnets do |t|
+      t.string :ems_ref
+      t.string :name
+      t.integer :lan_id
+    end
+
+    add_foreign_key :subnets, :lans
```
Fryguy: we don't use foreign keys, so please remove

djberg96:
@agrare @Fryguy Ok, updated.

db/migrate/20170919211745_create_subnets.rb
Ladas: We should use foreign keys though, for 5.0, I will be adding them to multiple places. Since it's a new table, it might be worth to treat the DB the right way. ;-) (subnet should always be under network, it's one of the easy models)

That being said, it might open an ugly can of worms, so might be good to have a full release to solve it. :-)

Fryguy: Sure, once we go to 5.0, we can discuss...the original reason was because of replication which is a moot point, but I don't feel comfortable adding them until a major release

db/migrate/20170919211256_add_parent_id_to_lans.rb
```diff
@@ -0,0 +1,5 @@
+class AddParentIdToLans < ActiveRecord::Migration[5.0]
+  def change
+    add_column :lans, :parent_id, :integer
```
Fryguy: Must be a bigint

db/migrate/20170919211745_create_subnets.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSubnets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :subnets do |t|
+      t.string :ems_ref
+      t.string :name
+      t.string :cidr
+      t.string :type
+      t.integer :lan_id
```
Fryguy: Must be a bigint

djberg96:
@Fryguy Updated to bigint.

Not sure what the failures are about. The specs pass locally.

agrare:
@djberg96 you need to add the rest of the new columns (cidr, type) to the db/schema file

djberg96:
@agrare I thought I had. Anyway, updated. Thanks!

djberg96:
I don't understand what codeclimate is whining about.

djberg96:
@Fryguy Updated to bigint.

PR #70, opened 2017-09-20
Add an ID to the event that refer to origin system's event

AndreyMenezes:
This PR do:
- Add an `ems_ref` attribute. This attribute refers to the unique sequential ID for the events in the source system.

### DIFF
--- a/db/migrate/20170920212535_add_event_id_to_event_streams.rb
+++ b/db/migrate/20170920212535_add_event_id_to_event_streams.rb
@@ -1,0 +1,5 @@
+class AddEventIdToEventStreams < ActiveRecord::Migration[5.0]
+  def change
+    add_column :event_streams, :ems_ref, :string
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -1263,6 +1263,7 @@ event_streams:
 - middleware_deployment_name
 - generating_ems_id
 - physical_server_id
+- ems_ref
 ext_management_systems:
 - id
 - name


AndreyMenezes:
@juliancheal

db/migrate/20170920212535_add_event_id_to_event_streams.rb
```diff
@@ -0,0 +1,5 @@
+class AddEventIdToEventStreams < ActiveRecord::Migration[5.0]
+  def change
+    add_column :event_streams, :event_id, :bigint
```
agrare: I think we want the reference here to be a string so that it can be used by providers where the event unique reference isn't an int.

As for the name of the column, typically we use either `:ems_ref` and/or `:uid_ems` as the "provider reference".  @Fryguy thoughts/preferences?

AndreyMenezes: No problem to change to string.

For the name of the column, I think use `ems` in the name could not be a good choice. We already have others attributes to refer a provider like ems_id, physical_server_id...

agrare:
@miq-bot assign @Fryguy

juliancheal:
LGTM

juliancheal:
This PR needs to be merged before https://github.com/ManageIQ/manageiq-providers-lenovo/pull/82

AndreyMenezes:
@Fryguy Could you merge this PR if agree with the approval?

Issue #4, opened 2017-06-20
Document the migration "rules"


Fryguy:
- Use say_with_time
- All data migrations *must* have specs

lpichler:
- when you are working with the model in data migration, define the model in the migration
- when there is model defined in your migration and it has column `type`, set 
   `self.inheritance_column = :_type_disabled` example
```ruby
 class CloudNetwork < ActiveRecord::Base
    self.inheritance_column = :_type_disabled
  end
```

Fryguy:
- Describe the "why" behind having the rule that we don't use default, null, foreign keys, etc.

lpichler:
- don't mix data and structural migration to the one migration, create separate migrations

Fryguy:
- Explain the pitfalls of deserializing columns that might have serialized objects in them.  In particular, MiqExpression, which is common, and describe avoiding desrialization and using a LIEK query can be used instead.
- Describe the benefits of update_all
- Describe batching and when it's possible to use a reentrant migration for LARGE data migrations.

lpichler:
@miq-bot remove_label stale

Fryguy:
- We need a note about if `remove_column` appears in a PR, there needs to be a linked PR in other repos showing that those columns are no longer used.

- Might wants to describe the different "hats" needed to be worn for review.
  - DBA - review for database schema issues that could lead to performance/inefficiency problems, as well as index selection
  - Rails - review for Rails-invalid column names such as non-foreign-keys that end in `_id` or a `type` column that isn't being used for STI
  - ManageIQ - review for understanding where the new tables fits in the grand scheme and what use cases it is supporting.  For providers, it's important to understand if the new tables and the naming of things fits into the single pane of glass, and how those new tables would be reused by other provider of the same provider type.

Fryguy:
Migrations should not create new rows if rows didn't previously exist...that is, an empty database migrated should still be empty in the end, particularly for test reasons.  Data migrations of existing data can transform into new rows if necessary.

lpichler:
Use `ActiveRecord::Base` for models in migrations.

> Part of the advantage of ApplicationRecord is that we can leave AR::Base unmolested, which is exactly what migrations want: a plain ordinary AR object with no app-code-of-the-day surprises.

taken from
https://github.com/ManageIQ/manageiq/pull/7312/files#r56356275

PR #72, opened 2017-09-21
Add status column to MiddlewareDomain

xeviknal:
Related to issue [#44](https://github.com/ManageIQ/manageiq-providers-hawkular/issues/44)

Adding status to MiddlewareDomain model.


xeviknal:
@josejulio could you please check this out? Happy to have a code review :)

lpichler:
@miq-bot assign @Fryguy

abonas:
cc @agrare

xeviknal:
@lpichler good point. I'll remove it, I am testing a method which is already tested by rails.

Gotta say, that I wanted to check some of the tooling you have for testing migrations. My first time on a project with that :)

db/migrate/20170921110959_add_status_to_middleware_domain.rb
```diff
@@ -0,0 +1,8 @@
+class AddStatusToMiddlewareDomain < ActiveRecord::Migration[5.0]
+  class MiddlewareDomain < ActiveRecord::Base
+  end
```
agrare: I think you want to disable STI here
`self.inheritance_column = :_type_disabled # disable STI`

xeviknal: Could you please elaborate that a little bit?
I don't understand how the migration can change by disabling the STI

agrare: @Fryguy can correct me if I'm wrong but I believe it is so we don't end up loading the models during the migration if the record is subclassed with STI

abonas:
@josejulio @israel-hdez I've seen a bit different handling/persisting of status in the middleware server case - isn't all status related info stored in a properties object and not in a separate field? (availability, state, etc) I just want to make sure we are not doing something inconsistent here, but if there is indeed some difference that is ok - then we can leave this as is.

josejulio:
Deployments have their status as a [separate](https://github.com/xeviknal/manageiq-schema/blob/e32cc425fa579d5b62627bff84e1aee508b0521b/db/schema.yml#L4685) field.
Not sure about Servers, but could not see any separate field.

israel-hdez:
Servers don't have status in its own column. The status and the availability are both stored in the `properties` column/hash.

Note that `MiddlewareDomain` is storing the following data in the properties hash:

```
Suspend State: 
Running Mode: NORMAL
Home Directory: "/opt/jboss/wildfly"
Version: 10.1.0.Final
Server State: 
Product Name: WildFly Full
Host State: running
Is Domain Controller: 'true'
UUID: 2dc495bf-5c2c-4074-8487-023ca7427fae
Name: master
```

Note that there is a _Host State_ property. Is this what you want?

Also, be warned that the _Host State_ most likely is reported by inventory and may not be the current status. Usually, you'll want to check the availability metric of the domain host (like the case of the servers).

xeviknal:
thanks for that @israel-hdez and @josejulio.

Regarding your question, I don't exactly know which attribute of the hash is the one that says the availability of the domain. I guess it is, yes, the `Host State`. Could you please confirm that?

Regarding to the last sentence, I guess you mean that I should add logic to AvailabilityUpdates parser/collector in order to have current domain availability. My approach, thanks to @josejulio, [here](https://github.com/ManageIQ/manageiq-providers-hawkular/compare/master...xeviknal:add-status-domain?expand=1#diff-3ed566a427a970586d08bbbc5af1614cR28)
Is that what you meant?

josejulio:
@xeviknal I think we are missing this part: https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L177-L198

db/migrate/20170921110959_add_status_to_middleware_domain.rb
lpichler: `self.inheritance_column = :_type_disabled # disable STI`
here is when the table has column `type`. For more details see:https://github.com/ManageIQ/manageiq-schema/issues/4

Anyway, you don't need to define model here, as you are not using the model in migration either in spec.

xeviknal:
@josejulio don't you thing that properties for each domain are already stored before ([`MiddlewareManager#fetch_domains_with_servers`](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L14)), thus, the availability is already stored.
Here the stacktrace:
[MiddlewareManager#L14](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L14): fetch domains with servers.
[MiddlewareManager#L62](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L62): parsing middleware domain.
[MiddlewareManager#L312](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L312): mapping hawkular item to inventory_object
[MiddlewareManager#L359](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L359): map :ems_ref, :nativeid, :feed, :properties.

Then, probably the lines you sent, are not needed in this case. Anything to say @israel-hdez ?

josejulio:
> @josejulio don't you thing that properties for each domain are already stored before (MiddlewareManager#fetch_domains_with_servers), thus, the availability is already stored.
> Here the stacktrace:
> MiddlewareManager#L14: fetch domains with servers.
> MiddlewareManager#L62: parsing middleware domain.
> MiddlewareManager#L312: mapping hawkular item to inventory_object
> MiddlewareManager#L359: map :ems_ref, :nativeid, :feed, :properties.

IIRC that fetches the data/properties from the Inventory. As @israel-hdez pointed out **Host State** is reported on a given refresher and could be outdated. 
We fetch the availability metric for the given resource ([server](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L190-L198) or [deployment](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L183-L188)) and would need to do the same for the domain.

I have checked the default [configuration](https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml), and I don't see any avail metric for the domain itself(only for the domain servers). I suppose we might need to add one.

@israel-hdez, Can you please correct me if I'm wrong?

israel-hdez:
@josejulio @xeviknal Yes, _Host State_ is reported by inventory and usually is not updated when the domain goes down.
Hmm... Yes, there is no avail metric in the config for the domain. Perhaps you could try querying Hawkular to see if, by any chances, it's there?

BTW, I checked the changes in @xeviknal branch. The `availability_changes` parser/collector/persister are there for use in the event catcher. So, you are missing some changes here: https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/stream.rb#L49

And also, the changes that @josejulio said are also required so that availabilities are collected when a refresh is done.

abonas:
@miq-bot add_label providers/middleware

xeviknal:
Okay, I am closing this PR since there is no need to add a new column to domain. Availability is going to be stored on MiddlewareDomain#properties.

PR #74, opened 2017-09-24
Add requests and limits to Persistent Volume Claim

zakiva:
Related PRs: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/116, https://github.com/ManageIQ/manageiq/pull/16026

@simon3z @zeari Please review

### DIFF
--- a/db/migrate/20170924102942_add_requests_and_limits_to_persistent_volume_claims.rb
+++ b/db/migrate/20170924102942_add_requests_and_limits_to_persistent_volume_claims.rb
@@ -1,0 +1,6 @@
+class AddRequestsAndLimitsToPersistentVolumeClaims < ActiveRecord::Migration[5.0]
+  def change
+    add_column  :persistent_volume_claims, :requests, :text
+    add_column  :persistent_volume_claims, :limits, :text
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -5612,6 +5612,8 @@ persistent_volume_claims:
 - created_at
 - updated_at
 - container_project_id
+- requests
+- limits
 physical_servers:
 - id
 - ems_id


Fryguy:
@zakiva @simon3z I'm ok with this, but why is this modeled as a serialized Hash as opposed to fully modeled out?  Will we ever want it more modeled out in the future?

simon3z:
> Will we ever want it more modeled out in the future?

@zakiva can you reference the relevant BZ in the description? I wonder if indeed we want to model that now.

zakiva:
> @zakiva can you reference the relevant BZ in the description? I wonder if indeed we want to model that now.

I didn't find a relevant BZ for that (maybe @Loicavenel knows if we have one).
Anyway, this is part of the storage for chargeback feature @zeari is working on. We currently collect only the actual capacity of the pvc's bounded volume, which also saved as  a serialized hash and been accessed in some places, [here](https://github.com/ManageIQ/manageiq/pull/15973/files#diff-b0d42d6c800ea62f58c6e0c68ae483adR6) for example.

zakiva:
@simon3z Do you want to merge this before the freeze?

Loicavenel:
@zakiva @zeari can you tell me what is this PR is about? I am not sure I understand it. Is it to have Chargeback for VolumeS?

zeari:
>@zakiva @zeari can you tell me what is this PR is about? I am not sure I understand it. Is it to have Chargeback for VolumeS?

@loicavenel yes.

Loicavenel:
@zeari can you be more explicite? We do have PV and the Size... this is fixed.. how it can be wrong?

zakiva:
> @zeari can you be more explicite? We do have PV and the Size... this is fixed.. how it can be wrong?

@Loicavenel The claim can have 3 different values related to the storage size: 
1. allocated - this is the actual allocated storage of the bounded volume. 
2. requested - the requested size by the claim, this is a lower bound for the allocated size. 
3. limit - upper bound. 

We currently persist only the first one, this PR adds the other two.

zeari:
> @zeari can you be more explicite? We do have PV and the Size... this is fixed.. how it can be wrong?

The capacity we are currently listing is the size of PV that was assigned to a pod\project instead of how much the claim asked for. If you request 1GB volume, you might be assigned a 2GB volume since that is the volume closest to the size to what you ask. So the size we know about would be 2GB instead if what was actually requested.
It makes more sense to charge for what someone asked for instead of what they were *upgraded* to.
@zakiva @simon3z Correct me if im wrong.

Loicavenel:
@zeari so, Openshift is rounding up to the closest higher GB?

zeari:
The process AFAIK
1) a project\pod requests a PV of size x
2) openshift has PVs of size 1GB, 2GB, 3GB...
3) openshift finds a PV with a size closest to x and binds it to that project

Loicavenel:
I look in the doc here: https://docs.openshift.com/container-platform/3.5/dev_guide/compute_resources.html.. I don't see anything in Limits regarding PV storage. But I see it in Quota... requests.storage.
Where the PoD request will come from? how do we know what was requested?

zakiva:
> I look in the doc here: https://docs.openshift.com/container-platform/3.5/dev_guide/compute_resources.html.. I don't see anything in Limits regarding PV storage. But I see it in Quota... requests.storage.

These limits/quotas are in project level (summary of resources used in the project). The requested storage in this PR is per specific Persistent Volume Claim which represents a specific user request for storage, e.g. one pod needs 1GB.

zakiva:
@Fryguy PTAL

PR #78, opened 2017-09-26
Change rake spec to run all specs and be the default

### LINKED ISSUE #65: Change rake tasks so that rake spec does everything
Unlike other repos, `rake spec` does not run `spec:migrations` but instead only the "other" specs.  We should change the current spec to something like `spec:common` and then `rake spec` will do both.  (The `rake` default would then just be calling `rake spec`)

Fryguy:
Fixes #65

@chessbyte Please review.


Fryguy:
cc @sergio-ocon

PR #79, opened 2017-09-27
Drop Container Component Statuses

zakiva:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1485431

### DIFF
--- a/db/migrate/20170927114802_drop_container_component_statuses.rb
+++ b/db/migrate/20170927114802_drop_container_component_statuses.rb
@@ -1,0 +1,16 @@
+class DropContainerComponentStatuses < ActiveRecord::Migration[5.0]
+  def up
+    drop_table :container_component_statuses
+  end
+
+  def down
+    create_table :container_component_statuses do |t|
+      t.belongs_to :ems, :type => :bigint
+      t.string     :name
+      t.string     :condition
+      t.string     :status
+      t.string     :message
+      t.string     :error
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


zakiva:
@simon3z @cben @zeari Please review

zakiva:
@miq-bot add_label cleanup

cben:
Please describe in which order those should be merged.

simon3z:
@zakiva @cben @moolitayer do we want to remove this down to the schema (as this PR is doing) or do we think that keeping it would be nice if in the future it is properly implemented on the OpenShift side?

I am OK to remove this (also because future implementation is not laid out yet) but I also know that this is a nice OpenShift feature that may be scheduled for one of the upcoming releases.

moolitayer:
I don't have a strong preference. :+1: to remove.

I think this PR depends on https://github.com/ManageIQ/manageiq/pull/16055 (removing the class definition)

cben:
I also lean towards removing.
But schema is special, because re-adding it won't be backportable.  Is there chance we'll want to re-add it in z-streams?

simon3z:
> But schema is special, because re-adding it won't be backportable. Is there chance we'll want to re-add it in z-streams?

@cben no chance but also deleting it in this release to re-add it next release seems an overhead (although, again, we do not have details on the future implementation, we may have to add many new fields).

Fryguy:
I'm ok with deleting now and adding back later.  There's a good chance you'll have different columns anyway.

PR #80, opened 2017-09-27
Migrate existing dialog field association data to use new relationship

d-m-u:
The planned targeted auto-refresh enhancement to dialog fields requires the dialog fields to be able to be associated to one another. This migration takes our old system of relating dialog fields by order-dependent auto_refresh and trigger_auto_refresh fields and updates them to use the recently added DialogFieldAssociation table. It's worth noting that this may not upgrade all existing dialog associations correctly and that it may be necessary to manually check and possibly fix some of the associations not caught by this migration.

### DIFF
--- a/db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
+++ b/db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
@@ -1,0 +1,76 @@
+class MigrateDialogFieldAssociationsToUseNewRelationship < ActiveRecord::Migration[5.0]
+  class Dialog < ActiveRecord::Base
+  end
+
+  class DialogTab < ActiveRecord::Base
+    belongs_to :dialog, :class_name => "MigrateDialogFieldAssociationsToUseNewRelationship::Dialog"
+  end
+
+  class DialogGroup < ActiveRecord::Base
+    belongs_to :dialog_tab, :class_name => "MigrateDialogFieldAssociationsToUseNewRelationship::DialogTab"
+  end
+
+  class DialogField < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+    belongs_to :dialog_group, :class_name => "MigrateDialogFieldAssociationsToUseNewRelationship::DialogGroup"
+  end
+
+  class DialogFieldAssociation < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Migrating existing dialog field associations to new relationship") do
+      dialog_with_fields.each do |_dialog, fields|
+        dialog_triggers = dialog_fields_with_trigger_auto_refresh(fields).sort_by! { |trigger| trigger[:position] }
+        dialog_responders = dialog_fields_with_auto_refresh(fields)
+        dialog_triggers.each_with_index do |trigger, index|
+          specific_responders = dialog_responders.select { |responder| responder_range(trigger, trigger[index + 1]).cover?(responder[:position]) }
+          set_associations(trigger, specific_responders)
+        end
+      end
+    end
+  end
+
+  def down
+    DialogFieldAssociation.delete_all
+  end
+
+  private
+
+  def absolute_position(dialog_fields)
+    dialog_fields.collect do |f|
+      field_position = f.position
+      dialog_group_position = f.dialog_group.position
+      dialog_tab_position = f.dialog_group.dialog_tab.position
+      index = field_position + dialog_group_position * 1000 + dialog_tab_position * 100_000
+      {:id => f.id, :position => index}
+    end
+  end
+
+  def dialog_fields_with_auto_refresh(dialog_fields)
+    absolute_position(dialog_fields.select(&:auto_refresh))
+  end
+
+  def dialog_fields_with_trigger_auto_refresh(dialog_fields)
+    absolute_position(dialog_fields.select(&:trigger_auto_refresh))
+  end
+
+  def responder_range(trigger_min, trigger_max)
+    min = trigger_min[:position] + 1
+    max = trigger_max.present? ? trigger_max[:position] - 1 : 100_000_000
+    (min..max)
+  end
+
+  def set_associations(trigger, specific_responders)
+    specific_responders.each { |responder| DialogFieldAssociation.create!(:trigger_id => trigger[:id], :respond_id => responder[:id]) }
+  end
+
+  def dialog_with_fields
+    DialogField
+      .where(:auto_refresh => true)
+      .or(DialogField.where(:trigger_auto_refresh => true))
+      .includes(:dialog_group => {:dialog_tab => :dialog})
+      .group_by { |f| f.dialog_group.try(:dialog_tab).try(:dialog) }
+      .except(nil)
+  end
+end
--- a/spec/migrations/20170927135007_migrate_dialog_field_associations_to_use_new_relationship_spec.rb
+++ b/spec/migrations/20170927135007_migrate_dialog_field_associations_to_use_new_relationship_spec.rb
@@ -1,0 +1,89 @@
+require_migration
+
+describe MigrateDialogFieldAssociationsToUseNewRelationship do
+  let(:dialog_field_stub) { migration_stub(:DialogField) }
+  let(:dialog_group_stub) { migration_stub(:DialogGroup) }
+  let(:dialog_tab_stub) { migration_stub(:DialogTab) }
+  let(:dialog_stub) { migration_stub(:Dialog) }
+  let(:dialog_field_association_stub) { migration_stub(:DialogFieldAssociation) }
+  let(:dialog_group_id) { dialog_group_stub.first.id }
+  let(:dialog_tab_id) { dialog_tab_stub.first.id }
+  let(:dialog_id) { dialog_stub.first.id }
+
+  migration_context :up do
+    before do
+      dialog_stub.create!
+      dialog_tab_stub.create!(:dialog_id => dialog_id, :position => 4)
+      dialog_group_stub.create!(:dialog_tab_id => dialog_tab_id, :position => 7)
+    end
+
+    it "does not create a reference when trigger field is after responder field" do
+      dialog_field_stub.create!(:name => "dialog_field4", :auto_refresh => true, :position => 0, :dialog_group_id => dialog_group_id, :type => "DialogFieldTextbox")
+      dialog_field_stub.create!(:name => "dialog_field1", :trigger_auto_refresh => true, :position => 4, :dialog_group_id => dialog_group_id, :type => "DialogFieldDropdown")
+      expect(dialog_field_association_stub.count).to eq(0)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(0)
+    end
+
+    it "does not create a circular reference" do
+      dialog_field_stub.create!(:name => "dialog_field4", :trigger_auto_refresh => true, :auto_refresh => true, :position => 0, :dialog_group_id => dialog_group_id, :type => "DialogFieldRadioButton")
+      dialog_field_stub.create!(:name => "dialog_field1", :auto_refresh => true, :trigger_auto_refresh => true, :position => 4, :dialog_group_id => dialog_group_id, :type => "DialogFieldTagControl")
+      expect(dialog_field_association_stub.count).to eq(0)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(1)
+      expect(dialog_field_association_stub.first.trigger_id).to eq(dialog_field_stub.first.id)
+      expect(dialog_field_association_stub.first.respond_id).to eq(dialog_field_stub.second.id)
+    end
+
+    it "does create a reference when valid one is present" do
+      dialog_field_stub.create!(:name => "dialog_field4", :auto_refresh => true, :position => 2, :dialog_group_id => dialog_group_id)
+      dialog_field_stub.create!(:name => "dialog_field1", :trigger_auto_refresh => true, :position => 0, :dialog_group_id => dialog_group_id)
+      expect(dialog_field_association_stub.count).to eq(0)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(1)
+      expect(dialog_field_association_stub.first.trigger_id).to eq(dialog_field_stub.second.id)
+      expect(dialog_field_association_stub.first.respond_id).to eq(dialog_field_stub.first.id)
+    end
+
+    it "does not create association when dialog tab is missing" do
+      dialog_tab_stub.delete_all
+      dialog_field_stub.create!(:name => "dialog_field4", :auto_refresh => true, :position => 2, :dialog_group_id => dialog_group_stub.first.id)
+      dialog_field_stub.create!(:name => "dialog_field1", :trigger_auto_refresh => true, :position => 0, :dialog_group_id => dialog_group_stub.first.id)
+      expect(dialog_field_association_stub.count).to eq(0)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(0)
+    end
+
+    it "only creates dialog associations if the fields are on the same dialog" do
+      dialog_field_stub.create!(:name => "dialog_field4", :trigger_auto_refresh => true, :position => 4, :dialog_group_id => dialog_group_stub.first.id, :type => "DialogFieldDateControl")
+
+      expect(dialog_field_association_stub.count).to eq(0)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(0)
+    end
+  end
+
+  migration_context :down do
+    it "should delete dialog field associations" do
+      dialog_field_stub.create!(:name => "dialog_field6")
+      dialog_field_stub.create!(:name => "dialog_field9")
+      dialog_field_association_stub.create!(:trigger_id => dialog_field_stub.first.id, :respond_id => dialog_field_stub.second.id)
+
+      expect(dialog_field_association_stub.count).to eq(1)
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(0)
+    end
+  end
+end


db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
+class MigrateDialogFieldAssociationsToUseNewRelationship < ActiveRecord::Migration[5.0]
+  Rails.application.eager_load!
```
Fryguy: nope!

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def up
+    Dialog.all.each do |dialog|
```
Fryguy: Create a stub model, as is done in other data migrations, for Dialog and DialogFieldAssociation

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
...
+      fields = dialog.dialog_fields
+      dialog_fields_with_associations = fields.select { |df| !df.auto_refresh.nil? || !df.trigger_auto_refresh.nil? }.sort { |n| n.position * -1 }
```
Fryguy: `!df.auto_refresh.nil? || !df.trigger_auto_refresh.nil?`

simplifies to 

`df.auto_refresh || df.trigger_auto_refresh`

... Also, should those be `auto_refresh?` and `trigger_auto_refresh?` (with trailing `?`)

---

`.sort { |n| n.position * -1 }`

simplifies to

`.sort { |n| -n.position }

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
+class MigrateDialogFieldAssociationsToUseNewRelationship < ActiveRecord::Migration[5.0]
+  Rails.application.eager_load!
+
+  def up
+    Dialog.all.each do |dialog|
+      fields = dialog.dialog_fields
```
Fryguy: you might want an `includes` in the original query or this is going to be a nasty N+1

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
...
+      dialog_fields_with_associations = fields.select { |df| !df.auto_refresh.nil? || !df.trigger_auto_refresh.nil? }.sort { |n| n.position * -1 }
+      triggers = dialog_fields_with_associations.select { |df| df.trigger_auto_refresh == true }
```
Fryguy: if trigger_auto_refresh is a boolean column, this simplifies to 

`.select(&:trigger_auto_refresh)`

(same with the next line)

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,31 @@
...
+      responders = dialog_fields_with_associations.select { |df| df.auto_refresh == true }
+      triggers.each_with_index do |t, index|
+        responder_fields_for_t = if !triggers[index + 1].nil?
```
Fryguy: Prefer coding in the positive as opposed to `! nil?` because double-negative are hard for people to reason about.  So,

```ruby
responder_fields_for_t =
  if triggers[index + 1]
    responders.select { |r| r.position
 ....
```

Fryguy:
Don't forget the specs 😄   All data migrations require specs.

d-m-u:
@eclarizio just working on the testing for this one now

d-m-u:
@miq-bot assign eclarizio

eclarizio:
I'm guessing you meant to add me as a reviewer, since I don't really have any powers as an assignee 😛 

@miq-bot assign @Fryguy 

@Fryguy Looking at the miq-bot repo though, can we even set a reviewer with miq-bot?

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,46 @@
...
+                                 responders.select { |r| r.position > t.position && r.position <= triggers[index + 1].position }.pluck(:id)
+                               else
+                                 responders.select { |r| r.position > t.position }.pluck(:id)
+                               end
+      trigger_field_dialog_id = DialogTab.find(DialogGroup.find(t.dialog_group_id).dialog_tab_id).dialog_id
+      responder_fields_for_t.each do |responder|
+        respond_field_dialog_id = DialogTab.find(DialogGroup.find(DialogField.find(responder).dialog_group_id).dialog_tab_id).dialog_id
+        if trigger_field_dialog_id == respond_field_dialog_id
+          DialogFieldAssociation.create(:trigger_id => t.id, :respond_id => responder)
+        end
```
eclarizio: I don't think we have to change this, but at first glance it was kinda weird to see "if the trigger id equals the responder id, let's create an association", which seems wrong. Maybe a variable name change? Same with donig `:respond_id => responder` *looks* like you're setting it to the object even though you're iterating through a list of ids.

Another way to do it instead of taking *all* the dialog fields right from the start is to iterate through the dialogs themselves and then you can just do `dialog.dialog_fields` instead of having to make sure that the dialog ids match for the fields you're comparing?

I think if we want to do it this way, that's ok, but there needs to be a spec that tests the fact that the fields need to belong to the same dialog.

d-m-u: After discussions with Fryguy and bdunne it seemed easier to go by field first because otherwise the tests have to include all the information about dialogs and tabs and groups and those relationships whereas all that's needed here is the class definition.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,44 @@
+class MigrateDialogFieldAssociationsToUseNewRelationship < ActiveRecord::Migration[5.0]
+  class DialogFieldAssociation < ActiveRecord::Base
+    include ReservedMigrationMixin
```
gmcculloug: We are not using Reserved data columns so this should not be needed in any of the models.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,44 @@
...
+    include MigrationStubHelper # NOTE: Must be included after other mixins
+  end
+  class DialogGroup < ActiveRecord::Base
+    include ReservedMigrationMixin
+    include MigrationStubHelper # NOTE: Must be included after other mixins
+  end
+
+  def up
+    dialog_fields_with_associations = DialogField.all.select { |df| df.auto_refresh || df.trigger_auto_refresh }.sort { |n| -n.position }
```
gmcculloug: Sorting the DialogFields by the `position` column is only taking into account the field position within the group box that it is in.  Tabs and groups also have a `position` column so it is easy to rearrange the different dialog elements.  What you are really looking for here is an overall field position which would need to take into account the position of the tab and the group that the field is in.

Does it make sense to add a calculation method in these stub methods to calculate that distinct position?
Something like:  `(tab.position * 1000) + (group.position * 100) + field.position`

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def up
+    dialog_fields_with_associations = DialogField.all.select { |df| df.auto_refresh || df.trigger_auto_refresh }
+    absolute_position = {}
+    unless dialog_fields_with_associations.blank?
+      dialog_fields_with_associations.each do |f|
+        index =
+          f.position +
+          DialogGroup.find(f.dialog_group_id).position * 100 +
+          DialogTab.find(DialogGroup.find(f.dialog_group_id).dialog_tab_id).position * 1_000 +
```
gmcculloug: Maybe make this one 10_000

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,62 @@
...
+          DialogTab.find(DialogGroup.find(DialogField.find(f.id).dialog_group_id).dialog_tab_id).dialog_id * 1_000_000
```
gmcculloug: Why is this one needed?

d-m-u: Because we can have two tabs with the same position and I need to check to make sure the fields are on the same dialog anyway so why not do it there?

spec/migrations/20170927135007_migrate_dialog_field_associations_to_use_new_relationship_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:dialog_group_stub) { migration_stub(:DialogGroup) }
+  let(:dialog_tab_stub) { migration_stub(:DialogTab) }
+  let(:dialog_stub) { migration_stub(:Dialog) }
+  let(:dialog_field_association_stub) { migration_stub(:DialogFieldAssociation) }
+
+  migration_context :up do
+    before(:each) do
+      dialog_group_stub.create!(:id => 2, :dialog_tab_id => 8, :position => 7)
+      dialog_tab_stub.create!(:id => 8, :dialog_id => 10, :position => 4)
+      dialog_stub.create!(:id => 10)
```
Fryguy: Do not use manual ids because it bypasses proper region testing.

Fryguy: Instead create the record, let ActiveRecord create a proper id, then use that id after the fact to update the associations.

spec/migrations/20170927135007_migrate_dialog_field_associations_to_use_new_relationship_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+describe MigrateDialogFieldAssociationsToUseNewRelationship do
+  let(:dialog_field_stub) { migration_stub(:DialogField) }
+  let(:dialog_group_stub) { migration_stub(:DialogGroup) }
+  let(:dialog_tab_stub) { migration_stub(:DialogTab) }
+  let(:dialog_stub) { migration_stub(:Dialog) }
+  let(:dialog_field_association_stub) { migration_stub(:DialogFieldAssociation) }
+
+  migration_context :up do
+    before(:each) do
```
Fryguy: :each is redundant, so you can remove it.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+  def absolute_position(dialog_fields)
+    positions = []
+    dialog_fields.each do |f|
+      field_position = f.position
+      dialog_group_position = DialogGroup.find(f.dialog_group_id).position
+      dialog_tab_position = DialogTab.find(DialogGroup.find(f.dialog_group_id).dialog_tab_id).position
+      index = field_position + dialog_group_position * 100 + dialog_tab_position * 10_000
+      positions << {:id => f.id, :position => index}
+    end
+    positions
```
Fryguy: prefer .collect over the 

```ruby
x = []
foo.each { x << y }
x
```

pattern.

```ruby
x = foo.collect { y }
```

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def absolute_position(dialog_fields)
+    positions = []
+    dialog_fields.each do |f|
+      field_position = f.position
+      dialog_group_position = DialogGroup.find(f.dialog_group_id).position
+      dialog_tab_position = DialogTab.find(DialogGroup.find(f.dialog_group_id).dialog_tab_id).position
+      index = field_position + dialog_group_position * 100 + dialog_tab_position * 10_000
```
Fryguy: Why these numbers?  Are they just arbitrarily for segmenting?  I only ask because I remember a customer with over 100 fields on a form, and so this could cause issues.  Even so, since you are updating the position, I still think it's weird to segment and then leave it that way...I would expect something to post-process the "weird" numbers, and put them into "normal" numbers by sorting and mapping the indexes or something.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+    absolute_position(dialog_fields_with_associations.select { |df| df.auto_refresh == true })
+  end
+
+  def dialog_fields_with_trigger_auto_refresh(dialog_fields_with_associations)
+    absolute_position(dialog_fields_with_associations.select { |df| df.trigger_auto_refresh == true })
+  end
+
+  def responder_range(trigger_min, trigger_max)
+    min = trigger_min[:position] + 1
+    max = trigger_max.present? ? trigger_max[:position] - 1 : 10_000_000
```
Fryguy: Also curious about this 10 million number.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+  def set_associations(trigger, specific_responders)
+    specific_responders.each { |responder| DialogFieldAssociation.create(:trigger_id => trigger[:id], :respond_id => responder[:id]) }
+  end
+
+  def dialog_fields_with_associations_by_dialog(dialog_fields_with_associations)
+    dialog_fields_with_associations.group_by { |df| Dialog.find(DialogTab.find(DialogGroup.find(DialogField.find(df.id).dialog_group_id).dialog_tab_id).dialog_id) }
+  end
+
+  def down
+    DialogFieldAssociation.delete_all
```
Fryguy: Will the previous migration just drop the table?  If so, then there's no need for this at all.  (You could just put a comment stating that fact)

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+    max = trigger_max.present? ? trigger_max[:position] - 1 : 10_000_000
+    (min..max)
+  end
+
+  def set_associations(trigger, specific_responders)
+    specific_responders.each { |responder| DialogFieldAssociation.create(:trigger_id => trigger[:id], :respond_id => responder[:id]) }
+  end
+
+  def dialog_fields_with_associations_by_dialog(dialog_fields_with_associations)
+    dialog_fields_with_associations.group_by { |df| Dialog.find(DialogTab.find(DialogGroup.find(DialogField.find(df.id).dialog_group_id).dialog_tab_id).dialog_id) }
```
Fryguy: This looks like an N+1 nightmare.  There are much better approaches for querying all the data, then mixing and matching (we do something very similar with the RelationshipMixin for child objects.  I can sit with you and show you.

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+  end
+
+  def absolute_position(dialog_fields)
+    positions = []
+    dialog_fields.each do |f|
+      field_position = f.position
+      dialog_group_position = DialogGroup.find(f.dialog_group_id).position
+      dialog_tab_position = DialogTab.find(DialogGroup.find(f.dialog_group_id).dialog_tab_id).position
```
Fryguy: Also an N+1 issue (on both of these lines)

d-m-u:
@gmcculloug can you take a quick looksee at this, 'specially the specs, again please?

db/migrate/20170927135007_migrate_dialog_field_associations_to_use_new_relationship.rb
```diff
@@ -0,0 +1,75 @@
...
+  def set_associations(trigger, specific_responders)
+    specific_responders.each { |responder| DialogFieldAssociation.create!(:trigger_id => trigger[:id], :respond_id => responder[:id]) }
+  end
+
+  def dialog_with_fields
+    DialogField
+      .where(:auto_refresh => true)
+      .or(DialogField.where(:trigger_auto_refresh => true))
+      .includes(:dialog_group => {:dialog_tab => :dialog})
+      .group_by { |f| f.dialog_group.dialog_tab.dialog }
```
gmcculloug: @Fryguy had the suggest to protect against this relationship chain hitting a `nil` here and I would agree.  If we use `try` and allow the fields to get grouped together until `nil` we can throw those fields away before processing.

spec/migrations/20170927135007_migrate_dialog_field_associations_to_use_new_relationship_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      migrate
+
+      expect(dialog_field_association_stub.count).to eq(0)
+    end
+  end
+
+  migration_context :down do
+    it "should delete dialog field associations" do
+      migrate
```
Fryguy: This should create a dialog_field_association record before the migrate, otherwise you are testing nothing.

Fryguy:
minor problem with the down migration spec but otherwise looks good to me 👍

PR #81, opened 2017-09-29
Migrate MiddlewareServer to MiddlewareServerWildfly and MiddlewareServerEap

tumido:
Sub-classing MiddlewareServer into more specific classes for Wildfly and EAP
 
Related to: https://github.com/ManageIQ/manageiq-providers-hawkular/pull/60
Resolves: https://issues.jboss.org/browse/HAWKULAR-1248


tumido:
cc @cfcosta please review

db/migrate/20170928202851_migrate_middleware_server_to_wildfly_and_eap.rb
```diff
@@ -0,0 +1,33 @@
+class MigrateMiddlewareServerToWildflyAndEap < ActiveRecord::Migration[5.0]
+  class MiddlewareServer < ApplicationRecord
```
Fryguy: Be sure to add `self.inheritance_column = :_type_disabled` for this stub.

lpichler: `class MiddlewareServer < ActiveRecord::Base`

see details:
https://github.com/ManageIQ/manageiq-schema/issues/4

Fryguy:
Just want approval from @abonas here and on the dependent PR before merging.

tumido:
Thanks @Fryguy 👍 :)

abonas:
@cfcosta please review.
@tumido please see the comment in https://github.com/ManageIQ/manageiq-providers-hawkular/pull/60#pullrequestreview-66319096

db/migrate/20170928202851_migrate_middleware_server_to_wildfly_and_eap.rb
```diff
@@ -0,0 +1,34 @@
+class MigrateMiddlewareServerToWildflyAndEap < ActiveRecord::Migration[5.0]
+  class MiddlewareServer < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time('Migrating middleware_server to middleware_server_wildfly') do
+      MiddlewareServer
+        .where("type = ? AND product ~* ?", 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServer', 'wildfly')
+        .update_all(:type => 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerWildfly')
```
abonas: shouldn't the new wildfly/eap class be a subclass/subtype of middleware server class? here it seems it's in hierarchy of middleware_manager, or I am probably missing something

abonas:
@cfcosta please review

abonas:
approved

spec/migrations/20170928202851_migrate_middleware_server_to_wildfly_and_eap_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      migrate
+      expect(mw_server.reload).to have_attributes(
+        :type => "ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServer"
+      )
+    end
+  end
+
+  migration_context :down do
+    it "reverts type for WildFly" do
+      mw_server = mw_server_stub.create!(:type => "ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerWildfly")
```
cfcosta: Both local assignments should be extracted to `let`, and the migration to be run in a `before` filter.

db/migrate/20170928202851_migrate_middleware_server_to_wildfly_and_eap.rb
cfcosta: I think they are better included in middleware_manager as well, so that we have proper namespacing for the resources, what do you think?

Another option would be `ManageIQ::Providers::Hawkular::WildFlyServer`. Any of the two are fine by me.

spec/migrations/20170928202851_migrate_middleware_server_to_wildfly_and_eap_spec.rb
tumido: Can you please point me to an example? I can't find any other migration which would be using it... Or maybe I just don't understand the suggestion.

tumido: @cfcosta, does it make any difference here to move two assignments from test cases into two `let`s?

cfcosta: Just checked on other migration tests, it seems like all of them are done the way you did them, so :ship: it.

abonas:
@Fryguy @chessbyte lets get this in b4 the feature ❄️

PR #83, opened 2017-10-02
Add user_id,group_id,tenant_id to MiqQueue

mkanoor:
This is needed so that the Queue can run commands in the context of the user, group and tenant.

### DIFF
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -899,8 +899,8 @@ container_templates:
 - objects
 - created_at
 - updated_at
-- type
 - object_labels
+- type
 container_volumes:
 - id
 - parent_id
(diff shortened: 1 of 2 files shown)


mkanoor:
@bdunne @Fryguy 
Please review

db/schema.yml
```diff
@@ -899,8 +899,8 @@ container_templates:
 - objects
 - created_at
 - updated_at
-- object_labels
 - type
+- object_labels
```
Fryguy: This should not happen.

Fryguy:
@kbrock Please review.

db/schema.yml
bdunne: Maybe the migrations were run out of order?

PR #87, opened 2017-10-09
Remove the configuration_templates table

skovic:
This PR removes the configuration_templates table because it is no longer used.


skovic:
@Fryguy @bdunne The PR that removes the old table is PR #87

db/migrate/20171009173946_remove_configuration_templates_table.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveConfigurationTemplatesTable < ActiveRecord::Migration[5.0]
+  def change
+    drop_table :configuration_templates do |t|
```
Fryguy: COOL I've never seen a drop_table with a block like this so that `change` will work 👍

PR #89, opened 2017-10-09
Add the api attributes spec

carbonin:
Moved from:
https://github.com/ManageIQ/manageiq/blob/53b733b7a254f5677cd26ee7ccd5a7c3c879a63a/spec/replication/util/api_attributes_spec.rb


carbonin:
Removed from ManageIQ/manageiq in https://github.com/ManageIQ/manageiq/pull/16157

PR #91, opened 2017-10-11
Move Openstack refresher settings

mansam:


db/migrate/20171011180000_move_openstack_refresher_settings.rb
```diff
@@ -0,0 +1,13 @@
+class MoveOpenstackRefresherSettings < ActiveRecord::Migration[5.0]
+  class SettingsChange < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time('Move Openstack provider refresher settings') do
+      SettingsChange.where(:key => '/ems/ems_openstack/refresh/inventory_object_refresh').update(:key => '/ems/ems_refresh/openstack/inventory_object_refresh')
```
Fryguy: Prefer update_all over update as it will go direct to the DB instead of instantiating one-by-one.

mansam: Cool. The changes aren't all exactly 1-1 so I'm going to stick with the explicit calls, if that's alright with you.

PR #93, opened 2017-10-13
Add sub_metric column to chargeback_rate_detail

lpichler:
This is for implementation for chargeback of storage (cloud volume) type.

Metric for allocated storage is now divided according to storage types and we need to store cloud volume type for the ChargebackRateDetail (the be able get metric value and then calculate cost)

So generally: it is adding posibility to divide metric to the small pieces - so this is the why I am calling it 'sub metric'.

In rate editor:

screen shot 2017-10-13 at 15 31 41

(gp2, .. - are cloud volume storage types)

cc @gtanzillo 

@miq-bot assign @Fryguy


lpichler:
@Fryguy

PR #94, opened 2017-10-16
Add user_id,group_id,tenant_id to EventStream.

lfu:
Add user_id,group_id,tenant_id to EventStream.

User wants to get the event initiator from MiqEvent and uses it in automate code.
Blocks https://github.com/ManageIQ/manageiq/pull/16179.

https://bugzilla.redhat.com/show_bug.cgi?id=1487749

cc @kbrock @Fryguy @mkanoor

### DIFF
--- a/db/migrate/20171016192352_add_user_to_event_stream.rb
+++ b/db/migrate/20171016192352_add_user_to_event_stream.rb
@@ -1,0 +1,7 @@
+class AddUserToEventStream < ActiveRecord::Migration[5.0]
+  def change
+    add_column :event_streams, :user_id, :bigint
+    add_column :event_streams, :group_id, :bigint
+    add_column :event_streams, :tenant_id, :bigint
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -1253,6 +1253,9 @@ event_streams:
 - ems_ref
 - middleware_domain_id
 - middleware_domain_name
+- user_id
+- group_id
+- tenant_id
 ext_management_systems:
 - id
 - name


lfu:

mkanoor:
👍

kbrock:
Do wish we did not have user_id in event - and wish it could be derived in the receiver (since an event from vmware would not actually have a user_id on it)

But, this seems like a good way to make sure that a user is passed through the system. (rather than the bogus `"system"` phrase)

:+1:

Fryguy:
My concern here is that storing user_ids typically has complications for the global region and we usually store the user "name" in those cases.  However, I don't think event_streams are replicated, so maybe it doesn't matter

cc @gtanzillo @gmcculloug Thoughts?

lfu:
@Fryguy In my understanding ```user_id``` is unique within and across the region while ```userid``` is unique within the region but not across the region. So ```user_id``` seems a better choice than ```userid```. What do I miss?

Fryguy:
The problem is that "jsmith" in region 1 will have an id of 1_000_000_000_001, but in region 99 his id will be 99_000_000_000_001.  If we replicate region 1 *into* region 99, then his name exists twice, and then you might need to merge his events from both regions.  If you link by id only, then it can only see his events from a particular region, while if you link by userid, then you can see all of them.  Since we do replicate events (they're not in the default_replication_exclude_tables.yml), I want @gtanzillo and @gmcculloug to approve first before merge.

gtanzillo:
I'm good with this. Discussing with @gmcculloug, automate needs to reference the actual user of the event and not necessarily the user with the same name in the global region. Also, `event_streams` is replicated. But, we can always use the `user_id` to look up the actual user and the find the one with the same username in the global if necessary. We've done that in other places.

gmcculloug:
@Fryguy Please review/merge.

PR #95, opened 2017-10-16
Get rid of the condition modifier which is not needed and confusing.

lfu:
Blocks https://github.com/ManageIQ/manageiq/pull/16213.

https://bugzilla.redhat.com/show_bug.cgi?id=1499161

### DIFF
--- a/db/migrate/20171031200739_drop_condition_modifier.rb
+++ b/db/migrate/20171031200739_drop_condition_modifier.rb
@@ -1,0 +1,31 @@
+class DropConditionModifier < ActiveRecord::Migration[5.0]
+  class Condition < ActiveRecord::Base
+    def reverse_expression(migration)
+      val = expression
+      if val.include?("!ruby/object:MiqExpression")
+        val.sub!(/MiqExpression/, 'Hash')
+      else
+        migration.say("#{self.class} Id: #{id} does not have an MiqExpression, skipping conversion")
+        return
+      end
+
+      raw_hash = YAML.safe_load(val)
+      raw_hash["exp"] = { "not" => raw_hash["exp"]}
+      new_hash = YAML.dump(raw_hash).sub(/---/, "--- !ruby/object:MiqExpression")
+      update_attributes(:expression => new_hash)
+    end
+  end
+
+  def up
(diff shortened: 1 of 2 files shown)


Fryguy:
Seems ok to me, but I think this has to merged in tandem with ManageIQ/manageiq#16213.

db/migrate/20171031200739_drop_condition_modifier.rb
```diff
@@ -0,0 +1,30 @@
+class DropConditionModifier < ActiveRecord::Migration[5.0]
+  class Condition < ActiveRecord::Base
+    def reverse_expression(val, migration)
+      if val.include?("!ruby/object:MiqExpression")
+        val.sub!(/MiqExpression/, 'Hash')
+      elsif val.starts_with?('--- !')
```
bdunne: Should this just be `else`?

bdunne: If that's the case, maybe the condition should be reversed so that we return early

spec/migrations/20171031200739_drop_condition_modifier_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      condition_changed = condition_stub.create!(:modifier => 'deny', :expression => expression)
+      condition_ignored = condition_stub.create!(:modifier => 'allow', :expression => expression)
+
+      migrate
+
+      expect(condition_changed.reload.expression).to eq(expected_expression)
+      expect(condition_ignored.reload.expression).to eq(expression)
+    end
+  end
```
bdunne: Can you add a test for the non-MiqExpression case?

spec/migrations/20171031200739_drop_condition_modifier_spec.rb
```diff
@@ -0,0 +1,34 @@
+require_migration
+
+describe DropConditionModifier do
+  let(:condition_stub) { migration_stub(:Condition) }
+
+  migration_context :up do
+    it "reverse the expression if :modifier = 'deny'" do
+      expression = <<-EOS
```
bdunne: `<<~EOS` will allow you to indent the heredocs

PR #98, opened 2017-10-18
Convert containers hawkular Endpoints port=nil to port=443

cben:
For a time (before https://github.com/ManageIQ/manageiq-ui-classic/pull/1172), it was possible in UI to save containers provider with blank hawkular port, which got saved as port = nil.
Then it'd be impossible to edit the provider, as UI would crash:

manageiq containers providers port nil

On fine this was fixed by https://github.com/ManageIQ/manageiq-ui-classic/pull/1565 adding special case in UI to tolerate nil.
On master, @AparnaKarve [asked](https://github.com/ManageIQ/manageiq-ui-classic/pull/1305#issuecomment-309489658) to not introduce special cases but do a migration.
Better late than never :-)

After investigating, @yaacov @moolitayer and me concluded such provider does connect to hawkular with port 443 (it builds a url without port, confirmed experimentally).
This migration normalizes such endpoints to port 443, works same but possible to edit in UI:

manageiq containers providers port 443

Tested by checking out fine-1, creating such provider, migrating all the way to master, it was still broken, and fixed by this migration.  That's how I took above screenshots.
(in some cases such endpoints might have already been deleted by https://github.com/ManageIQ/manageiq/pull/14990 migration.  But not if hostname was autodetected from route => this is still sometimes needed.)

https://bugzilla.redhat.com/show_bug.cgi?id=1432070

@miq-bot add-label bug, data
@moolitayer @AparnaKarve please review

### DIFF
--- a/db/migrate/20171030131403_fix_hawkular_endpoints_with_port_nil.rb
+++ b/db/migrate/20171030131403_fix_hawkular_endpoints_with_port_nil.rb
@@ -1,0 +1,25 @@
+class FixHawkularEndpointsWithPortNil < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Endpoint < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    container_hawkular_endpoints.where(:port => nil).update_all(:port => 443)
+  end
+
+  def container_hawkular_endpoints
+    ems_container_ids = ExtManagementSystem.where(
+      :type => %w(ManageIQ::Providers::Openshift::ContainerManager ManageIQ::Providers::Kubernetes::ContainerManager)
+    ).pluck(:id)
+
+    Endpoint.where(
+      :resource_type => 'ExtManagementSystem',
+      :resource_id   => ems_container_ids,
+      :role          => 'hawkular'
+    )
+  end
+end
--- a/spec/migrations/20171030131403_fix_hawkular_endpoints_with_port_nil_spec.rb
+++ b/spec/migrations/20171030131403_fix_hawkular_endpoints_with_port_nil_spec.rb
@@ -1,0 +1,65 @@
+require_migration
+
+describe FixHawkularEndpointsWithPortNil do
+  let(:ext_management_system_stub) { migration_stub(:ExtManagementSystem) }
+  let(:endpoint_stub) { migration_stub(:Endpoint) }
+
+  migration_context :up do
+    it 'Modifies hawkular endpoints with nil port' do
+      ems = ext_management_system_stub.create!(
+        :name => 'container',
+        :type => 'ManageIQ::Providers::Openshift::ContainerManager'
+      )
+      hawk = endpoint_stub.create!(
+        :role          => "hawkular",
+        :hostname      => "hawkname",
+        :port          => nil,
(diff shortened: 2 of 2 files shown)


moolitayer:
> This migration normalizes such endpoints to port 0, still invalid but possible to edit in UI:

@cben what if we remove these endpoints altogether? that way we will not have failed authentication checks for an endpoint the user did not intend to create (event if the user does not go into the edit screen and disables metrics altogether)

cben:
These endpoints were only possible by actually going into Hawkular tab and
clearing the port field, not sure it's "didn't intend to create" but never
mind.
Yeah, I'm fine with deleting too.  Don't have an opinion, please decide for
me :)  cc @yaacov @zeari
Deleting will result in UI showing "Metrics: disabled" in Provider Edit
right?
In provider details, will it show as metrics error (ideal IMHO), or not
show metrics at all?

yaacov:
> it was possible in UI to save containers provider with blank hawkular port, which got saved as port = nil.

a. This is fine, when `port==nil` we fallback to default port [1] , IMHO, if you want to migrate it, it should be to the default port.

b. In old versions of ManageIQ it was not possible to have no metrics, metrics where always on, and values fall-back to default url. port and token if not given explicitly. IMHO migration of missing endpoint should be to a default values, and to nil/0.

[1]  it was possible in UI to save containers provider with blank hawkular port, which got saved as port = nil.

Fryguy:
This looks good to me, I just want approval from @simon3z or @moolitayer before merge.

simon3z:
> a. This is fine, when port==nil we fallback to default port [1] , IMHO, if you want to migrate it, it should be to the default port.

+1 for what @yaacov is suggesting here.
@cben what do you think?

> b. In old versions of ManageIQ it was not possible to have no metrics, metrics where always on, and values fall-back to default url. port and token if not given explicitly. IMHO migration of missing endpoint should be to a default values, and to nil/0.

This one is actually sad, but I don't have any other good suggestions here.

moolitayer:
> These endpoints were only possible by actually going into Hawkular tab and clearing the port field, not sure it's "didn't intend to create" but never mind.

I thought this was a situation where the user had pressed add from the `default` endpoint tab
I'm good with @yaacov's suggestion in this case :+1: 
I was really surprised that fine screens allowed that... (I checked that master blocks it so I can sleep at night)

> Deleting will result in UI showing "Metrics: disabled" in Provider Edit right?

Right

cben:
Confirmed experimentally, such endpoint is not broken as I originally thought but accesses https URL with port omitted, so it uses port 443.  This is also what @yaacov @moolitayer and me saw in code.

=> Changed migration to set port 443.  Dropped down migration, will not change 443 into nil.
Updated PR description and screenshot.  PTAL.

cben:
@yaacov PTAL

yaacov:
what about the update_all , don't you want to fix that first ?

o/w LGTM 👍

cben:
you mean miq-bot?  IIUC we routinely ignore this advice in migrations, because we want migrations to be fast where feasible, and because the model stubs don't normally have validations anyway.

@Fryguy please review (nice to have but not blocker)

Fryguy:
yeah, ignore the update_all warning (I keep forgetting to change that in the rubocop.yml)

@moolitayer You good with this?

chessbyte:
@simon3z @moolitayer can one of you guys approve this PR?

PR #100, opened 2017-10-19
[WIP] Add missing index on handler_id and type

jrafanie:
https://bugzilla.redhat.com/show_bug.cgi?id=1504287

Queries on a 1 million+ row miq_queue table using handlers is slow due
to a missing index.

miq_worker.rb has various messages methods that scope the miq_queue where
it is the "handler",
see: https://github.com/ManageIQ/manageiq/blob/aba53d3d633fe4af7a59a44e3b7058e343d92c2c/app/models/miq_worker.rb#L9-L12

When a server is not responding, the master server will deactivate this
server and destroy any activate queue messags being worked on by workers
on this not responding server, here:
https://github.com/ManageIQ/manageiq/blob/aba53d3d633fe4af7a59a44e3b7058e343d92c2c/app/models/miq_server/server_monitor.rb#L15

Additionally, when we want to destroy worker rows, we hit the
before_destroy, calling log_destroy_of_worker_messages here:
https://github.com/ManageIQ/manageiq/blob/aba53d3d633fe4af7a59a44e3b7058e343d92c2c/app/models/miq_worker.rb#L6

This method loops through ready_messages and processed_messages for this
worker, logs a message and cleans them up here:
https://github.com/ManageIQ/manageiq/blob/aba53d3d633fe4af7a59a44e3b7058e343d92c2c/app/models/miq_worker.rb#L455-L464

Summary: 1.4 million row no data miq_queue with 62 workers goes from
16.5 seconds to 4.2 seconds to loop through the workers and
clean_active_messages on them.

Before:
```
$ bin/rails r 'time1 = Time.now; MiqServer.all.each { |s| s.miq_workers.all.each(&:clean_active_messages)}; puts Time.now - time1; puts "Messages: #{MiqQueue.count}"; puts "workers: #{MiqWorker.count}"'
** override_gem: manageiq-schema, [{:path=>"/Users/joerafaniello/Code/manageiq-schema"}], caller: /Users/joerafaniello/Code/manageiq/bundler.d/Gemfile.dev.rb
** Using session_store: ActionDispatch::Session::MemCacheStore
16.503509
272
Messages: 1400000
workers: 62
```

```
$ ~/Code/manageiq (master) (2.4.2) + bin/rake db:migrate
== 20171019173933 AddIndexOnTypeAndIdOnMiqQueue: migrating ====================
-- add_index(:miq_queue, [:handler_id, :handler_type])
   -> 2.8617s
== 20171019173933 AddIndexOnTypeAndIdOnMiqQueue: migrated (2.8618s) ===========
```

After:
```
$ bin/rails r 'time1 = Time.now; MiqServer.all.each { |s| s.miq_workers.all.each(&:clean_active_messages)}; puts Time.now - time1; puts "Messages: #{MiqQueue.count}"; puts "workers: #{MiqWorker.count}"'
4.275245
Messages: 1400000
workers: 62
```

### DIFF
--- a/db/migrate/20171019173933_add_index_on_type_and_id_on_miq_queue.rb
+++ b/db/migrate/20171019173933_add_index_on_type_and_id_on_miq_queue.rb
@@ -1,0 +1,5 @@
+class AddIndexOnTypeAndIdOnMiqQueue < ActiveRecord::Migration[5.0]
+  def change
+    add_index :miq_queue, [:handler_id, :handler_type]
+  end
+end


jrafanie:
@Fryguy @gtanzillo please review

Fryguy:
@kbrock Please review.

Fryguy:
Only 16 seconds to 4?  How is this a problem at all, then?

jrafanie:
> Only 16 seconds to 4? How is this a problem at all, then?

@Fryguy My miq_queue table was populated with nearly 1.4 million empty rows as I wasn't populating the args column with any data.  My data wasn't a great recreation of the problem but it did expose some slowness.  I could try a different database export if it makes sense.  I'll just have to find one.

kbrock:
rant:
> Sure looking forward to deleting the `handler` column when we go to a different queue.

Due to high throughput, let's try and keep indexes to the minimum here.
An extra index will slow down every insert and delete since it will need to be updated.
But a slow delete is not good either. It could cause contention (although I am only semi-certain on this one)

This will happen when we kill a worker and when a new worker starts up. Right?
Do these deletes really need to happen inline?
Would it be too "meta" to run this as a generic miq queue task?

I'm leaning towards creating the index, but I just wonder if the best way to speed this up is to "not run it" - or at least not in a high priority section of code.

jrafanie:
@kbrock I agree, I'm leaning towards using an index here but am not 100% convinced of it's need.  I wish I could reliably recreate the many hour hang and subsequent memory growth.

> This will happen when we kill a worker and when a new worker starts up. Right?

This happens when a server is marked as not responding and we deactivate it's roles and remove any queue messages being processed by workers on that server.

> Do these deletes really need to happen inline?

I believe we could delay the removal of these message items.

It feels like we shouldn't be doing 
[miq_workers.each(&:clean_active_messages)](https://github.com/ManageIQ/manageiq/blob/aba53d3d633fe4af7a59a44e3b7058e343d92c2c/app/models/miq_server/server_monitor.rb#L15) and should be instead use a delete from miq_queue where handler_id in .... (and build a list of worker ids) but in the case of the logs provided, we don't even get to instantiating the active message rows, we try to find the first worker's active messages and it hangs for hours.

jrafanie:
@miq-bot add_label wip

jrafanie:
I'll try to get a better representation of a miq_queue table and test this change with it.

kbrock:
@jrafanie I'm confused why we didn't just merge this?

jrafanie:
@kbrock I didn't get a reasonable test demonstrating this was needed and didn't really have a strong case to dig deeper.

PR #101, opened 2017-10-20
Add jsonb data into miq_queue

Ladas:
Add jsonb data into miq_queue, it will allow much more
effective way of adding targets into MiqQueue record


Ladas:
closing in favor of https://github.com/ManageIQ/manageiq-schema/pull/102

PR #102, opened 2017-10-23
Add array data into miq_queue

Ladas:
Add array data into miq_queue, it will allow much more
effective way of adding targets into MiqQueue record


Ladas:
@miq-bot assign @Fryguy

Ladas:
closing in favor of https://github.com/ManageIQ/manageiq/pull/16280

PR #103, opened 2017-10-23
Ensure EMS relationships are established

djberg96:
This addresses an issue, originally on the Amazon provider side, where the EMS relationships were not established when a StorageManager was added.

Originally brought to our attention via https://bugzilla.redhat.com/show_bug.cgi?id=1480629, but 5.7 -> 5.8 will be handled with a `self.seed` method.

### DIFF
--- a/db/migrate/20171023170841_ensure_cloud_managers.rb
+++ b/db/migrate/20171023170841_ensure_cloud_managers.rb
@@ -3,41 +3,45 @@ class EnsureCloudManagers < ActiveRecord::Migration[5.0]
   class BaseManager < ExtManagementSystem; end
   class Settings < ActiveRecord::Base; end
 
-  class CloudManager < ActiveRecord::Base
-    def ensure_managers
-      build_network_manager unless network_manager
-      network_manager.name = "#{name} Network Manager"
+  class ManageIQ
+    class Providers
+      class CloudManager < ActiveRecord::Base
+        def ensure_managers
+          build_network_manager unless network_manager
+          network_manager.name = "#{name} Network Manager"
 
-      build_ebs_storage_manager unless ebs_storage_manager
-      ebs_storage_manager.name = "#{name} EBS Storage Manager"
+          build_ebs_storage_manager unless ebs_storage_manager
+          ebs_storage_manager.name = "#{name} EBS Storage Manager"
 
-      if Settings.prototype.amazon.s3
-        build_s3_storage_manager unless s3_storage_manager
-        s3_storage_manager.name = "#{name} S3 Storage Manager"
-      end
+          if Settings.prototype.amazon.s3
+            build_s3_storage_manager unless s3_storage_manager
+            s3_storage_manager.name = "#{name} S3 Storage Manager"
+          end
 
-      ensure_managers_zone_and_provider_region
-    end
+          ensure_managers_zone_and_provider_region
+        end
 
-    def ensure_managers_zone_and_provider_region
-      if network_manager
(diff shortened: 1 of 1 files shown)


Fryguy:
The migrations will need a stub with a copy of the ensure_managers method.  As migrations are a point-in-time snapshot, we cannot have changes to the underlying ensure_managers method from affecting old migrations.  Thus, we need a copy.

db/migrate/20171023170841_ensure_cloud_managers.rb
```diff
@@ -0,0 +1,5 @@
+class EnsureCloudManagers < ActiveRecord::Migration[5.0]
+  def change
```
Fryguy: This needs to be up, because it is not a reversible operation.

bronaghs:
@djberg96 - bump ^^
Migrations have to be merged by Monday 10/30

djberg96:
@Fryguy I don't know if I did it the "expected" way, but please take a look and see what you think. I can squash the commits if you're happy with it.

Fryguy:
all data migrations require specs as well

db/migrate/20171023170841_ensure_cloud_managers.rb
```diff
@@ -0,0 +1,47 @@
+class EnsureCloudManagers < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base; end
+  class BaseManager < ExtManagementSystem; end
+  class Settings < ActiveRecord::Base; end
+
+  class ManageIQ
+    class Providers
+      class CloudManager < BaseManager
```
Fryguy: This should be modifying the stub model created above.

db/migrate/20171023170841_ensure_cloud_managers.rb
```diff
@@ -0,0 +1,47 @@
+class EnsureCloudManagers < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base; end
+  class BaseManager < ExtManagementSystem; end
+  class Settings < ActiveRecord::Base; end
```
Fryguy: I don;t know why Settings is needed...Settings is not even an ActiveRecord model in the manageiq app.

db/migrate/20171023170841_ensure_cloud_managers.rb
```diff
@@ -0,0 +1,47 @@
+class EnsureCloudManagers < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base; end
```
Fryguy: This needs the 

```ruby
self.inheritance_column = :_type_disabled # disable STI
```

db/migrate/20171023170841_ensure_cloud_managers.rb
djberg96: @Fryguy Sorry, I don't follow.

db/migrate/20171023170841_ensure_cloud_managers.rb
djberg96: Settings (originally referenced as `::Settings`) is referred to within the `ensure_managers` method. I've modified it so that it's not a subclass of AR any more.

db/migrate/20171023170841_ensure_cloud_managers.rb
djberg96: Done.

chessbyte:
Replaced by #117

PR #107, opened 2017-10-25
Migrate EmsRefresh.refresh queue args to data

agrare:
Move existing EmsRefresh.refresh queue args to data for performance.

Related to: https://github.com/ManageIQ/manageiq/pull/16271

### DIFF
--- a/spec/migrations/20171025122732_move_ems_refresh_args_to_data_spec.rb
+++ b/spec/migrations/20171025122732_move_ems_refresh_args_to_data_spec.rb
@@ -1,0 +1,88 @@
+require_migration
+
+describe MoveEmsRefreshArgsToData do
+  let(:miq_queue_stub) { migration_stub(:MiqQueue) }
+  let(:targets) { [['Vm', 1], ['Host', 2]] }
+  let(:refresh_queue_options) do
+    {
+      :class_name  => 'EmsRefresh',
+      :method_name => 'refresh',
+      :role        => 'ems_inventory',
+      :queue_name  => 'ems_1',
+      :zone        => 'default',
+    }
+  end
+  let(:refresh_new_target_queue_options) do
+    {
+      :class_name  => 'EmsRefresh',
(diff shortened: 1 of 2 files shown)


Fryguy:
As discussed, since Marshal.dump/load is not guaranteed to work across Ruby versions, we should bracket this around upgrades and just delete rows that can't be Marshaled (which if one fails they all will likely fail).

We may want to be resilient on the MiqQueue side as well...not sure we ever caught that error there.

agrare:
@Fryguy added tests with msg_data containing an empty array with marshal format version `3.0`

spec/migrations/20171025122732_move_ems_refresh_args_to_data_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    it "If there are no args" do
+      q_item = miq_queue_stub.create!(refresh_queue_options.merge(:msg_data => nil))
+
+      migrate
+
+      expect(q_item.reload.args).to match_array([])
+    end
+
+    it "Handle invalid marshal format errors" do
+      # "\x03\x00\x00" is an empty array in Marshal version 3.0
```
Fryguy: Is there a version of this for the up migration?

agrare: I wasn't sure how to test `Marshal.dump` failing due to a version issue, ideas?

agrare: Let me try putting an anonymous class in the args, that should do it.

agrare: Scratch that, that won't even serializing the value to args

Fryguy: Oh I see...it's only the load part that fails.  :+1:

PR #108, opened 2017-10-26
Add an 'Allocated cpu cores' chargeback rate detail to existing chargeback rates

zeari:
Adds the rate detail used in https://github.com/ManageIQ/manageiq/pull/15616 to existing chargeback rates
@lpichler Please review
cc @moolitayer @cben @simon3z

### DIFF
--- a/db/migrate/20171026103833_add_cores_allocated_rate_detail.rb
+++ b/db/migrate/20171026103833_add_cores_allocated_rate_detail.rb
@@ -1,0 +1,43 @@
+class AddCoresAllocatedRateDetail < ActiveRecord::Migration[5.0]
+  class ChargebackRate < ActiveRecord::Base
+    has_many :chargeback_rate_details, :class_name => 'AddCoresAllocatedRateDetail::ChargebackRateDetail'
+  end
+
+  class ChargebackTier < ActiveRecord::Base
+    FORM_ATTRIBUTES = %i(fixed_rate variable_rate start finish).freeze
+  end
+
+  class ChargeableField < ActiveRecord::Base; end
+
+  class ChargebackRateDetail < ActiveRecord::Base
+    belongs_to :chargeback_rate, :class_name => 'AddCoresAllocatedRateDetail::ChargebackRate'
+    belongs_to :chargeable_field
+    has_many   :chargeback_tiers, :class_name => 'AddCoresAllocatedRateDetail::ChargebackTier'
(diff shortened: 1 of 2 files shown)


db/migrate/20171026103833_add_cores_allocated_rate_detail.rb
```diff
@@ -0,0 +1,27 @@
+class AddCoresAllocatedRateDetail < ActiveRecord::Migration[5.0]
+  def up
```
lpichler: @zeari 

need to ad stub
```
  class ChargeableField < ActiveRecord::Base
  end
```

Fryguy:
Doesn't seeding handle this?

lpichler:
@Fryguy no, this is needed for existing `ChargebackRate`s. Seeding will just provide ability to use this new metric when new ChargebackRate is added.

spec/migrations/20171026103833_add_cores_allocated_rate_detail_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  migration_context :up do
+    it 'Adds a "Allocated CPU Cores" rate detail to existing details' do
+      rate_details.each do |field|
+        detail_stub.create!(field.merge(:chargeback_rate_id => default_rate.id))
+        detail_stub.create!(field.merge(:chargeback_rate_id => custom_rate.id))
+      end
+
+      migrate
+
+      expect(default_rate.chargeback_rate_details.where(:description => "Allocated CPU Cores").count).to eq(1)
```
Fryguy: Since the default_rate object is memoized from the call on line 25, can you change this to a `default_rate.reload.chargeback_rate_details...`?   This way we can ensure it's getting fresh data.

spec/migrations/20171026103833_add_cores_allocated_rate_detail_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(custom_rate.chargeback_rate_details.where(:description => "Allocated CPU Cores").count).to eq(1)
```
Fryguy: Same

zeari:
>Since the default_rate object is memoized from the call on line 25, can you change this to a default_rate.reload.chargeback_rate_details...? This way we can ensure it's getting fresh data.

Done @Fryguy

PR #110, opened 2017-10-26
Add deleted_on (with index) to container_quotas, container_quota_items

cben:
We need to store history of quotas during project lifetime.
Reporting needs unclear yet, for now just want to store it.

As discussed with @gtanzillo @bazulay @zeari, we are going to (ab)use archiving to store history in same tables — not just when user deletes quota in kubernetes but also when it's changed!

(Not yet clear if when only a container_quota_item changes, we'll archive and create only new item, or archive whole quota "snapshot" and create new quota with children.  Want to keep door open to both designs in z-streams, so need them in both tables...)

Added same indexes on `deleted_on` as other archived container items (#18).

## Impact on table churn/size

There are O(1) container_quotas rows per container_project, and most users change them rarely and manually.  That's why we'll store these only on change, not periodically as in metrics/rollups/vim_performance_states.

However `container_quota_items.quota_observed` column may change each refresh.
At this stage, we don't plan to archive when solely `quota_observed` changes, but that may change.

=> I don't expect these tables to become huge.  Should stay significantly smaller than `container_groups`.

(P.S. quota scopes coming in separate PR.)

@zeari @gtanzillo @Fryguy please review, want to get at least schema before freeze.

### DIFF
--- a/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
+++ b/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
@@ -1,0 +1,10 @@
+class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
+  def change
+    add_column :container_quotas, :deleted_on, :datetime
+    add_index :container_quotas, :deleted_on
+
+    add_timestamps :container_quota_items
+    add_column :container_quota_items, :deleted_on, :datetime
+    add_index :container_quota_items, :deleted_on
+  end
+end
--- a/db/schema.yml
+++ b/db/schema.yml
@@ -806,6 +806,9 @@ container_quota_items:
 - quota_desired
 - quota_enforced
 - quota_observed
+- created_at
+- updated_at
+- deleted_on
 container_quotas:
 - id
 - name
@@ -815,6 +818,7 @@ container_quotas:
 - container_project_id
 - ems_id
 - created_on
+- deleted_on
 container_replicators:
 - id
 - ems_ref


db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
```diff
@@ -0,0 +1,10 @@
+class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
+  def change
+    add_column :container_quotas, :deleted_on, :datetime
+    add_index :container_quotas, :deleted_on
+
+    add_column :container_quota_items, :created_on, :datetime
```
Fryguy: This will be a problem because `created_on` is the old name for the built-in ActiveRecord timestamps, and ActiveRecord still supports that name, thus ActiveRecord will auto-fill this column with Time.now when you create the record.

cben: That's what I had in mind, the standard created_on.  most containers tables (including container_quotas) already have this, items didn't...
Should I prefer `created_at`?

Do you see reason a custom named, manually filled column instead?  I just tried and I can override `created_on` if I'll really need.

cben: (the reason to want both created & deleted is we might want to answer "what were the quotas at moment T". I imagine we'd do something like `created_on <= T AND T < deleted_on`...)

Fryguy: So, if you want to store the "native" creation date, you probably want a different column that created_at/created_on as those are reserved as "the date of when the row in the database was created".  If the other containers tables are inserting the "native" creation data into that column, then it's technically wrong.  We really should not override built-in ActiveRecord things, as it only leads to confusion later.

If you still want to create a created_at column, prefer using the built-in .timestamps method in migrations over adding the column manually

```ruby
change_table :container_quota_items do |t|
  t.timestamps
end

cben: Ah, I see what you mean.
AFAIK we don't currently override any AR timestamps.
We do get some "native" dates from k8s and stuff those into our own columns like `container_quotas.ems_created_on`, `containers.finished_at` etc. :heavy_check_mark: 

"created_on/deleted_on" in other archived container entities already means "creation/archival *persisted* at" (creation set by AR magic ts / `deleted_on = Time.now`), and we were fine with that.
Here for quotas we'll also abuse "deleted/created_on" to mean "persisted change at" and not just "persisted creation/deletion".  But all 3 are approx refresh timestamps...

In theory at some point we might prefer "collected on" timestamps as closer to truth (esp. with watch worker — @Ladas does watch give us real changed-in-k8s timestamp?), but that'd be a global move, would need our own columns on many tables, and out of scope here.

- However, yes, I do feel a temptation to achieve "subversion-like" history where several items changed in same refresh get *exactly* same timestamp, so as you do "svn log" you'll see an "atomic change".
  I'm trying to resist it :-)
  More seriously, *if* we'll want this, I think the sane design is not syncronized timestamps but archive only `container_quota`, and attach brand new items & scopes under it (items & scopes deleted_on would be unused in this scenario).

Need to head home, I'll rewrite to `t.timestamps`.

cben:
Done, PTAL.
`add_timestamps` also added `container_quota_items.updated_at`, unnecessary and a bit misleading as when it's all implemented, we'll *never* update in that table.  Well, I'm lying, setting `deleted_on` will be a row update.
But I don't mind if you don't mind ;)

Fryguy:
oh cool I didn't know about add_timestamps :+1:

PR #111, opened 2017-10-26
Create container_quota_scopes table

cben:
https://kubernetes.io/docs/concepts/policy/resource-quotas/#quota-scopes
We don't yet store them, but should.
ContainerQuota has_many scopes, each a string (currently enum with 4 possible values `Terminating`, `NotTerminating`, `BestEffort`, `NotBestEffort`).

Didn't see any reason to choose clever serialized/json representation so went with standard has_many to new table.

See #110 for motivation having created_at/deleted_on.

### DIFF
--- a/db/migrate/20171026142653_create_container_quota_scopes.rb
+++ b/db/migrate/20171026142653_create_container_quota_scopes.rb
@@ -1,0 +1,10 @@
+class CreateContainerQuotaScopes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :container_quota_scopes do |t|
+      t.bigint :container_quota_id
+      t.string :scope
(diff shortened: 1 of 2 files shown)


cben:
@zeari @gtanzillo @Fryguy please review, want to get at least schema before freeze.

cben:
rebased, switched to standard `timestamps`

cben:
Note to self: I forgot to index `container_quota_scopes` on `deleted_on` field. (`container_quotas` and `container_quota_items` do have this index)

This could slow down refresh (and display) if there will be tons of archived rows.
Luckily, turns out scopes are immutable for parent container_quota's lifetime, so I'm not planning to use archiving in this table at all :-)

PR #113, opened 2017-10-26
Drop blueprints and blueprint_id in dialogs and service_templates

bzwei:
The usage has been removed in API and core model. This is the last step to clean up in schema.

https://github.com/ManageIQ/manageiq-api/pull/139
https://github.com/ManageIQ/manageiq/pull/16268


bzwei:
@miq-bot add_label technical debt
@miq-bot assign @Fryguy

gmcculloug:
@bzwei You also need to modify the `db/schema.yml` file based on these changes.

See https://github.com/ManageIQ/manageiq-schema/pull/87 for an example.

PR #115, opened 2017-10-29
Retroactive fix to broken #110

cben:
So #110 migration `20171026114327_add_deleted_on_to_container_quota_and_items` failed for people that had container providers :-(
[`add_timestamps` sets NOT NULL](https://makandracards.com/makandra/19325-your-database-tables-should-always-have-timestamps), which existing records violate:

    PG::NotNullViolation: ERROR:  column "created_at" contains null values

It was also semantically wrong, we'll want all records to have `created_at` timestamps so we can rely on `created_at`..`deleted_on` ranges in future reports.

This PR must converge to same outcome in two scenarios:

Timeline Z
----------

Some people with zero records in `container_quota_items` successfully run original #110 migration and will not re-execute changes I do here.

They end up with:

    t.datetime "created_at",         null: false
    t.datetime "updated_at",         null: false

and zero records, future records will get both timestamps filled.

Timeline Q
----------

People that had 1 or more `container_quota_items` got stuck before the broken migration ("An error has occurred, this and all later migrations canceled").
**I can't just append a new migration because these people can't get past the broken one.
They will re-execute these changes once they're merged.**

@miq-bot add-label bug
@zeari @zakiva @Fryguy please review (carefully!)

- [x] added tests for original breakage
- [x] run the two scenarios on separate DBs and compared resulting schema.rb — same

### DIFF
--- a/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
+++ b/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
@@ -1,12 +1,31 @@
 class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
   class ContainerQuota < ActiveRecord::Base; end
-  class ContainerQuotaItem < ActiveRecord::Base; end
+  class ContainerQuotaItem < ActiveRecord::Base
+    belongs_to :container_quota, :class_name => "AddDeletedOnToContainerQuotaAndItems::ContainerQuota"
+  end
 
   def change
     add_column :container_quotas, :deleted_on, :datetime
     add_index :container_quotas, :deleted_on
 
-    add_timestamps :container_quota_items
+    add_timestamps :container_quota_items, :null => true # temporarily for backfilling, then disallow nulls
+    reversible do |dir|
+      dir.up do
+        say_with_time("Backfilling container_quota_items timestamps from parent quotas") do
+          # All quota items SHOULD have a parent quota, which SHOULD have deleted_on
+          # but just in case, fallback to current time.
+          now = Time.zone.now
+          ContainerQuotaItem.includes(:container_quota).find_each do |item|
+            # This also sets updated_at to migration time.
+            item.update_attributes(:created_at => item.container_quota.try(:created_on) || now)
+          end
+        end
+      end
+      # down: unnecessary, created_at/updated_at columns get deleted.
+    end
+    change_column_null :container_quota_items, :created_at, false
+    change_column_null :container_quota_items, :updated_at, false
+
     add_column :container_quota_items, :deleted_on, :datetime
     add_index :container_quota_items, :deleted_on
   end
--- a/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
+++ b/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
@@ -1,4 +1,7 @@
 class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
+  class ContainerQuota < ActiveRecord::Base; end
+  class ContainerQuotaItem < ActiveRecord::Base; end
+
   def change
     add_column :container_quotas, :deleted_on, :datetime
     add_index :container_quotas, :deleted_on
--- a/spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
+++ b/spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
@@ -33,7 +33,7 @@ describe AddDeletedOnToContainerQuotaAndItems do
       migrate
       item.reload
 
-      expect(item.deleted_on).to eq(nil)
+      expect(item.created_at).to eq(quota.created_on)
       expect(item.updated_at).not_to eq(nil)
     end
 
--- a/spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
+++ b/spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
@@ -1,0 +1,101 @@
+require_migration
+
+describe AddDeletedOnToContainerQuotaAndItems do
+  let(:container_quota_stub) { migration_stub(:ContainerQuota) }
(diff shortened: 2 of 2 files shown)


cben:
now with more tests and actually working fallback for missing parent (or missing parent created_on — neither should happen but don't want this to ever fail)

cben:
cc @moolitayer I broke the world, please review too if you can.

moolitayer:
@cben you can use this to remove the n+1: (unless you feel it's too evil :smile: )
```
join_sql = ContainerQuota.arel_table.project(:created_at).where(ContainerQuota.arel_table[:id].eq(ContainerQuotaItem.arel_table[:container_quota_id])).to_sql
ContainerQuotaItem.update_all("created_at = (#{join_sql})" )
# you might need to add a query for the `updated_at`
```

BTW I don't remember we use updated_at anywhere else. What was your goal with that
(in this PR of curse we won't change it)

cben:
N+1?  Oh, I see, `includes` helps for reading but not writing.
Thanks!  I wanted to figure out a way to do single SQL UPDATE FROM, but gave up...
However, that's still missing a fallback if parent is missing (which *shouldn't* be necessary but I don't want this migration to ever fail)...
I'll play with it if I have more time, which I probably won't.

`updated_at` is completely unnecessary in this table (with our plan to record history by never updating but archiving and creating new records), but @Fryguy asked to use standard `t.timestamps` / `add_timestamps` in #110 and it brought `updated_at` along for the ride...

cben:
confirmed: both timelines converge to [same schema.rb](https://gist.github.com/cben/3392d245b1edd26c015c3958a57689f3#file-clean-db-passed-original-schema-rb-L959-L989).

spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    it "works with a container_quota_item" do
+      quota = container_quota_stub.create!(quota_data_with_timestamps)
+      container_quota_item_stub.create!(:container_quota_id => quota.id, **item_data_with_timestamps)
+
+      migrate
+    end
+
+    it "works with an item whose quota.deleted_on is null" do
```
zakiva: quota.created_on

cben: thanks, fixed

Loicavenel:
@cben can we get this one merged today?

cben:
We must, currently everybody with a containers provider got broken.
Waiting for @Fryguy.
@zeari or @zakiva, you both started looking at this, can you approve?

spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      :quota_desired  => 0.02,
+      :quota_enforced => 0.02,
+      :quota_observed => 0.0,
+    }
+  end
+
+  migration_context :up do
+    it "succeeds with zero container_quota_items" do
+      migrate
+    end
```
Fryguy: This test is not necessary...every data migration has an automatic "empty tables" test.  https://github.com/ManageIQ/manageiq-schema/blob/master/spec/support/migration_helper.rb#L16

I'm not going to let this hold up the merge, as we can fix in a follow up PR.

spec/migrations/20171026114327_add_deleted_on_to_container_quota_and_items_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    let(:item_data_with_timestamps) do
+      item_data.merge(:created_at => "2017-10-21 11:11:11 UTC",
+                      :updated_at => "2017-10-22 22:22:22 UTC",
+                      :deleted_on => "2017-10-23 23:33:33 UTC")
+    end
+
+    it "suceeds with zero container_quota_items" do
+      migrate
+    end
```
Fryguy: Same.

db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
```diff
@@ -1,9 +1,34 @@
...
+    belongs_to :container_quota
+  end
+
   def change
     add_column :container_quotas, :deleted_on, :datetime
     add_index :container_quotas, :deleted_on
 
-    add_timestamps :container_quota_items
+    # Historical note: this migration has been edited. Original just naively added NOT
+    # NULL timestamp columns and failed with non-empty container_quota_items table.
```
Fryguy: comment is not necessary, as the tests make it clear why this code is here.  Luckily, anybody with container_quota data can't get past this migration anyway, so we can't have any devs in an in-between state.

db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
```diff
@@ -1,9 +1,34 @@
...
+
+    add_timestamps :container_quota_items, :null => true # temporarily for backfilling, then disallow nulls
+    reversible do |dir|
+      dir.up do
+        say_with_time("Backfilling container_quota_items timestamps from parent quotas") do
+          # All quota items SHOULD have a parent quota, which SHOULD have deleted_on
+          # but just in case, fallback to current time.
+          now = Time.zone.now
+          ContainerQuotaItem.includes(:container_quota).find_each do |item|
```
Fryguy: If you are using `.includes`, you have to be sure that the stub marks the belongs_to to use the right class name.

db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
```diff
@@ -1,9 +1,34 @@
 class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
+  class ContainerQuota < ActiveRecord::Base; end
+  class ContainerQuotaItem < ActiveRecord::Base
+    belongs_to :container_quota
```
Fryguy: Here, you have to specify the `:class_name => "AddDeletedOnToContainerQuotaAndItems::ContainerQuota"`

cben:
all done, PTAL

PR #122, opened 2017-11-06
Migrate Zone NTP settings to SettingsChanges

bdunne:
https://bugzilla.redhat.com/show_bug.cgi?id=1476365

PR #123, opened 2017-11-06
Downcase ems_ref for Azure resource groups

djberg96:
This is a migration that corresponds with https://github.com/ManageIQ/manageiq-providers-azure/pull/156.

In short, the ems_ref for an Azure resource group should always be lower case because we cannot guarantee the case of the resource group returned by the Azure REST API.

Marked as a WIP for now because my local env seems to be in a bad state.

### DIFF
--- a/db/migrate/20171117201600_downcase_resource_group_ems_ref_for_azure.rb
+++ b/db/migrate/20171117201600_downcase_resource_group_ems_ref_for_azure.rb
@@ -1,0 +1,9 @@
+class DowncaseResourceGroupEmsRefForAzure < ActiveRecord::Migration[5.0]
+  class ResourceGroup < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Downcase ems_ref for Azure resource groups") do
+      ResourceGroup.update_all("ems_ref = lower(ems_ref)")
+    end
+  end
+end
--- a/spec/migrations/20171117201600_downcase_resource_group_ems_ref_for_azure_spec.rb
+++ b/spec/migrations/20171117201600_downcase_resource_group_ems_ref_for_azure_spec.rb
@@ -1,0 +1,44 @@
+require_migration
+
+describe DowncaseResourceGroupEmsRefForAzure do
+  let(:resource_group_stub) { migration_stub :ResourceGroup }
+
+  let!(:uppercase_stub) do
+    resource_group_stub.create!(
+      :name    => 'UPPER',
+      :ems_ref => '/subscriptions/xyz/resourceGroups/UPPER'
+    )
+  end
+
+  let!(:mixedcase_stub) do
+    resource_group_stub.create!(
+      :name    => 'MiXeD',
+      :ems_ref => '/subscriptions/xyz/resourceGroups/MiXeD'
+    )
+  end
+
+  let!(:lowercase_stub) do
+    resource_group_stub.create!(
+      :name    => 'lower',
+      :ems_ref => '/subscriptions/xyz/resourceGroups/lower'
+    )
+  end
+
+  migration_context :up do
+    it 'Downcases the resource group ems_ref column' do
+      migrate
+
+      uppercase_stub.reload
+      expect(uppercase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/upper')
+      expect(uppercase_stub.name).to eql('UPPER')
+
+      mixedcase_stub.reload
+      expect(mixedcase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/mixed')
+      expect(mixedcase_stub.name).to eql('MiXeD')
+
+      lowercase_stub.reload
+      expect(lowercase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/lower')
+      expect(lowercase_stub.name).to eql('lower')
+    end
+  end
+end


spec/migrations/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  let(:lowercase_stub) do
+    resource_group_stub.create!(:name => 'lower', :ems_ref => '/subscriptions/xyz/resourceGroups/lower')
+  end
+
+  migration_context :up do
+    it 'Downcases the resource group ems_ref column' do
+      migrate
+      expect(uppercase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/upper')
```
Fryguy: I'm curious how this works since it creates the record *after* the migration occurs (because you are using `let` instead of `let!`

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
```diff
@@ -0,0 +1,11 @@
+class DowncaseResourceGroupEmsRefForAzure < ActiveRecord::Migration[5.0]
+  class ResourceGroup < ActiveRecord::Base; end
```
Fryguy: Is this an STI class?

spec/migrations/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
Fryguy: Oh I know how it "works"...it's because it's not running at all...the filename needs to be changed to have _spec.rb at the end

Fryguy:
Is this intended for gaprindashvili?  If so, we just have to verify that the migration number is correct.  @carbonin can you help out here...sorry...I've got to learn how to verify this one day :smile:

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
djberg96: There is a `type` column. Should I have subclassed that in the provider repo? If so I can add it in the other PR and update that column as well as part of the data migration as well.

spec/migrations/20171106143927_downcase_resource_group_ems_ref_for_azure_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    it 'Downcases the resource group ems_ref column' do
+      migrate
+      expect(uppercase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/upper')
+      expect(uppercase_stub.name).to eql('UPPER')
+
+      expect(mixedcase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/mixed')
+      expect(mixedcase_stub.name).to eql('MiXeD')
+
+      expect(lowercase_stub.ems_ref).to eql('/subscriptions/xyz/resourcegroups/lower')
+      expect(lowercase_stub.ems_ref).to eql('lower')
```
bronaghs: should this be `expect(lowercase_stub.name).to eql('lower')`

djberg96: Yep, too much copy-pasta.

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
Fryguy: If there is a type column, then it's STI.  You must

- create a record in your test with the type column set to a "real" model, to ensure it works
- when that test fails, you will need to set the inheritance column like we do in other model.

> Should I have subclassed that in the provider repo?

They should already be subclasses in the other repo, however migrations must work by not having those models present (which is why we use a dummy Rails app here).

carbonin:
As long as https://github.com/ManageIQ/manageiq-schema/pull/122 is also in gaprindashvili, this is also fine to be backported.

I would also try to merge and backport the PRs in migration order to keep the specs green (so #122 first)

spec/migrations/20171106143927_downcase_resource_group_ems_ref_for_azure_spec.rb
```diff
@@ -0,0 +1,31 @@
+require_migration
+
+describe DowncaseResourceGroupEmsRefForAzure do
+  let(:resource_group_stub) { migration_stub :ResourceGroup }
+
+  let!(:uppercase_stub) do
+    resource_group_stub.create!(:name => 'UPPER', :ems_ref => '/subscriptions/xyz/resourceGroups/UPPER')
```
Fryguy: Since the resource_group table uses STI, add a type here because a "real" record will always have a type (the subclass fully qualified name, eg. ManageIQ::Providers::Azure::ResourceGroup - I think that's the name, but get a real one to be sure).  It really doesn't even matter what the type is as long and it's not "ResourceGroup", and once you add that the tests will fail.  To get them green, you have to do what I said [up there](https://github.com/ManageIQ/manageiq-schema/pull/123/files#r149191728).

djberg96:
@Fryguy ok - https://github.com/ManageIQ/manageiq-providers-azure/pull/165

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def up
+    ResourceGroup.all.each do |resource_group|
+      if resource_group.ems_ref =~ /^\/subscriptions\/.*?\/resourcegroups\/\w+/i
+        resource_group.update_attributes(
+          :ems_ref => resource_group.ems_ref.downcase,
+          :type    => 'ManageIQ::Providers::Azure::ResourceGroup'
```
Fryguy: This needs to be set on *ALL* resource groups and the current logic prevents that...I recommend making this a separate data migration that runs prior to this one.

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
djberg96: The subclass was recently added to the Azure provider repo.

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
djberg96: I'm afraid that I don't understand. I wasn't totally sure if any other providers were using this. If they are, it would botch those records, wouldn't it?

Fryguy:
Pulling out the hidden discussion from https://github.com/ManageIQ/manageiq-schema/pull/123#discussion_r151531428 to here

> > This [the type] needs to be set on ALL resource groups and the current logic prevents that...I recommend making this a separate data migration that runs prior to this one.
>
> I'm afraid that I don't understand. I wasn't totally sure if any other providers were using this. If they are, it would botch those records, wouldn't it?

I had asked earlier if Azure was the only provider with ResourceGroups, and I was told it was, though maybe that's wrong?  @blomquisg 

Even so, it feels better to have one migration that just updates the type (on all records ideally or with whatever logic separates them), and a separate migration that downcases the ems_refs.

db/migrate/20171106143927_downcase_resource_group_ems_ref_for_azure.rb
blomquisg: I believe that `resource_groups` was added specifically for Azure.  So, no other provider should be using it.

blomquisg:
To @Fryguy's comment, he's right, all of the `resource_groups` need to have their type migrated to be Azure ResourceGroups.  Resource Groups were added specifically for Azure and no other provider is using them.

djberg96:
@Fryguy @blomquisg Ok, updated to just downcase ems_ref. I'll submit a separate PR for type.

Fryguy:
Once https://github.com/ManageIQ/manageiq-schema/pull/131 is merged in the date for this should change to come after it.

djberg96:
@Fryguy Update timestamp.

Fryguy:
@blomquisg Let me know when the corresponding PR is done and we can merge this.

simaishi:

```
$ git log -1
commit 303b0ed4abf2540a6bda147b87db8969e83297fe
Date:   Mon Nov 20 10:18:01 2017 -0500

    Merge pull request #123 from djberg96/azure_resource_group_emsref
    
    Downcase ems_ref for Azure resource groups
    (cherry picked from commit 8e5a9a549925c5614f535d2cdb811c2d64dd62c4)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1515452
```

PR #132, opened 2017-11-17
Change migration to use batch output instead of print "."

### LINKED ISSUE #2: Fix certain migration that puts "."
Instead use the batch commenting or `say`

Fryguy:
@chessbyte @gtanzillo Please review.

This also changes the batch size from 1000 to 25000 because 1000 is WAY too chatty, particular for this table where millions of rows is not unexpected.   I was going to make it 100000, but on my machine it took about 5 seconds to do 100000 rows, so I thought worst case that could be 10 seconds on a customer's machine which might be too long.  25000 was a little over 1 second on my machine.

cc @lfu 

Closes #2

### DIFF
--- a/db/migrate/20150806211453_rename_ems_event_table_to_event_stream.rb
+++ b/db/migrate/20150806211453_rename_ems_event_table_to_event_stream.rb
@@ -2,6 +2,8 @@ class RenameEmsEventTableToEventStream < ActiveRecord::Migration[4.2]
   disable_ddl_transaction!
   include MigrationHelper
 
+  BATCH_SIZE = 1000
+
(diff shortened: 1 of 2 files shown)


simaishi:

```
$ git log -1
commit 87bd79d3619a544a5f8cf5dd9818a04e9e88154c
Author: Gregg Tanzillo <gtanzill@redhat.com>
Date:   Mon Nov 20 14:29:13 2017 -0500

    Merge pull request #132 from Fryguy/remove_print_statement
    
    Change migration to use batch output instead of print "."
    (cherry picked from commit cb45bf4f148a8cae66b221f23e9ad243275f3a13)
```

PR #136, opened 2017-11-21
Fix migration column caching.

Fryguy:
When this migration is run in the same process as
20170827091406_change_container_quota_items_columns_to_float.rb, that
migration does some column caching for its ContainerQuotaItem stub model
which ends up being inherited by this migration.  The symptom is that
this migration's stub model does not actually have the created_at
column, and thus never writes the value.  Later on when
change_column_null is called, the column has null values in it causing
that statement to fail.

https://bugzilla.redhat.com/show_bug.cgi?id=1515376

@gtanzillo Please review
cc @cben 

cc @chrisarcand I think this is a Rails bug, but I'm not sure how to prove it, *and* we're not on Rails 5.1, so we'd have to reproduce on Rails 5.1 first, anyway.

### DIFF
--- a/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
+++ b/db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
@@ -9,6 +9,8 @@ class AddDeletedOnToContainerQuotaAndItems < ActiveRecord::Migration[5.0]
     add_index :container_quotas, :deleted_on
 
     add_timestamps :container_quota_items, :null => true # temporarily for backfilling, then disallow nulls
+    ContainerQuotaItem.reset_column_information
+
     reversible do |dir|
       dir.up do
         say_with_time("Backfilling container_quota_items timestamps from parent quotas") do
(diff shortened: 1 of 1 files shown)


db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb
```diff
@@ -9,6 +9,8 @@ def change
     add_index :container_quotas, :deleted_on
 
     add_timestamps :container_quota_items, :null => true # temporarily for backfilling, then disallow nulls
+    ContainerQuotaItem.reset_column_information
```
cben: I'm surprised you put this in the middle, from your description of the problem I'd expect clearing things "between" the migrations.
BTW, down direction runs `change` statements in reverse — did you test down also works?  (it's likely caching is not a problem if this migration runs first anyway.)

Hmm, I guess the way `change` works, to implement running in reverse, is first only recording steps to be done.  Since this line is not in a `reversible` block, it probably runs before all the steps, for both directions, whereever you place it?

BTW, why didn't the migrate-all-the-way-up test catch this?

[I don't care enough to play with this myself, and not saying we should touch it again, if it works]

PR #140, opened 2017-12-01
Add status to ext_management_system

agrare:
Adds a status column to ext_management_system to track when the EMS has
been marked for deletion.  This will allow it to be displayed to the
user as being in the process of being deleted and we can prevent the
user from re-enabling an EMS which is in the process of deletion.

Related: https://github.com/ManageIQ/manageiq/pull/16574


agrare:
@miq-bot assign @Fryguy

agrare:
Couldn't decide on state vs status, opinions?  Should I initialize existing records to `"ok"` or leave nil?

agrare:
@Fryguy still think this is a good idea?

agrare:
ping @Fryguy

agrare:
Going to close this for now since it isn't getting any traction.  If we find we need this in the future I will re-open it.

PR #142, opened 2017-12-04
[WIP] Renames dialog label to name

d-m-u:
We're using `label` to refer to the dialog name and we call it `name` in the UI (and we're aliasing the darn thing here: https://github.com/ManageIQ/manageiq/blob/master/app/models/dialog.rb#L17) so it kinda made sense to just name the thing `name`. 

## Related to:
~~https://github.com/ManageIQ/manageiq/pull/16507~~ (merged)
https://github.com/ManageIQ/manageiq/pull/16600


d-m-u:
@miq-bot assign @Fryguy

gmcculloug:
@d-m-u How does this effect older dialog exports from previous schema versions to this one?  I would expect that those exports would still be importable with this change.

PR #147, opened 2018-01-02
Drop `vim_performance_tag_values` table from DB

gtanzillo:
This table is no longer used as of https://github.com/ManageIQ/manageiq/pull/16728

Should be merged after #16728


Fryguy:
@gtanzillo Is this good to go?  Are all of the corresponding callers removed?

gtanzillo:
@Fryguy 
> Is this good to go? Are all of the corresponding callers removed?

https://github.com/ManageIQ/manageiq/pull/16728 removes all the callers. That needs to be merged first.

Fryguy:
@gtanzillo Since this is so old can you update this PR and bump the timestamp to something more recent?

gtanzillo:
>@gtanzillo Since this is so old can you update this PR and bump the timestamp to something more recent?

@Fryguy Done 👍

PR #151, opened 2018-01-10
Change container_quota_items float columns to decimals

cben:
@Fryguy @agrare @Ladas @moolitayer @zeari @bazulay
I'm told this might still be possible **for gaprindashvili**?!

Fixes https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/208
Kubernetes quotas are mostly integers but cpu-related quotas are
in "millicores", multiples of 0.001.
With decimal columns, we'll be able to round-trip quotas exactly to DB
and back to Ruby, so refresh can determine whether they changed or not.
(requires couple more code changes, but this fixes the deep root problem)

Let's not merge until I have all pieces together, but this does allow the exact fraction to come back to Ruby.

TODO: tests demonstrating what happens to existing values.

TODO: benchmark on lots of data.

### Alternative: fixed-point integer column (5 would mean 5milli) ?

Some of you said that'd simpler and faster.
We wouldn't want all quotas scaled (memory in millibytes, number of pods in millipods...) but only scale CPU in millicores.

- However it'd need data migration.  Hmm, I suppose if we can do this in gaprindashvili, then data migration is possible too?
- I'm concerned at this point we'd be likely to introduce unit mismatch bugs.

### DIFF
--- a/db/migrate/20180118131417_change_container_quota_items_to_decimals.rb
+++ b/db/migrate/20180118131417_change_container_quota_items_to_decimals.rb
@@ -1,0 +1,13 @@
+class ChangeContainerQuotaItemsToDecimals < ActiveRecord::Migration[5.0]
+  def up
+    %i(quota_desired quota_enforced quota_observed).each do |column|
+      change_column :container_quota_items, column, :decimal, :scale => 3, :precision => 30
+    end
+  end
+
+  def down
+    %i(quota_desired quota_enforced quota_observed).each do |column|
+      change_column :container_quota_items, column, :float
+    end
+  end
+end


oourfali:
@chessbyte you approved this schema change, right?

Fryguy:
I believe he approved to go into gaprindashvili but not the PR itself (wasn't sure what you meant by "approved")

Fryguy:
Code-wise LGTM :+1:

Fryguy:
> However it'd need data migration. Hmm, I suppose if we can do this in gaprindashvili, then data migration is possible too?

Yes, if this is backported to gap we would also need to bring back the corresponding data migration.  Once we cut gaprindashvili, then we can't backport migrations anymore.

cben:
Oops, I omitted in PR description what "would be simpler & faster" and "would need data migration". 
 Edited now.  This was referring to alternative of switching from float to integer columns, requiring cpu quotas to be scaled by fixed factor of 1000.

I don't like that, not at this point, concerned we'd miss off-by-1000x bugs in other places...  Decimal is overkill but conceptually clean Right way to do it.  "Last-moment feature, correct, fast — pick at most 2"...

Fryguy:
Ah ok, so then this is ready to go?

cben:
Let me test this without the other PRs.  I *think* existing refresh that doesn't archive yet would work with decimal just as well but need to confirm.

cben:
I believe this is safe to merge without any code changes.

- [x] Passes current k8s and openshift tests.
- [x] Passes tests from https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/216, https://github.com/ManageIQ/manageiq-providers-openshift/pull/83 for what happens with modified quota (which expect in-place update, no archiving).
- [x] Refresh works in `rails c`, updates in-place.
- [ ] Core ManageIQ tests: dunno.  Mostly passed, got 13 unrelated sounding errors mostly about "pictures", tried resetting all plugins to master, evm:db:reset, no luck and now can't run core specs at all :(  Let's try on Travis: https://travis-ci.org/cben/manageiq/jobs/330436694#L1427

cben:
- [ ] ooh, also need to test UI.

cben:
@Fryguy all tested.

simaishi:

```
$ git log -1
commit 90a252db428966120151c8be57a3745bda44ff76
Date:   Thu Jan 18 15:31:15 2018 -0500

    Merge pull request #151 from cben/quota-decimal
    
    Change container_quota_items float columns to decimals
    (cherry picked from commit 909736e0d08de464d352980306a451af5e4d4228)
```

PR #153, opened 2018-01-11
Migrate Picture content from BinaryBlobs to Pictures table

bdunne:
To be merged with: https://github.com/ManageIQ/manageiq/pull/16810

https://bugzilla.redhat.com/show_bug.cgi?id=1532286

### DIFF
--- a/db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
+++ b/db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
@@ -1,0 +1,52 @@
+class MovePicturesBlobsToPictures < ActiveRecord::Migration[5.0]
+  class BinaryBlob < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    has_many :binary_blob_parts, -> { order(:id) }, :dependent => :delete_all, :class_name => "MovePicturesBlobsToPictures::BinaryBlobPart"
+    def data
+      binary_blob_parts.pluck(:data).join
+    end
+  end
+  class BinaryBlobPart < ActiveRecord::Base; end
+  class Picture < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    add_column :pictures, :content, :binary
+    add_column :pictures, :extension, :string
+    add_column :pictures, :md5, :string
+
+    say_with_time("Moving picture content from BinaryBlobs to the pictures table") do
+      BinaryBlob.in_my_region.includes(:binary_blob_parts).where(:resource_type => "Picture").find_each do |blob|
+        Picture.update(blob.resource_id, :content => blob.data, :extension => blob.data_type, :md5 => blob.md5)
+        blob.destroy
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving picture content from the pictures table to BinaryBlobs") do
+      Picture.in_my_region.find_each do |picture|
+        size = picture.content.try(:length).to_i
+        next if size.zero?
+
+        BinaryBlob.create!(
+          :data_type     => picture.extension,
+          :md5           => picture.md5,
+          :part_size     => size,
+          :resource_id   => picture.id,
+          :resource_type => "Picture",
+          :size          => size,
+        ).binary_blob_parts.create!(
+          :data => picture.content,
+          :md5  => picture.md5,
+          :size => size,
+        )
+      end
+    end
+
+    remove_column :pictures, :content
+    remove_column :pictures, :extension
+    remove_column :pictures, :md5
+  end
+end
--- a/spec/migrations/20180111162442_move_pictures_blobs_to_pictures_spec.rb
+++ b/spec/migrations/20180111162442_move_pictures_blobs_to_pictures_spec.rb
@@ -1,0 +1,106 @@
+require_migration
+
+describe MovePicturesBlobsToPictures do
+  let(:binary_blob_stub)      { migration_stub :BinaryBlob }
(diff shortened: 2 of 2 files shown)


bdunne:
@miq-bot add_labels core enhancement gaprindashvili/yes

db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
```diff
@@ -0,0 +1,51 @@
+class MovePicturesBlobsToPictures < ActiveRecord::Migration[5.0]
+  class BinaryBlob < ActiveRecord::Base
+    has_many :binary_blob_parts, -> { order(:id) }, :dependent => :delete_all
```
Fryguy: needs the :class designation to ensure it uses the stub.

db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
```diff
@@ -0,0 +1,51 @@
...
+    belongs_to :resource, :polymorphic => true
```
Fryguy: You don't need this.

db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
```diff
@@ -0,0 +1,51 @@
...
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    add_column :pictures, :content, :binary
+    add_column :pictures, :extension, :string
+    add_column :pictures, :md5, :string
+
+    say_with_time("Moving picture content from BinaryBlobs to the pictures table") do
+      BinaryBlob.where(:resource_type => "Picture").all.each do |blob|
```
Fryguy: Let's change this to a find_each just to ensure we batch it properly just in case there are lot of these.

db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving picture content from the pictures table to BinaryBlobs") do
+      Picture.in_my_region.all.each do |picture|
+        size = picture.content.try(:length).to_i
+
+        BinaryBlob.create!(
```
Fryguy: We shouldn't create the binaryblob if the size is 0.  Though that seems like a real edge case.

db/migrate/20180111162442_move_pictures_blobs_to_pictures.rb
Fryguy: Also add a `.includes(:binary_blob_parts)` to avoid an N+1?

Fryguy: Also also do you need `in_my_region`?  Even though they are not replicated, someone *might* have enabled that.

himdel:
```
== 20180111162442 MovePicturesBlobsToPictures: migrating ======================          
-- add_column(:pictures, :content, :binary)                                              
   -> 0.0045s                                                                            
-- add_column(:pictures, :extension, :string)
   -> 0.0005s
-- add_column(:pictures, :md5, :string)
   -> 0.0010s
-- Moving picture content from BinaryBlobs to the pictures table
rails aborted!

Couldn't find MovePicturesBlobsToPictures::Picture without an ID
bin/rails:4:in `require'
bin/rails:4:in `<main>'
ActiveRecord::RecordNotFound: Couldn't find MovePicturesBlobsToPictures::Picture without an ID
bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:migrate                                                                 
(See full trace by running task with --trace)                                            
                                                                                         
== Command ["bin/rails db:migrate"] failed in /home/himdel/manageiq ==
```

himdel:
happens during `up`, blob looks like this:

```
#<MovePicturesBlobsToPictures::BinaryBlob id: 10000000014757, resource_type: "Picture", resource_id: nil, md5: "7d4b8a5fbbdfdd85da991e00e22e1545", size: 14853, part_size: 1048576, name: nil, data_type: "jpeg">
```

Fryguy:
How weird that you have nil pictures...that should not be possible, unless the original table was `dependent => nullify`?

himdel:
It *could* theoretically have to do with updating GenericObjectDefintion via the API to remove the picture.

Looks like it does something along the lines of `god.update_attributes!(:picture => nil)`.
But I'm not really sure if it's a problem still present in the current implementation or if this only happened during review.

simaishi:
I'm holding off backporting to Gaprindashvili. Let me know if/when it's ok to backport.

Fryguy:
@jntullo ?

Fryguy:
@bdunne Can you update the PR to "deal with" resource_id being nil?  However, separately, we should not be putting the DB in this state, so let's also fix whatever bug is causing this to happen.

jntullo:
@Fryguy will look into that - also look into other seemingly related spec failures in the API.

bdunne:
@simaishi I think this is all set for backport with https://github.com/ManageIQ/manageiq-schema/pull/158

simaishi:

```
$ git log -1
commit 6faac0b77f8d0ca90015ad6be847fcd52f4987e0
Date:   Wed Jan 17 16:21:22 2018 -0500

    Merge pull request #153 from bdunne/move_pictures
    
    Migrate Picture content from BinaryBlobs to Pictures table
    (cherry picked from commit 91b0fc0430a7ebe601960be693ffeddf2ce842b2)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1536046
```

PR #154, opened 2018-01-12
Drop BinaryBlobPart#md5 and #size

bdunne:
The database should ensure that the object returned is the same as that stored, making this column unnecessary.

Based on https://github.com/ManageIQ/manageiq/pull/16812


db/migrate/20180112171348_drop_unnecessary_blob_parts_md5.rb
```diff
@@ -0,0 +1,20 @@
...
+    remove_column :binary_blob_parts, :size
+  end
+
+  def down
+    add_column :binary_blob_parts, :md5,  :string
+    add_column :binary_blob_parts, :size, :decimal, :precision => 20, :scale => 0
+
+    say_with_time("Calculating md5 and size of all BinaryBlobParts") do
+      require 'digest'
+      BinaryBlobPart.all.each do |part|
```
Fryguy: Might want to do this with .in_my_region and also with .find_each.

PR #156, opened 2018-01-16
create a physical rack table

felipedf:
This create a new table physical_racks and provides a :physical_rack_id column to physical_servers table.

### DIFF
--- a/db/migrate/20180116210505_create_physical_racks_and_add_to_physical_servers.rb
+++ b/db/migrate/20180116210505_create_physical_racks_and_add_to_physical_servers.rb
@@ -1,0 +1,12 @@
+class CreatePhysicalRacksAndAddToPhysicalServers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_racks do |t|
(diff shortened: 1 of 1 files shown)


felipedf:
@Fryguy

agrare:
@miq-bot assign @Fryguy

db/migrate/20180116210559_add_physical_rack_to_physical_servers.rb
```diff
@@ -0,0 +1,5 @@
+class AddPhysicalRackToPhysicalServers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_servers, :physical_rack_id, :bigint
```
agrare: Don't think you need/want to do this in its own migration, adding a column to the new table in the above migration is fine /cc @Fryguy opinions?

felipedf:
@agrare resolved

Fryguy:
@felipedf I just realized this didn't include an ems_ref column, which is generally required if the table is to be filled in during inventory refresh.

PR #160, opened 2018-01-26
Updating switches data

saulotoledo:
This PR updates the Switches table and adds the switch id to the hardwares table in order to allow us to parse physical infra switches.

### DIFF
--- a/db/migrate/20180126144529_add_extra_information_to_switches.rb
+++ b/db/migrate/20180126144529_add_extra_information_to_switches.rb
@@ -1,0 +1,13 @@
+class AddExtraInformationToSwitches < ActiveRecord::Migration[5.0]
+  def change
+    add_column :switches, :ems_id, :bigint
+    add_column :switches, :type, :string
+    add_column :switches, :health_state, :string
+    add_column :switches, :power_state, :string
+    add_column :switches, :product_name, :string
+    add_column :switches, :part_number, :string
+    add_column :switches, :serial_number, :string
+    add_column :switches, :description, :string
+    add_column :switches, :manufacturer, :string
+  end
+end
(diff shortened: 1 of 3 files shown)


saulotoledo:
@Fryguy, could you take a look at this PR?

agrare:
@saulotoledo I think while we're at it we should add a `:type` to `switches` so that we can STI the Switch model

saulotoledo:
Hi @agrare. I was playing for some time with that before doing the change. I added the *:type* but the ActiveRecord is not proper saving the type. Sounds like the refresher is using a generic class to save the hash, so that will not work. Doing some research I see that actually the type for physical servers [is filled manually by the parser](https://github.com/saulotoledo/manageiq-providers-lenovo/blob/4bd578971c79cfa77d0abbba70db4b3e8110cda6/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb#L33) in order to do a workaround here. I can do the same, but I cannot assure that it will be filled for other providers. How do you think I should proceed?

agrare:
@saulotoledo try adding `NewWithTypeStiMixin` to the switch model like e.g. https://github.com/ManageIQ/manageiq/blob/master/app/models/host.rb#L11

saulotoledo:
@agrare, it is done. I also added a new migration to the guest_devices table. It is an information that we need to be able to discover the connected devices. Let me know what you think about.

saulotoledo:
I will add a property to store the switch's MAC addresses. I will remove the WIP prefix after the update.

saulotoledo:
@agrare, @rodneyhbrown7, the mac_addresses column was added. The parser has also been updated.

PR #161, opened 2018-01-29
Remove all instances of ManageIQ::Providers::Hawkular::MiddlewareManager from ext_management_systems

himdel:
Since https://github.com/ManageIQ/manageiq/pull/16834, there's no longer an entity called `ManageIQ::Providers::Hawkular::MiddlewareManager`.

This should remove those...

(Not sure if we should be trying to remove references, etc.?)

### DIFF
--- a/db/migrate/20180129152935_remove_hawkular_ems.rb
+++ b/db/migrate/20180129152935_remove_hawkular_ems.rb
@@ -1,0 +1,11 @@
+class RemoveHawkularEms < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Removing Hawkular MiddlewareManagers") do
+      ExtManagementSystem.where(:type => 'ManageIQ::Providers::Hawkular::MiddlewareManager').delete_all
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


himdel:
@miq-bot add_label gaprindashvili/no

db/migrate/20180129152935_remove_hawkular_ems.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveHawkularEms < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
```
lpichler:
```
  class ExtManagementSystem < ActiveRecord::Base
    self.inheritance_column = :_type_disabled
  end
```

https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-310643185

db/migrate/20180129152935_remove_hawkular_ems.rb
```diff
@@ -0,0 +1,8 @@
...
+  end
+
+  def up
+    ExtManagementSystem.where(:type => 'ManageIQ::Providers::Hawkular::MiddlewareManager').delete_all
```
lpichler: Can you wrap it by `say_with_time` ?
```ruby
    say_with_time("Removing ....") do
      ExtManagementSystem.where(:type...
    end
```

Fryguy:
Looks good, but needs a spec because it's a data migration.  A spec showing the record deleted and another spec showing that *other* records are unaffected should be plenty.

PR #165, opened 2018-02-02
Add lan info to a guest_device (port)

felipedf:
Adding vlan info to a guest_device when it assume a logical port function

### DIFF
--- a/db/migrate/20180202194924_add_vlan_info_to_guest_devices.rb
+++ b/db/migrate/20180202194924_add_vlan_info_to_guest_devices.rb
@@ -1,0 +1,6 @@
+class AddVlanInfoToGuestDevices < ActiveRecord::Migration[5.0]
+  def change
+    add_column :guest_devices, :vlan_key, :string
+    add_column :guest_devices, :vlan_enabled, :boolean
+  end
+end


agrare:
@miq-bot assign @Fryguy

agrare:
@felipedf could you add a description of what these fields mean?  I don't see how they relate to VLANs or how `vnicMode` is a boolean and not a string.

felipedf:
@agrare well, now guest_devices can assume a role of a port and a port can have many logical ports, which will also be a guest_device through child_devices field. Supposedly a logical port contains a reference to its vlan and the vnicMode would say whether or not vnic is enabled on this port

agrare:
@felipedf is this a port on a switch?  I wonder if we should add a port model and a switch can have many ports, then a guest_device belongs_to a port?

agrare:
Spoke offline with @felipedf and this is intended for host NICs not switch ports, and these properties are meant to represent the vlan_id.  I recommended just calling it vlan_id and using just one property.

felipedf:
Spoke with @agrare and decided to keep vlan_id and a vlan_enabled properties, so where vlan_enabled is meant to represent the status of a vlan, so a port can be associated to a vlan even though its status is disabled

rodneyhbrown7:
@agrare @Fryguy  can this be merged?

agrare:
Can you change the name from `:vlan_id` to something that doesn't end in `_id` since rails will assume this is a foreign key.
This will represent a physical port on a switch with a tagged vlan.

felipedf:
Done :smiley: :+1:

PR #166, opened 2018-02-06
Remove all VMware MKS console-related records from SettingSchanges

skateman:
This feature has been dropped in [fine](https://github.com/ManageIQ/manageiq-ui-classic/pull/1052) and [later](https://github.com/ManageIQ/manageiq-ui-classic/pull/979), however, the migration was missing. There's also a PR for EUWE: https://github.com/ManageIQ/manageiq/pull/16909

https://bugzilla.redhat.com/show_bug.cgi?id=1513592

@miq-bot assign @bdunne 
@miq-bot add_label fine/yes, gaprindashvili/yes

### DIFF
--- a/db/migrate/20180130165200_drop_webmks.rb
+++ b/db/migrate/20180130165200_drop_webmks.rb
@@ -1,0 +1,10 @@
+class DropWebmks < ActiveRecord::Migration[5.0]
+  class MiqServer < ActiveRecord::Base; end
+  class SettingsChange < ActiveRecord::Base; end
+
+  def up
+    say_with_time "Remove all VMware MKS console-related records from settings" do
+      SettingsChange.where(:key => '/server/remote_console_type', :value => 'MKS').delete_all
+    end
+  end
+end
--- a/spec/migrations/20180130165200_drop_webmks_spec.rb
+++ b/spec/migrations/20180130165200_drop_webmks_spec.rb
@@ -1,0 +1,17 @@
+require_migration
+
+describe DropWebmks do
+  let(:server_stub)   { migration_stub(:MiqServer) }
+  let(:settings_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    let(:server) { server_stub.create }
+
+    it 'resets the remote_console_type to default' do
+      settings_stub.create(:resource_id => server.id, :resource_type => server.class, :key => '/server/remote_console_type', :value => 'MKS')
+      migrate
+
+      expect(settings_stub.where(:key => '/server/remote_console_type', :value => 'MKS')).not_to exist
+    end
+  end
+end


spec/migrations/20180130165200_drop_webmks_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe DropWebmks do
+  let(:server_stub)   { migration_stub(:MiqServer) }
+  let(:settings_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    let(:server) { server_stub.create }
+
+    before do
+      settings_stub.create(:resource => server, :key => '/server/remote_console_type', :value => 'MKS')
```
bdunne: This can be inside the `it` since there's only 1 test.

spec/migrations/20180130165200_drop_webmks_spec.rb
```diff
@@ -0,0 +1,20 @@
+require_migration
+
+describe DropWebmks do
+  let(:server_stub)   { migration_stub(:MiqServer) }
```
bdunne: This is failing rspec since `class MiqServer < ActiveRecord::Base; end ` isn't defined in the migration namespace

db/migrate/20180130165200_drop_webmks.rb
```diff
@@ -0,0 +1,7 @@
+class DropWebmks < ActiveRecord::Migration[5.0]
+  class SettingsChange < ApplicationRecord; end
```
bdunne: I think we want `ActiveRecord::Base` to be the parent class in this case since migrations shouldn't be using any models.  cc @Fryguy

db/migrate/20180130165200_drop_webmks.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def up
+    SettingsChange.where(:resource_type => 'MiqServer', :key => '/server/remote_console_type', :value => 'MKS').delete_all
```
bdunne: We don't need the `:resource_type`.  Removing that from the scope would probably be a good idea in case someone added it to a zone settings.

spec/migrations/20180130165200_drop_webmks_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:settings_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    let(:server) { server_stub.create }
+
+    it 'resets the remote_console_type to default' do
+      settings_stub.create(:resource => server, :key => '/server/remote_console_type', :value => 'MKS')
```
bdunne: Now the tests are failing because you either need to specify `:resource_type` and `:resource_id` or define the relation  in the class defined above.

db/migrate/20180130165200_drop_webmks.rb
```diff
@@ -0,0 +1,8 @@
+class DropWebmks < ActiveRecord::Migration[5.0]
+  class MiqServer < ActiveRecord::Base; end
+  class SettingsChange < ActiveRecord::Base; end
+
+  def up
+    SettingsChange.where(:key => '/server/remote_console_type', :value => 'MKS').delete_all
```
lpichler: you can wrap it with `say_with_time`

Fryguy:
Migrations cannot be backported.  If these records are in the database is there a problem with that?

spec/migrations/20180130165200_drop_webmks_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:settings_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    let(:server) { server_stub.create }
+
+    it 'resets the remote_console_type to default' do
+      settings_stub.create(:resource_id => server.id, :resource_type => server.class, :key => '/server/remote_console_type', :value => 'MKS')
+      migrate
+
+      expect(settings_stub.where(:key => '/server/remote_console_type', :value => 'MKS').count).to be_zero
```
Fryguy: `expect(stuff).to_not exist`

Fryguy:
Code-wise looks good to me, but I want to understand the backport strategy before merging.

skateman:
@Fryguy well we had a [discussion](https://github.com/ManageIQ/manageiq/pull/16909) with @bdunne at the original euwe-only PR, check it out.

skateman:
@Fryguy I moved out the schema code from the #16909. After a little bit of thinking I don't think this is really necessary but it would not do any harm, so your call to merge or close.

PR #167, opened 2018-02-13
Add miq_task_id column to miq_queue table

yrudman:
Linking `miq_queue` and `miq_tasks` table will:
- allow to add miq_task_id to the list of arguments when message deliver
- simplify activating task (https://bugzilla.redhat.com/show_bug.cgi?id=1543289)

blocking https://github.com/ManageIQ/manageiq/pull/17015

\cc @Fryguy

### DIFF
--- a/db/migrate/20180213204846_link_queue_and_task.rb
+++ b/db/migrate/20180213204846_link_queue_and_task.rb
@@ -1,0 +1,5 @@
+class LinkQueueAndTask < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :miq_queue, :miq_task, :type => "bigint", :index => true
+  end
+end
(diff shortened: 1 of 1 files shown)


yrudman:
@miq-bot remove-label wip

yrudman:
@miq-bot add-label core,  sql migration

yrudman:
@miq-bot assign @Fryguy

db/migrate/20180213204846_link_queue_and_task.rb
```diff
@@ -0,0 +1,5 @@
+class LinkQueueAndTask < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :miq_queue, :miq_task, :type => "bigint", :index => true
```
Fryguy: You shouldn't need the explicit `:type => "bigint"`, but I could be wrong.  Even so, we should probably use a sybol there instead of a string.  Also, not sure on adding an index.  @kbrock Thoughts on this index?

PR #168, opened 2018-02-15
Add hostname to Vm

sasoc:
This migration adds attribute `hostname` to `Vm`. For example
virtual machine in Vmware provider has attribute computer name
and now we will be able to render it into hostname.

### DIFF
--- a/db/migrate/20180215164010_add_hostname_to_vm.rb
+++ b/db/migrate/20180215164010_add_hostname_to_vm.rb
@@ -1,0 +1,5 @@
+class AddHostnameToVm < ActiveRecord::Migration[5.0]
+  def change
+    add_column :vms, :hostname, :string
+  end
+end


sasoc:
cc @miha-plesko

miha-plesko:
@miq-bot assign @agrare

/cc @Ladas

Fryguy:
I'm not sure I understand why this is being added, since hostnames are already available on Vm.  Was this discussed somewhere?  I thought on Gitter we already pointed out that where you can see the hostname in the UI.

```ruby
Vm.first.hostnames # => ["windows_2012"]
```

agrare:
@Fryguy I think the problem is two-fold

1. Cloud VMs don't have `hardware.networks` so they don't have that option
2. A machine can only have one "hostname", the rest are DNS records that aren't really modeled well currently e.g. there can only be one hostname per ip address.

Currently InfraManagers are just duplicating the one hostname into multiple network records so it doesn't really match either case very well.

> Was this discussed somewhere

Yes it was discussed on gitter I think right after where you pointed out where it is shown in the UI

miha-plesko:
@Fryguy we would also need to backport this `vm.hostname` field to GA by means of reserved table, so I kindly ask to push this PR forward.

agrare:
@miq-bot assign @Fryguy

PR #175, opened 2018-02-28
Move SchemaMigration model from ManageIQ to ManageIQ::Schema plugin

bdunne:
Fixes https://github.com/ManageIQ/manageiq/issues/16670


bdunne:
@Fryguy Please review

PR #176, opened 2018-03-13
Add vm_ems_ref to event_streams

agrare:
Early events in a VM's lifecycle don't contain the vmPathName yet which
is used to connect events which were processed before the VM gets
created by EmsRefresh.  These events do however contain the ems_ref of
the VM which can be used for more reliable event -> vm association.

https://bugzilla.redhat.com/show_bug.cgi?id=1428797


agrare:
@miq-bot assign @Fryguy 

PR #177, opened 2018-03-13
RedHat is a system domain.

lfu:
https://bugzilla.redhat.com/show_bug.cgi?id=1478574

@miq-bot add_label bug, fine/yes, gaprindashvili/yes

cc @mkanoor @gmcculloug @tinaafitz


gmcculloug:
It is common practice to update the Automate domains stored on disk as part of an upgrade so this PR should not be needed.  This domain's system attribute has already been updated in a separate change.

gmcculloug:
In addition, anyone resetting the domains after this migration is applied would not see any effect, so if this type of logic is needed it would seem like seeding logic.  But again I do not think we need this.

@lfu @Fryguy Could one of you please close this PR.

PR #181, opened 2018-03-19
Adding product details columns to asset_details table

douglasgabriel:
__This PR is able to:__
- Add product details columns to `asset_details` table.

__Goal:__
This PR aims to centralize some common informations of products (i.e. physical servers, switches etc) on the `asset_details` table.

PR #182, opened 2018-03-23
Updating details for physical switches

saulotoledo:
Removes details from switches table, since [due to later changes](https://github.com/ManageIQ/manageiq-schema/pull/181/files) we can now use `asset_details` table to store the same information. We also add the `part_number` column to the asset details and rename the relationship in the hardwares table, since we will use physical_switches as the model in the [core PR](https://github.com/ManageIQ/manageiq/pull/16948).

### DIFF
--- a/db/migrate/20180323204738_add_part_number_to_asset_details.rb
+++ b/db/migrate/20180323204738_add_part_number_to_asset_details.rb
@@ -1,0 +1,5 @@
+class AddPartNumberToAssetDetails < ActiveRecord::Migration[5.0]
+  def change
+    add_column :asset_details, :part_number, :string
+  end
+end
--- a/db/migrate/20180323204821_remove_switch_details_from_switches.rb
+++ b/db/migrate/20180323204821_remove_switch_details_from_switches.rb
@@ -1,0 +1,9 @@
+class RemoveSwitchDetailsFromSwitches < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :switches, :product_name,  :string
+    remove_column :switches, :part_number,   :string
+    remove_column :switches, :serial_number, :string
+    remove_column :switches, :description,   :string
+    remove_column :switches, :manufacturer,  :string
+  end
+end


saulotoledo:
@bdunne, @agrare, could you take a look in the lint?

I think I should not have to fix the `Style/MethodCallWithArgsParentheses` rule above. There is no commit in the project by using parentheses. [Here](https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20160414192525_rename_configuration_manager_to_manager.rb) is an example.

agrare:
@saulotoledo don't worry about that something must be wrong with the rubocop exceptions in this repo, /cc @bdunne

bdunne:
That's correct, I updated the ignored methods list in https://github.com/ManageIQ/guides/pull/299

saulotoledo:
@rodneyhbrown7, @agrare, @bdunne, I think this is ok for review.

saulotoledo:
@miq-bot remove_label wip

db/migrate/20180326193542_rename_switch_id_column_in_hardware.rb
```diff
@@ -0,0 +1,5 @@
+class RenameSwitchIdColumnInHardware < ActiveRecord::Migration[5.0]
+  def change
+    rename_column :hardwares, :switch_id, :physical_switch_id
```
agrare: You should be fine with switch_id here since the table is still called switches the fk should be switch_id.  Physical_switches is just the name of the association

saulotoledo: Done

gtanzillo:
@Fryguy You ok with merging this one?

Fryguy:
@saulotoledo Is there a corresponding PR where the callers of these columns are changed such that they are no longer used?  I'm concerned that if I merge, then master will be broken.

agrare:
@Fryguy https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132 updates the parser to work with the new columns

Fryguy:
@agrare So that means that ManageIQ/manageiq-providers-lenovo#132 needs to merged in tandem with this one?

saulotoledo:
@Fryguy @agrare It is safe to merge this one. The only code who uses these columns are in [https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132), so this one needs to be merged first. Thse colums were created by me on [https://github.com/ManageIQ/manageiq-schema/pull/160/files](https://github.com/ManageIQ/manageiq-schema/pull/160/files) and are only used in my PRs (provider and UI on WIP state yet).

gtanzillo:
@saulotoledo We're (me and @Fryguy) ready to merge this PR but, we just need to know for certain that the columns that you're removing here are NOT referenced in any other code, including  ManageIQ/manageiq-providers-lenovo#132
```
+    remove_column :switches, :product_name,  :string
+    remove_column :switches, :part_number,   :string
+    remove_column :switches, :serial_number, :string
+    remove_column :switches, :description,   :string
+    remove_column :switches, :manufacturer,  :string
```
Please confirm, thanks

saulotoledo:
@gtanzillo @Fryguy @rodneyhbrown7 Yes, I confirm that.

The long answer: I created them to work in that PR only, but the same colums were created into asset_details table in another PR meanwhile. I just updated the https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132 code to use the asset_details table and those columns are not necessary anymore. There is no other place using them, not even the https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132 code.

Fryguy:
@saulotoledo I guess what Greg and I are asking is, if we merge the schema change, does the code in master currently use those removed columns, which could lead to a crash on the master?  Or, do I have to merge  ManageIQ/manageiq-providers-lenovo#132 into master simulatenously as I merge the schema change?

rodneyhbrown7:
@Fryguy There is currently no code that is using the columns that are being removed.  Those columns were added but no code was developed to put data into the column.

https://github.com/ManageIQ/manageiq-providers-lenovo/pull/132 is dependent on the addition of the asset_details column in this PR.

Fryguy:
Great.  Thanks for the clarification.

PR #183, opened 2018-03-29
Create a Physical Chassis table

caiocmpaes:
Creates the Physical Chassis table and adds a column in physical_servers.

### DIFF
--- a/db/migrate/20180327162308_create_physical_chassis.rb
+++ b/db/migrate/20180327162308_create_physical_chassis.rb
@@ -1,0 +1,24 @@
+class CreatePhysicalChassis < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_chassis do |t|
+      t.bigint :ems_id
+      t.string :uid_ems
+      t.string :ems_ref
+      t.bigint :physical_rack_id
+      t.string :name
+      t.string :vendor
+      t.string :type
+      t.string :location_led_state
+      t.string :health_state
+      t.string :overall_health_state
+      t.integer :management_module_slot_count
+      t.integer :switch_slot_count
+      t.integer :fan_slot_count
+      t.integer :blade_slot_count
+      t.integer :powersupply_slot_count
+      t.timestamps
+    end
+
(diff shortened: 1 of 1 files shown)


caiocmpaes:
@miq-bot assign @Fryguy

db/migrate/20180327162308_create_physical_chassis.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.string :ems_ref
+      t.bigint :physical_rack_id
+      t.string :name
+      t.string :vendor
+      t.string :type
+      t.string :location_led_state
+      t.string :health_state
+      t.string :overall_health_state
+      t.integer :management_module_slot_count
+      t.integer :switch_slot_count
```
agrare: I can see these being useful pulled out into their own model, that way you can track how many slots are populated and link switches/servers to which slot they're in.

caiocmpaes: I see your point. However, I think there is still the need for the slots_count in this table, because one thing is the number of slots on a Chassis, the other is how many servers/switches are connected.

One Chassis could have 6 slots for Switches (given number in switch_slot_count) and 4 switches plugged. If we get this by the Switches model, the only information is the 4 connected, not the total/open slots.

agrare: Agreed I was thinking there could be a slot record for every slot and could have `:populated => false` or something like that

agrare: Probably too much complexity for the first pass but would be interesting to extend it in the future

Fryguy:
I merged, but I'm wondering if we should add a type column for STI.  In the end we always need that.

@agrare Thoughts?

agrare:
@Fryguy there is one already, https://github.com/ManageIQ/manageiq-schema/pull/183/files#diff-9dca9fad9a5cac131257c8a6a8173679R10
I agree we always end up wanting one if there isn't

PR #184, opened 2018-04-05
Update provider_region to nil for Google provider

ZitaNemeckova:
Fixes UI failure after https://github.com/ManageIQ/manageiq-providers-google/pull/52 . 
To reproduce:
Make sure you have `ManageIQ::Providers::Google::CloudManager` or `ManageIQ::Providers::Google::NetworkManager` with `provider_region` **not** set to `nil`.
Go to summary page of said provider.
Before:
image
After:
image

With both providers not having `description` method but having historically set `provider_region` following lines are problem solved by removing values from `provider_region`.
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/helpers/ems_cloud_helper/textual_summary.rb#L42
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/helpers/ems_network_helper/textual_summary.rb#L40
 
@miq-bot add_label wip, bug, blocker

https://bugzilla.redhat.com/show_bug.cgi?id=1563600

Fix for Gaprindashvili: https://github.com/ManageIQ/manageiq-ui-classic/pull/3735

[More info](https://github.com/ManageIQ/manageiq-schema/pull/184#pullrequestreview-110086534)

### DIFF
--- a/db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
+++ b/db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
@@ -1,0 +1,11 @@
+class RemoveProviderRegionForGoogleProvider < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Removing provider_region from ManageIQ::Providers::Google::CloudManager") do
+      ExtManagementSystem.where(:type => 'ManageIQ::Providers::Google::CloudManager').update_all(:provider_region => nil)
+    end
+  end
+end
--- a/db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
+++ b/db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
@@ -7,5 +7,8 @@ class RemoveProviderRegionForGoogleProvider < ActiveRecord::Migration[5.0]
     say_with_time("Removing provider_region from ManageIQ::Providers::Google::CloudManager") do
       ExtManagementSystem.where(:type => 'ManageIQ::Providers::Google::CloudManager').update_all(:provider_region => nil)
     end
+    say_with_time("Removing provider_region from ManageIQ::Providers::Google::NetworkManager") do
+      ExtManagementSystem.where(:type => 'ManageIQ::Providers::Google::NetworkManager').update_all(:provider_region => nil)
+    end
   end
 end
--- a/spec/migrations/20180405145642_remove_provider_region_for_google_provider_spec.rb
+++ b/spec/migrations/20180405145642_remove_provider_region_for_google_provider_spec.rb
@@ -15,11 +15,21 @@ describe RemoveProviderRegionForGoogleProvider do
                        :provider_region => 42)
       ems_stub.create!(:type            => 'ManageIQ::Providers::Vmware::CloudManager',
                        :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Google::NetworkManager',
+                       :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Openstack::NetworkManager',
+                       :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Azure::NetworkManager',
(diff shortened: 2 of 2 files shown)


ZitaNemeckova:
@miq-bot remove_label wip

db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveProviderRegionForGoogleProvider < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Removing provider_region from ManageIQ::Providers::Google::CloudManager") do
+      ExtManagementSystem.where(:type => 'ManageIQ::Providers::Google::CloudManager').each do |provider|
```
lpichler: @ZitaNemeckova 
I think that we do it in one query.
```ruby
ExtManagementSystem.where(:type => 'ManageIQ::Providers::Google::CloudManager').update_all(:provider_region => nil)
```

spec/migrations/20180405145642_remove_provider_region_for_google_provider_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Amazon::CloudManager',
+                       :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Azure::CloudManager',
+                       :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Openstack::CloudManager',
+                       :provider_region => 42)
+      ems_stub.create!(:type            => 'ManageIQ::Providers::Vmware::CloudManager',
+                       :provider_region => 42)
+
+      migrate
```
lpichler: @ZitaNemeckova 

- I would join your `its` to one it with `expect(` because it would be nice to keep one `it` case for one migrating process and adapt 'it' message.

- I would also prefer creating object in lets although it is not needed in this case if there will be only one it.
- move `migrate` command to the `it` and remove before block.

What do you think about this suggestions ?

db/migrate/20180405145642_remove_provider_region_for_google_provider.rb
ZitaNemeckova: @lpichler I did but miq-bot complained about `update_all`. I don't mind changing it back if miq-bot's complain isn't important.

spec/migrations/20180405145642_remove_provider_region_for_google_provider_spec.rb
```diff
@@ -0,0 +1,25 @@
+require_migration
+
+describe RemoveProviderRegionForGoogleProvider do
+  migration_context :up do
+    let(:ems_stub) { migration_stub :ExtManagementSystem }
+
+    ems_stub.create!(:type            => 'ManageIQ::Providers::Google::CloudManager',
```
tumido: Please wrap these calls into `before` block

lpichler: prefer let before `before` block

tumido: I meant the `ems_stub.create!` calls. You want to wrap them in a `let` call?

h-kataria:
@ZitaNemeckova just a note: this PR contains changes to schema, so can not be backported to Gaprindashvili or Fine branches, will need a fix in the code to handle the issue on those branches.

lpichler:
@miq-bot assign @Fryguy

tumido:
@h-kataria, I don't think it changes the schema, only data for one (or two possibly) specific `ems`es

@ZitaNemeckova, I wonder, is this change needed only for `Google::CloudManager`? Don't we need this for `Google::NetworkManager` manager as well?

Fryguy:
Are you expecting this to be backported?  We can't backport migrations.

h-kataria:
@Fryguy UI PR with workaround for G branch https://github.com/ManageIQ/manageiq-ui-classic/pull/3735

Fryguy:
And there's the answer...thanks GitHub for not refreshing ;)

PR #185, opened 2018-04-09
Creating physical_network_ports table

douglasgabriel:
__This PR is able to:__
- Create `physical_network_ports` table;

__Goal:__
Actually the ports are saved as a `GuestDevice` instance that should only represent elements that are plugged into a mother board. With this new table we avoid the pollution of `guest_devices` table.

### DIFF
--- a/db/migrate/20180409120422_create_physical_network_ports.rb
+++ b/db/migrate/20180409120422_create_physical_network_ports.rb
@@ -1,0 +1,20 @@
+class CreatePhysicalNetworkPorts < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_network_ports do |t|
+      t.string  :ems_ref
+      t.string  :uid_ems
+      t.string  :type
+      t.string  :port_name
+      t.string  :port_type
+      t.string  :peer_mac_address
+      t.string  :vlan_key
+      t.string  :mac_address
+      t.integer :port_index
+      t.boolean :vlan_enabled
+      t.bigint  :guest_device_id
+      t.bigint  :switch_id
+
+      t.timestamps
+    end
+  end
+end


douglasgabriel:
Hey @Fryguy could you please take a look on this? @miq-bot assign @Fryguy

douglasgabriel:
@miq-bot assign @Fryguy

Fryguy:
@agrare please do a first-pass review?

db/migrate/20180409120422_create_physical_network_ports.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePhysicalNetworkPorts < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_network_ports do |t|
+      t.bigint  :ems_ref
+      t.string  :uid_ems
+      t.string  :port_name
+      t.string  :port_type
+      t.string  :peer_mac_address
+      t.string  :vlan_key
+      t.string  :address
```
agrare: Does it make sense that a port would have an address?  This is already on the GuestDevice model and I think of a port as a layer 1 concept and mixing address pulls in layer 3

db/migrate/20180409120422_create_physical_network_ports.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePhysicalNetworkPorts < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_network_ports do |t|
```
agrare: Can this be just `network_ports` instead of physical_?

agrare:
@douglasgabriel is the goal to have ports on the guest_device side as well as the switch side?

douglasgabriel:
Yes, it is

db/migrate/20180409120422_create_physical_network_ports.rb
douglasgabriel: `network_ports` table already exist, but seems like that its semantics for the system is not what we are looking for, once it has so many relationships, maybe you can tell it more accurately. The intention of this new table is really keep data about the physical ports (connections), so I think that the `physical_` prefix is required to make more clear this table semantic.

db/migrate/20180409120422_create_physical_network_ports.rb
douglasgabriel: Actually this field is about the port mac address, so I think that is better to change the field name to `mac_address`

rodneyhbrown7:
@agrare  Ideally we want all of the switch ports for both the physical server and the switch to live in a ports table with associations back to the parent device as appropriate.  The network_ports table is the right concept but we thought that we would overload the meaning a great deal if we added our physical ports ot it.

agrare:
@douglasgabriel @rodneyhbrown7 this should have a `switch_id` column then.

@Fryguy what do you think about re-using the `network_ports` model?

rodneyhbrown7:
@agrare agreed, each entry should then have either an association back to a switch or an association back to a guest_device.

agrare:
@douglasgabriel thanks for adding the switch_id
@Fryguy this looks okay to me but I'll leave it up to you if we would rather use the network_ports table instead of creating a new one.  I'm leaning towards trying to reuse the existing table since there is some overlap in the columns (ems_ref, mac_address) and I don't think the new columns are really specific to physical (vlan, peer_mac_address)

gtanzillo:
I'm ok either way - separate table or reusing the `network_ports` table. I'm not very familiar with the differences between physical and virtual network ports to have an opinion either way. I'll leave it to @Fryguy to have the final say.

db/migrate/20180409120422_create_physical_network_ports.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.string  :uid_ems
+      t.string  :port_name
+      t.string  :port_type
+      t.string  :peer_mac_address
+      t.string  :vlan_key
+      t.string  :mac_address
+      t.integer :port_index
+      t.boolean :vlan_enabled
+      t.bigint  :guest_device_id
+      t.bigint  :switch_id
```
Fryguy: is this relating to the switches table or the physical_switches table?  If the latter, this should be physical_switch_id.

Fryguy:
I'm ok with the separate table.  I'm not sure if overloading the existing table will have any other issues.

db/migrate/20180409120422_create_physical_network_ports.rb
Fryguy: Oh nevermind...For some reason I thought there was a physical_switches table separate from switches.

Fryguy:
@agrare I think I asked this also for a different table, but should we introduce STI now?

agrare:
> I think I asked this also for a different table, but should we introduce STI now?

Yeah I think that was the physical_chassis table, I'm always for adding it...not really any downside

douglasgabriel:
done @Fryguy , Let me know if there is some more thing to address here

Fryguy:
@douglasgabriel As @agrare and I were discussing above, can you also add a `type` column to introduce STI now?  Otherwise, this looks good to go.

PR #187, opened 2018-04-12
Add configuration_script_sources.last_update_error

Glutexo:
Add new `last_update_stdout` column to `configuration_script_sources` table. This allows to store last update stdout to display more information about why its refresh failed.

Fetching and presenting this information is a workaround for cloning repositories from HTTPS servers with untrusted SSL certificates. Tower API does not offer an option to disable SSL check, so with the stdout user can at least see, why the cloning actually failed.

https://bugzilla.redhat.com/show_bug.cgi?id=1513616

Followed by adding the column to model: [#17290 in `manageiq`](https://github.com/ManageIQ/manageiq/pull/17290).

/cc @jameswnl for review

### DIFF
--- a/db/migrate/20180329141018_add_last_update_error_to_configuration_script_source.rb
+++ b/db/migrate/20180329141018_add_last_update_error_to_configuration_script_source.rb
@@ -1,0 +1,5 @@
+class AddLastUpdateErrorToConfigurationScriptSource < ActiveRecord::Migration[5.0]
+  def change
+    add_column :configuration_script_sources, :last_update_error, :text
+  end
+end


Glutexo:
@miq-bot add_reviewer @jameswnl 

Can this get merged, please? There are no dependencies for this and it has been hanging there for a long time.

Glutexo:
@Ladas Typical Ansible outputs are a few (say, 1-15) kilobytes. Since `TEXT` columns are ([almost](https://www.postgresql.org/docs/9.5/static/datatype-character.html)) unlimited and the stdout is text (not binary), I’d stick with this type. But maybe there are some implications I’m not aware about?

jameswnl:
@miq-bot add_reviewer @Fryguy

Glutexo:
Just pinging @Fryguy. Is this PR ok to be merged? There are other PRs that depend on this one. See https://github.com/ManageIQ/manageiq/issues/17307

Glutexo:
@Fryguy Hey, may I ask you for a review/merge of this PR? It’s required by https://github.com/ManageIQ/manageiq/pull/17290 and through that by other PRs too. Thanks!

db/migrate/20180329141018_add_last_update_stdout_to_configuration_script_source.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastUpdateStdoutToConfigurationScriptSource < ActiveRecord::Migration[5.0]
+  def change
+    add_column :configuration_script_sources, :last_update_stdout, :text
```
Fryguy: @bdunne Aside from my naming questions in https://github.com/ManageIQ/manageiq/pull/17290#issuecomment-395879777, does this even make sense on configuration_script_sources?  I thought sources was the content of the configuration script.  But content is not invocation, and I'd expect an "output" field on the "runtime invocation" model.

bdunne: See https://github.com/ManageIQ/manageiq/pull/17290#issuecomment-395887146 maybe we should continue the conversation over there?

Glutexo: We still store some data concerning the last update: status and timestamp. The only difference is that in Tower these are propagated to the project itself, so they are accessible even without explicitly asking for the last update resource.

Glutexo:
@Fryguy Renamed to last_update_error. Ready to merge?

PR #192, opened 2018-04-26
Azure backslash to forward slash

djberg96:
Data migration to go along with https://github.com/ManageIQ/manageiq-providers-azure/pull/245 and https://github.com/ManageIQ/manageiq/pull/17337.

~~WIP for now because it's not finished and because it's causing core dumps somehow.~~

### DIFF
--- a/db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
+++ b/db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
@@ -1,0 +1,73 @@
+class AzureBackslashToForwardSlash < ActiveRecord::Migration[5.0]
+  class Vm < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStackOutput < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStackParameter < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStackResource < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class EventStream < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Updating Azure VM ems_ref, uid_ems and description") do
+      Vm.where(:type => "ManageIQ::Providers::Azure::CloudManager::Vm")
+        .update_all("
+          ems_ref = replace(ems_ref, '\\', '/'),
+          uid_ems = replace(uid_ems, '\\', '/'),
+          description = replace(description, '\\', '/')
+        ")
+    end
+
+    say_with_time("Updating Azure template ems_ref, uid_ems and description") do
+      Vm.where(:type => "ManageIQ::Providers::Azure::CloudManager::Template")
+        .update_all("
+          ems_ref = replace(ems_ref, '\\', '/'),
+          uid_ems = replace(uid_ems, '\\', '/'),
+          description = replace(description, '\\', '/')
+        ")
+    end
+
+    say_with_time("Updating Azure orchestration stack ems_ref") do
+      OrchestrationStack
+        .where(:type => "ManageIQ::Providers::Azure::CloudManager::OrchestrationStack")
+        .update_all("ems_ref = replace(ems_ref, '\\', '/'), description = replace(description, '\\', '/')")
+    end
+
+    say_with_time("Updating Azure orchestration stack output ems_ref") do
+      OrchestrationStackOutput
+        .where(:stack_id => OrchestrationStack.where(:type => "ManageIQ::Providers::Azure::CloudManager::OrchestrationStack"))
+        .update_all("ems_ref = replace(ems_ref, '\\', '/')")
+    end
+
+    say_with_time("Updating Azure orchestration stack parameter ems_ref") do
+      OrchestrationStackParameter
+        .where(:stack_id => OrchestrationStack.where(:type => "ManageIQ::Providers::Azure::CloudManager::OrchestrationStack"))
+        .update_all("ems_ref = replace(ems_ref, '\\', '/')")
+    end
+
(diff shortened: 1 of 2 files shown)


spec/migrations/20180424141617_azure_backslash_to_forward_slash_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    stack_output_stub.create!(
+      :stack_id => 1,
+      :key      => 'adminUserName',
+      :value    => 'me',
+      :ems_ref  => '/subscriptions/xyz/resourceGroups/some_group/providers/Microsoft.Resources/deployments/RedHat\adminUsername'
+    )
+  end
+
+  let!(:stack_param) do
+    stack_param.create!(
```
NickLaMuro: @djberg96 This is a infinite recursion loop here.  Is it possible that this is causing your core dump?

NickLaMuro: I changed this to `stack_output_stub.create!` instead of `stack_param.create!`, and it seemed to allow the tests to pass.

djberg96: Yep, that was it. Will update the stub, thanks.

db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
```diff
@@ -0,0 +1,75 @@
...
+  class Vm < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStackOutput < ActiveRecord::Base
+    belongs_to :stack, :class_name => 'OrchestrationStack'
```
Fryguy: class_name should be fully qualified with the migration namespac.  e.g. AzureBackslashToForwardSlash::OrchestrationStack

db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  class OrchestrationStackTemplate < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    Vm.where(:type => "ManageIQ::Providers::Azure::CloudManager::Vm").each do |vm|
+      vm.ems_ref = vm.ems_ref.tr("\\", '/')
+      vm.uid_ems = vm.ems_ref.tr("\\", '/')
+      vm.description = vm.description.tr("\\", '/')
```
Fryguy: I think we can do all of this with a pure SQL query, which might save a lot of iterations and cycles... see https://www.postgresql.org/docs/9.5/static/functions-string.html

db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
```diff
@@ -0,0 +1,61 @@
...
+  class Vm < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class OrchestrationStackOutput < ActiveRecord::Base
+    belongs_to :stack, :class_name => 'AzureBackslashToForwardSlash::OrchestrationStack'
```
Fryguy: Remove these belongs_to statements now that the relation is no longer used.

spec/migrations/20180424141617_azure_backslash_to_forward_slash_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      :name        => 'azure_test_template',
+      :ems_ref     => 'xyz\some_group\microsoft.compute/virtualmachines\bar',
+      :type        => 'ManageIQ::Providers::Azure::CloudManager::Template',
+      :description => 'some\other\description'
+    )
+  end
+
+  let!(:stack) do
+    stack_stub.create!(
+      :id      => 1,
```
Fryguy: Do not hardcode ids in tests as it circumvents the ability to test with a random region id.

spec/migrations/20180424141617_azure_backslash_to_forward_slash_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+describe AzureBackslashToForwardSlash do
+  let(:vm_stub) { migration_stub :Vm }
+  let(:stack_stub) { migration_stub :OrchestrationStack }
+  let(:stack_output_stub) { migration_stub :OrchestrationStackOutput }
+  let(:stack_param_stub) { migration_stub :OrchestrationStackParameter }
+  let(:stack_resource_stub) { migration_stub :OrchestrationStackResource }
+  let(:event_stream_stub) { migration_stub :EventStream }
+
+  let!(:vm) do
```
Fryguy: test-wise this is really inefficient because *all* of the `let!` are created for every test.  Since each one of these is only used in one test, you could also just not use `let`s at all, and just put the record creation before the `migrate` in each test.  This also makes it easier to read the test because the object is created/migrated/checked all in the same place in the test block.

djberg96:
@Fryguy Ok, added `say_with_time`, and removed most of the hard coded id's, and most of the `let!` blocks were changed into local variables.

spec/migrations/20180424141617_azure_backslash_to_forward_slash_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+        expect(template.ems_ref).to eql('xyz/some_group/microsoft.compute/virtualmachines/bar')
+        expect(template.uid_ems).to eql('xyz/some_group/microsoft.compute/virtualmachines/bar')
+        expect(template.description).to eql('some/other/description')
+      end
+    end
+
+    context "orchestration" do
+      let!(:stack) do
+        stack_stub.create!(
+          :id      => 1,
```
Fryguy: Do not hardcode ids

Fryguy:
@djberg96 Ignore the rubocops about update_all...I gotta remove those from this repo :)

Fryguy:
Ah crap, I forgot that this should be merged with  ManageIQ/manageiq-providers-azure#245 .  I should set up a label for those cases :(

PR #194, opened 2018-04-26
[WIP] Add a migration to document the tables in the current schema

bdunne:


db/migrate/20180424220424_add_table_comments.rb
```diff
@@ -0,0 +1,323 @@
+class AddTableComments < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :accounts, ""
+    change_table_comment :advanced_settings, ""
+    change_table_comment :arbitration_profiles, ""
+    change_table_comment :arbitration_rules, ""
+    change_table_comment :arbitration_settings, ""
+    change_table_comment :assigned_server_roles, "Internal: Join table to show relation between MiqServers and their ServerRoles"
+    change_table_comment :audit_events, ""
+    change_table_comment :authentications, "Internal: A safe place to store credentials used to authenticate with other services (reversible encrypted)"
```
Fryguy: Not internal...we partially(?) expose this via the API I believe so users can get back their SSH keys, for example

bdunne: Maybe "Internal" is not the right term.  I wanted to differentiate between things and concepts that are ours vs things that represent something from a provider / manager.  Also, maybe there are more than just these two types of things.

bdunne:
Also, I'm going to need some help describing all of the things.  Looks like there are a lot of models that I don't know the purpose of.

PR #195, opened 2018-04-27
Migrate evm dba

juliancheal:
Used my code extraction library to migrate over evm_dba.rake.

Fixes issue https://github.com/ManageIQ/manageiq-schema/issues/3

### DIFF
--- a/lib/tasks/evm_dba.rake
+++ b/lib/tasks/evm_dba.rake
@@ -1,0 +1,260 @@
+require 'awesome_spawn'
+require 'evm_rake_helper'
+
+# TODO: move into DatabaseYml
+# TODO: can we use EvmDatabseOps directly?
+module EvmDba
+  def self.database_configuration_file
+    File.expand_path(File.join(Rails.root, 'config', 'database.yml'))
+  end
+
+  def self.load_config
+    require 'yaml'
+    YAML::load((IO.read(self.database_configuration_file)))
+  end
+
+  def self.local?
+    config = self.load_config[Rails.env]
+    return false unless config['adapter'] == 'postgresql'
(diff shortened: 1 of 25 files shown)


lib/tasks/evm_dba.rake
```diff
@@ -0,0 +1,260 @@
...
+        opt :reindex,    "Reindex database (or table if --table specified)"
+        opt :analyze,    "Vacuum with analyze"
+        opt :full,       "Vacuum full"
+        opt :verbose,    "Vacuum with verbose information printed"
+
+        opt :table,      "Tablename to reindex (if only perorm on one)", :type => :string
+      end
+
+      opts = opts.delete_if { |_k, v| v.nil? || v == false }
+      EvmDatabaseOps.gc(opts)
```
Fryguy: EvmDatabaseOps should probably come over as well.

Fryguy:
I generally like this idea, but want input from @bdunne / @jrafanie / @carbonin

lib/tasks/evm_dba.rake
carbonin: There are a handful of places where we use models in that file. Are we okay with those being in this repo?

jrafanie:
hmm, I like the idea of moving things where they belong but there's something weird about moving gc/backup/restore/postgresql stop/start/database configuration/ etc. to a schema repo.  They're only vaguely related in that they're for the same postgresql database.  For example, why not bring over postgresql HA stuff, our pg stat activity stuff and other PG related things here?  I can't say where they belong but it feels unrelated to schema.  Perhaps the wrapper for reset and migrate as they're more schema related but the others don't seem to fit.

PR #198, opened 2018-05-07
Clean Azure image names

djberg96:
Due to a quirk of File.dirname, the Azure provider can end up with leading "./" in the image name. This looks dumb and should be fixed.

Should be merged in conjunction with https://github.com/ManageIQ/manageiq-providers-azure/pull/240

### DIFF
--- a/db/migrate/20180507134810_azure_normalize_image_name.rb
+++ b/db/migrate/20180507134810_azure_normalize_image_name.rb
@@ -1,0 +1,20 @@
+class AzureNormalizeImageName < ActiveRecord::Migration[5.0]
+  class VmOrTemplate < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    self.table_name = 'vms'
+  end
+
+  def up
+    say_with_time("Updating Azure template name") do
+      VmOrTemplate.where("type = ? and uid_ems like ?", "ManageIQ::Providers::Azure::CloudManager::Template", "http%")
+        .update_all("name = regexp_replace(name, '^./', '')")
+    end
+  end
+
+  def down
+    say_with_time("Reverting Azure template name") do
+      VmOrTemplate.where("type = ? and uid_ems like ?", "ManageIQ::Providers::Azure::CloudManager::Template", "http%")
+        .update_all("name = './' || name")
+    end
+  end
+end
--- a/spec/migrations/20180507134810_azure_normalize_image_name_spec.rb
+++ b/spec/migrations/20180507134810_azure_normalize_image_name_spec.rb
@@ -1,0 +1,87 @@
+require_migration
+
+describe AzureNormalizeImageName do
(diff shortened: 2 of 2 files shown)


db/migrate/20180507134810_azure_normalize_image_name.rb
```diff
@@ -0,0 +1,12 @@
+class AzureNormalizeImageName < ActiveRecord::Migration[5.0]
+  class Vm < ActiveRecord::Base
```
bronaghs: Should this be `class VmOrTemplate < ActiveRecord::Base`?

db/migrate/20180507134810_azure_normalize_image_name.rb
```diff
@@ -0,0 +1,12 @@
...
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Updating Azure template name") do
+      Vm.where(:type => "ManageIQ::Providers::Azure::CloudManager::Template")
```
bronaghs: Should this be `VmOrTemplate.where(...)`?

djberg96: Works either way afaik, since I'm specifying the type.

db/migrate/20180507134810_azure_normalize_image_name.rb
```diff
@@ -0,0 +1,12 @@
...
+        .update_all("name = regexp_replace(name, '^./', '')")
+    end
+  end
```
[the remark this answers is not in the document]
djberg96: It's not something that can be rolled back consistently I don't think, so no.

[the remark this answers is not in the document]
bronaghs: I think it is risky not to add one and writing one isn't a large effort. One of the scenarios I could see as being a problem is if the user does not back up the DB before an upgrade and the upgrade has to be undone.

db/migrate/20180507134810_azure_normalize_image_name.rb
bronaghs: I think templates will not be returned by `Vm.where()`, I think it requires `VmOrTemplate`. Can you check on the console?

djberg96:
@bronaghs ok, changed it to VmOrTemplate.

Fryguy:
> Should be merged in conjunction with ManageIQ/manageiq-providers-azure#239

except that's an issue and not a PR...do you have a PR on the other side?

Fryguy:
Waiting on merge until the other side is ready

bronaghs:
@Fryguy - I had asked Dan to add a down method or replace the `up` with a `change` method. Do you not think this is necessary?

djberg96:
@bronaghs @Fryguy ok, added a `down` method. At first I didn't think it was practical, but then I remembered this only affects unmanaged images, so I can check the uid_ems.

PR #201, opened 2018-05-15
Optimize migration which creates task for each job

yrudman:
There are several long running migrations observed on client's DB (`Job.count` around 1.3 millions)  : 
`CreateTaskForEachJobAndTransferAttributes`        - 4863 seconds
`CopyAgentIdToMiqServerIdInJobsTable`                  -  168 seconds
`CopyServerIdFromJobsToMiqTask`                           - 1645

In UI (Task management screen) we show only tasks for the last 6 days, so there is no need to do anything with jobs/task older than 7 days.

This PR:  changed exiting `CreateTaskForEachJobAndTransferAttributes` by:
- deleting finished `Job` older than 7 days first and creating task for each job after that.
- deleting finished `MiqTask` older than 7 days. This change will no speed-up migration, it is just clen-up and done t be consistent with `Job` deletion. 

on the same client's DB deleting old records from `Job` table took 16 seconds:
```
>> Job.where("updated_on < ?", t).where("state = 'finished'").delete_all
  SQL (16330.1ms)  DELETE FROM "jobs" WHERE (updated_on < '2018-04-14 16:05:23.679929') AND (state = 'finished')
=> 1232220
```

Depends on https://github.com/ManageIQ/manageiq/pull/17727 for backports

### DIFF
--- a/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
+++ b/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
@@ -8,13 +8,28 @@ class CreateTaskForEachJobAndTransferAttributes < ActiveRecord::Migration[5.0]
     has_one :job, :class_name => 'CreateTaskForEachJobAndTransferAttributes::Job'
   end
 
+  class LogFile < ActiveRecord::Base
+  end
+
+  class BinaryBlob < ActiveRecord::Base
+  end
+
   def up
     say_with_time("Deleting finished jobs older than 7 days") do
-      Job.where("updated_on < ?", Time.now.utc - 7.days).where("state = 'finished'").delete_all
+      Job.where("updated_on < ?", 7.days.ago.utc).where("state = 'finished'").delete_all
     end
+
+    say_with_time("Deleting logs linked to finished tasks older than 7 days") do
+      LogFile.where(:miq_task_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
+    end
+    say_with_time("Deleting BinaryBlobs linked to finished tasks older than 7 days") do
+      BinaryBlob.where(:resource_type => 'MiqTask', :resource_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
+    end
+
     say_with_time("Deleting finished tasks older than 7 days") do
-      MiqTask.where("updated_on < ?", Time.now.utc - 7.days).where("state = 'Finished'").delete_all
+      MiqTask.where("updated_on < ?", 7.days.ago.utc).where("state = 'Finished'").delete_all
     end
+
     say_with_time("Creating tasks associated with jobs") do
       Job.find_each do |job|
         job.create_miq_task(:status        => job.status.try(:capitalize),
--- a/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
+++ b/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
@@ -9,6 +9,9 @@ class CreateTaskForEachJobAndTransferAttributes < ActiveRecord::Migration[5.0]
   end
 
   def up
+    say_with_time("Deleting finished jobs older than 7 days") do
+      Job.where("updated_on < ?", Time.now.utc - 7.days).where("state = 'finished'").delete_all
+    end
     say_with_time("Creating tasks associated with jobs") do
       Job.find_each do |job|
         job.create_miq_task(:status        => job.status.try(:capitalize),
--- a/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
+++ b/db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
@@ -15,19 +15,20 @@ class CreateTaskForEachJobAndTransferAttributes < ActiveRecord::Migration[5.0]
   end
 
   def up
+    purge_date = 7.days.ago.utc
     say_with_time("Deleting finished jobs older than 7 days") do
-      Job.where("updated_on < ?", 7.days.ago.utc).where("state = 'finished'").delete_all
+      Job.where("updated_on < ?", purge_date).where("state = 'finished'").delete_all
     end
 
+    old_finished_tasks = MiqTask.where("updated_on < ?", purge_date).where("state = 'Finished'")
     say_with_time("Deleting logs linked to finished tasks older than 7 days") do
-      LogFile.where(:miq_task_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
+      LogFile.where(:miq_task_id => old_finished_tasks.select(:id)).delete_all
     end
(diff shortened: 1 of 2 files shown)


Fryguy:
@kbrock please review for performance.

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
```diff
@@ -9,6 +9,12 @@ class MiqTask < ActiveRecord::Base
   end
 
   def up
+    say_with_time("Deleting finished jobs older than 7 days") do
+      Job.where("updated_on < ?", Time.now.utc - 7.days).where("state = 'finished'").delete_all
```
jrafanie: I prefer `7.days.ago.utc` to `Time.now.utc - 7.days`

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
```diff
@@ -9,6 +9,12 @@ class MiqTask < ActiveRecord::Base
...
+    end
+    say_with_time("Deleting finished tasks older than 7 days") do
+      MiqTask.where("updated_on < ?", Time.now.utc - 7.days).where("state = 'Finished'").delete_all
```
jrafanie: 7.days.ago.utc

jrafanie:
Looks good to me.  Do we care that some people will have their tasks/jobs tables cleared and others will not?  (they ran this migration previously)  

@kbrock can you review?

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
kbrock: +1 on the `utc` there - I think it may work without, but there are strange edge cases

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
kbrock: Will any of these have associated logs or blobs
```ruby
class MiqTask
  has_one :log_file, :dependent => :destroy
  has_one :binary_blob, :as => :resource, :dependent => :destroy
  # ...
  has_one :job, :dependent => :destroy
end
```

Do we need something like
```ruby
LogFile.where(:miq_task_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
BinaryBlob.where(:resource_type => 'MiqTask', :resource_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
```

(code not tested)

yrudman: 👍

yrudman:
@kbrock ready for the second review

yrudman:
@jrafanie @Fryguy @kbrock could you check if it is 👍 ?

yrudman:
I think originally we did not clean-up tasks since report were not shown if associated task removed.
It was fixed in https://github.com/ManageIQ/manageiq-ui-classic/pull/1965https://github.com/ManageIQ/manageiq-ui-classic/pull/1965 and https://github.com/ManageIQ/manageiq-ui-classic/pull/1488

7 days limit was picked-up because we are not showing in UI tasks with latest update older than 6 days:
screen shot 2018-05-30 at 2 53 24 pm

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
```diff
@@ -8,7 +8,28 @@ class MiqTask < ActiveRecord::Base
...
+
+    say_with_time("Deleting logs linked to finished tasks older than 7 days") do
+      LogFile.where(:miq_task_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
+    end
+    say_with_time("Deleting BinaryBlobs linked to finished tasks older than 7 days") do
+      BinaryBlob.where(:resource_type => 'MiqTask', :resource_id => MiqTask.where("updated_on < ?", 7.days.ago.utc).select(:id)).delete_all
+    end
+
+    say_with_time("Deleting finished tasks older than 7 days") do
+      MiqTask.where("updated_on < ?", 7.days.ago.utc).where("state = 'Finished'").delete_all
```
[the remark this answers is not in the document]
Fryguy: Minor, but I would move that out to a variable at the top.  If any of these actions take some time, then it's a moving window, which is something we don't want as it could create orphans.

Fryguy:
Something feels off about deleting old tasks.  Or put another way, this feels like a "purge" of tasks, but the value being chosen is "7 days" and it only happens once during this migration and then never again.  I think we need @bascar and @chessbyte's opinion here.

gtanzillo:
@jdeubel Regarding https://github.com/ManageIQ/manageiq-schema/pull/201#issuecomment-393623651, do you have a feel for whether purging jobs older than 7 days would have and impact on customers?

jdeubel:
@gtanzillo If that is the max amount of time that a user can see then that is fine with me.

jrafanie:
> @gtanzillo If that is the max amount of time that a user can see then that is fine with me.

yeah, I agree, if we choose what max time in the pull down for how far "back" you can go to see old tasks/jobs throughout the UI, there is very little risk to removing older records.

Fryguy:
I would prefer we add purging for MiqTasks defaulting to 7 days in manageiq core, to avoid growing the table *after* the migration.  Then, this PR would be fine (or perhaps deleting old should be a separate PR dated one before this one, but if we need to backport, putting them in one PR is fine)

kbrock:
I merged the other pruning PR associated with this one.

@Fryguy this good to go?

db/migrate/20170120164058_create_task_for_each_job_and_transfer_attributes.rb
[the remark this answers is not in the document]
Fryguy: I think the `.where("state = 'Finished'")` needs to apply to the conditionals above for LogFile and BinaryBlob.  Perhaps extracts the query fragment out to a variable and reuse it.

Fryguy:
One last thing...the date on this PR is *really* old...can you modify the file names to be a more recent date?  Technically it should be fine, but it can be confusing because we merged the purging PR only recently

yrudman:
@Fryguy as we discussed - it ot optimization PR which is editing existing migration and will be backported together with https://github.com/ManageIQ/manageiq/pull/17727

Fryguy:
@yrudman Thanks, I forgot this was an edit :+1:

yrudman:
@miq-bot add-label blocker

simaishi:

```
$ git log -1
commit 44ade48a30ce898b4d5a456fbcbd5f817d379133
Date:   Tue Aug 7 12:26:31 2018 -0400

    Merge pull request #201 from yrudman/delete-old-job-while-creating-linked-task
    
    Optimize migration which creates task for each job
    (cherry picked from commit 8d48f1bedf9c42661ecb83ae3ab5ab9fb2b21545)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1618808
```

PR #202, opened 2018-05-17
Creating Switches group migration

EsdrasVP:
This PR is able to:
- Create a Switch group migration
- Add PhysicalSwitches to Switches group

This change enable Switches to share its table with other implementations like PhysicalSwitches. This was necessary because making Switch as base model for PhysicalSwitch produced strange behaviors.


EsdrasVP:
This was resolved with a better solution, so I'll close this PR.

PR #205, opened 2018-05-24
Add the severity column to event_streams table

CharlleDaniel:
This pr is able to:
* Add a new column in the event_streams table and allow the providers access the severity information more quickly.

This is an advantage because a physical infra provider have about 5000 events and is a lot expensive to access each event and recover the severity inside the full_data column.

### DIFF
--- a/db/migrate/20180524133942_add_severity_to_events.rb
+++ b/db/migrate/20180524133942_add_severity_to_events.rb
@@ -1,0 +1,5 @@
+class AddSeverityToEvents < ActiveRecord::Migration[5.0]
+  def change
+    add_column :event_streams, :severity, :string
+  end
+end


CharlleDaniel:
@agrare Could you review this PR?

agrare:
> a physical infra provider have about 5000 events

You mean a physical infra provider has ~5000 types of events?  I don't see how this is related to the number of events which were actually raised by the provider which will change over time.

Where are you accessing `full_data[:severity]` from a large number of events in code?  I wasn't able to find it except in MiqAlert, but that pulls severity, url, and resolved out of full_data so I assume this isn't the place that you're talking about.

agrare:
Spoke to @saulotoledo offline and he said this isn't something which is used now but could be used in the future and I got hung up trying to find where it already used.

This looks fine to me, we should have some well defined list of severities in `EventStream` so this isn't just some free-form field.

agrare:
@miq-bot assign @Fryguy

CharlleDaniel:
@agrare @saulotoledo  So, we don't have ~5000 event types, but ~5000 events inside the event_streams table for each physical provider registered. Today we aren't using this information and I will create a new PR in the UI:Classic to see this information in the Timeline page and maybe in the API to allow us to filter the event instances by severity for both. By severity I mean something like Critical (e.g. server unreachable), Warning (e.g. server using more than 70% resources), Info (e.g. database has started) and others.

CharlleDaniel:
@agrare Just a followup I thought about using the event_type but it is a little different from the severity, we can have many events with the type `power_off_activity` but a piece of these events the severity can be `Critical` and the remaining would be `Informational`. I'm just trying to simplify the access to this information and we will use the filters in the future. If you know another way to do this let me know, I appreciate that :+1: .

CharlleDaniel:
@agrare I closed this PR, because we not need any more add a severity after the PR [17582](https://github.com/ManageIQ/manageiq/pull/17582) was merged. Now we can filter events in the API using attributes like as group, group level and group_name. Thank you guys.

PR #206, opened 2018-05-24
Creating physical_components table

douglasgabriel:
__This PR is able to__
- Create the `physical_components` table.

__Goal__
The goal is to generalize some physical components relationships as well as have a common way to describe these components.

In a first perspective, this way we solve a problem with the `event_streams` where we want to relate to the physical components (Servers, Switches, Chassis, etc), so instead to add a new column for each component in `event_streams` we just add a relation `event_streams -> physical_components` and all components could be bind to an event. Also, the MangeIQ API could return events with its related components in a common way, making more easier for the consumer to figure out the source component of an event.

__Future benefits__
Looking more forward we also can migrate other relationships to use this table, looking more like the following diagram:

image

- __`has_many` relationship between `PhysicalInfraManager` -> `PhysicalComponent`:__ This way we can grant that a `PhysicalInfraManager` can manage any physical component and a new component can be managed just creating a relation to `PhysicalComponent` (for example the upcoming component `PhysicalStorage`)

- __Add physical components common relationships:__
  - `asset_details`;
  - `computer_systems`.

### DIFF
--- a/db/migrate/20180524135736_create_physical_components.rb
+++ b/db/migrate/20180524135736_create_physical_components.rb
@@ -1,0 +1,10 @@
+class CreatePhysicalComponents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_components do |t|
+      t.belongs_to :component, :type => :bigint, :polymorphic => true
+      t.string     :ems_ref
+      t.timestamps :null => true
+      t.index      %i(component_id component_type)
+    end
+  end
+end


douglasgabriel:
@rodneyhbrown7 could take a look on this?

rodneyhbrown7:
Douglas, I think this looks good.  I was initially thinking that we could use ComputerSystem itself as a common parent, but that breaks down in the case of Physical Rack.   PhysicalRack is not a computer system, nor will it have a computer system like the others do.   I agree with this approach.

douglasgabriel:
@miq-bot assign @Fryguy

Fryguy:
@agrare Please do a first pass review.

agrare:
Hm @douglasgabriel I think one of the benefits of having relationships to each type is that an event might relate to a server and a switch so we could set both, with this approach we wouldn't be able to do that.  For example quite often an Infra event might have the vm and the host set in the event, of course it is possible to get the host from `event.vm.host` but that is only valid at the time of the event, where if it is represented in the event we have the historical info as well.

douglasgabriel:
Got it! But the current problem is that, to know which component an event is related you must search each relationship field to see which one is filled. This problem is growing along with the amount of components in our structure, so, IMHO, a common way to describe these component is welcome.

I've don't realized that an event could be bind to more than one component, but I think that in this case a `many_to_many` relationship could do the trick, don't you think?

agrare:
I definitely see your point but I think there are few issues with just moving to a generic many-to-many join table

1. That means we would have to do a query on the join table to find the resource type that you want (e.g. `event.resources.find_by(:resource_type => 'VmOrTemplate')` to replace `event.vm`
2. For some events we have src and dest of the same type (e.g. `event.host` and `event.dest_host`) so we'd have to add a column to track src or dest.

@Fryguy I'll leave it up to your judgement on adding a few more columns (`:physical_rack_id`, `:physical_switch_id`, `:physical_storage_id`, and `:physical_chassis_id` from what I can tell) or trying to save that and add one `:physical_component_id` to a polymorphic component.

gtanzillo:
@douglasgabriel, @Fryguy, @agrare After discussing this with @agrare  I think that a 30 min discussion could help get a mutual understanding of this and whether this it the best way to go. I'm not sure this can be merged without that. Thoughts? If all are in agreement, I could schedule it.

agrare:
@miq-bot assign @Fryguy

Fryguy:
There are problems with both approaches.  I am also concerned that not having a direct relationship will require changes to the existing timeline presentation code, since it will be inconsistent to know what to query when on a component's page.

In general, I feel like creating a hub table is going to be a bigger headache if we start tacking more stuff onto it, but @gtanzillo please set up the meeting to discuss.

douglasgabriel:
Closing this as was recommended on 06/27 meeting. Mainly because the large refactor that it would demand on timeline portion and some performance issues that it could provide.

PR #207, opened 2018-05-25
Add deleted_on to ServiceTemplate for archival

agrare:
This adds a deleted_on datetime column to service_templates to allow for the use of the ArchivedMixin.
Also included here is code to migrate this out of the reserved table since this is intended to be backported.

https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/314

Core change: https://github.com/ManageIQ/manageiq/pull/17480
Backport: https://github.com/ManageIQ/manageiq/pull/17481

### DIFF
--- a/db/migrate/20180525171150_add_deleted_on_to_service_template.rb
+++ b/db/migrate/20180525171150_add_deleted_on_to_service_template.rb
@@ -1,0 +1,29 @@
+class AddDeletedOnToServiceTemplate < ActiveRecord::Migration[5.0]
+  class ServiceTemplate < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    include ReservedMigrationMixin
+    include MigrationStubHelper
+  end
+
+  def up
+    add_column :service_templates, :deleted_on, :datetime
+
+    say_with_time("Migrate data from reserved table to service_templates") do
+      ServiceTemplate.includes(:reserved_rec).each do |st|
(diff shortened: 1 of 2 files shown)


agrare:
@miq-bot assign @Fryguy

db/migrate/20180525171150_add_deleted_on_to_service_template.rb
```diff
@@ -0,0 +1,29 @@
...
+        st.reserved_hash_migrate(:deleted_on)
+      end
+    end
+  end
+
+  def down
+    say_with_time("Migrate data from service_templates to reserved table") do
+      ServiceTemplate.includes(:reserved_rec).each do |st|
+        st.reserved_hash_set(:deleted_on, st.deleted_on)
+        st.save!
```
Fryguy: Is there no reversed migrate method?

agrare: Nope just https://github.com/ManageIQ/manageiq-schema/blob/master/lib/reserved_migration_mixin.rb#L6-L10

agrare: This is how all of our reserved table down migrations are today but I can write a reserved_hash_migrate down equivalent if you want.

`reserved_hash_unmigrate` or something?  It would just be `self.reserved_hash_set(k, v); self.save!`

PR #209, opened 2018-05-30
Add column report_base_model to ChargebackRate table

hstastna:
**Related to:** https://bugzilla.redhat.com/show_bug.cgi?id=1581817

**What:**
We need to store the type of the Rate which is related to type of Report: _Chargeback for VMs/Images/Projects_.

**Why:**
We need it for adding new drop down to editing screen for Rate to be able to choose the type of the rate and then to display only appropriate columns of the rate, based on the chosen type of the rate (VMs, Images, Projects). I will add appropriate screenshots later.

### DIFF
--- a/db/migrate/20180530160321_add_report_base_model_to_chargeback_rate.rb
+++ b/db/migrate/20180530160321_add_report_base_model_to_chargeback_rate.rb
@@ -1,0 +1,5 @@
+class AddReportBaseModelToChargebackRate < ActiveRecord::Migration[5.0]
+  def change
+    add_column :chargeback_rates, :report_base_model, :string
+  end
+end


hstastna:
@lpichler @gtanzillo Any update about this PR? Thanks! ;)

Fryguy:
I'm not sure about this one...something feels off about the design...can we discuss?

lpichler:
@Fryguy yes are right, Let me bring some facts:
current diagram:
screen shot 2018-08-09 at 10 01 53

These days as you know we have **3 types** of chargeback reports: for **VMs/Images/Projects.**
Each report is reporting different columns(in **DB ChargeableField**) but some of them are common.

This information is now done at application level[ by this hash.](https://github.com/lpichler/manageiq/blob/ce3bb5dc39eb168a70a9f6be73ed496cb5dc6c22/app/models/chargeback_vm.rb#L2) in **ChargebackVm, ChargebackContainerImage, ChargebackContainerProject.**

It looks that  we don't need to extra store this type in ChargebackRate table. We rather want to know which column is related to which types.

**ChargebackRate are always related to particular report type.** but if we **go deeper the chargeback field are making the relations to chargeback report type.**

### We can think about these three options:

**1) Store report type in ChargebackRate**
This will simplify  work in UI (https://github.com/ManageIQ/manageiq-ui-classic/pull/4001).
It could be also used in backend when chargeback rate are picked up.

**2) Store report type in ChargeableField**
Because basically it is about chargeable fields and their relations to report type.
But one chargeback field can be related to more chargeback type. 
So I guess this option will require another table to represent chargeback report type.
This option looks like cleanest DB solution.

**3) Leave as it is and use application code**
This will bring lot of code and complex methods.


@Fryguy which option are you prefer ? I am starting to prefer option 2.

thanks

carbonin:
+1 for option 2.

Based on what I've read, the issue you're trying to solve is that different models have different fields associated with them and we should only display the fields for the selected model when building a report, right? (If this is wrong then just let me know and disregard my opinion here :laughing: )

A lookup table for model names could be a good solution for this, but I wonder if there is any other place that we store what metrics apply to what model? Do we have the concept of "metric types" that we can link here instead of using a metric string? That would probably be a more useful solution generally, but definitely not suggesting we build that for this PR if it doesn't exist currently.

hstastna:
@lpichler @Fryguy @carbonin @gtanzillo Any update about this? Does it mena that we have chosen option 2? Thanks! :)

hstastna:
@lpichler @Fryguy @carbonin @gtanzillo Any update about this? Does it mean that we have chosen option 2? Thanks! :)

carbonin:
@hstastna I think we're waiting on @Fryguy 's input here.

Fryguy:
I am generally not a fan of storing model names in the database unless they are part of a polymorphic association (and thus Rails takes care of it), and that's part of the reason for my initial concern.  That being said, in this particular case, there are only 3 specific models and they are all subclasses of Chargeback, so the scope of what can be stored here is limited and can probably even be mitigated with a model validation to ensure only those types are stored there.  Additionally, those 3 models are not actual models, but ActsAsArModel types, so they couldn't be stored via polymorphic relations anyway.  Given all that I'm good with option 2.

carbonin:
@Fryguy Just to be clear, I don't think this PR actually implements Libor's option 2.

lpichler:
@Fryguy yes, I agree with you, but as @carbonin said this is not option 2, this is option 1, should I do revert PR or create migration  which is removing the column  `report_base_model` back ?

hstastna:
Maybe I had better to move this PR to WIP. Because originally, option 1 was implemented and then we started discussion about another options and I was waiting for others to choose the best option, before making any change in this PR. But now, maybe option 1 is not the worst :D

Fryguy:
Gah I totally misread the PR.  I saw it adding the column to chargeable_rates, but when I read option 2 I saw it say chargeable_fields and was like "yup, that's what the PR is doing...merge".

@lpichler You'll have to create a new PR that deletes the one column and adds another...sorry about that.

hstastna:
Just a note that this PR was 'reverted'.

PR #210, opened 2018-06-01
Inventory Tower Workflow and it's nodes

jameswnl:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1468339

refer to https://github.com/ManageIQ/manageiq/issues/17441


db/migrate/20180601171211_create_workflow_node.rb
```diff
@@ -0,0 +1,13 @@
+class CreateWorkflowNode < ActiveRecord::Migration[5.0]
+  def change
+    create_table :workflow_nodes do |t|
```
blomquisg: `workflow_nodes` is probably too generic for this.  I think the name needs to be scoped to the configuration models.  Maybe even just something like `configuration_workflow_nodes`.

@Fryguy

jameswnl: done prefixing the table with `configuration_`

jameswnl:
Endup, we don't need to collect workflow_nodes ATM.  Closing this.

PR #212, opened 2018-06-05
Add a polymorphic resource to MiqSchedules

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/17581

To be merged with: https://github.com/ManageIQ/manageiq/pull/17545


bdunne:
Added WIP because I just got news that the feature I'm working on may need to be backported

PR #213, opened 2018-06-06
Convert user quadicon settings from the old format to the new one

skateman:
The [new quadicon settings logic](https://github.com/ManageIQ/manageiq-ui-classic/pull/4034) requires to move the value from `settings[:quadicons][:ems]` into settings[:quadicons][:ems_infra] for each user.

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/4027

### DIFF
--- a/spec/migrations/20180606083431_convert_quadicon_settings_keys_spec.rb
+++ b/spec/migrations/20180606083431_convert_quadicon_settings_keys_spec.rb
@@ -1,0 +1,31 @@
+require_migration
+
+describe ConvertQuadiconSettingsKeys do
+  let(:user_stub) { migration_stub :User }
+
+  migration_context :up do
+    it 'moves the value of :ems to :ems_infra' do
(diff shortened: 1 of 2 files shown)


spec/migrations/20180606083431_convert_quadicon_settings_keys_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  migration_context :down do
+    it 'moves the value of :ems to :ems_infra' do
+      user = user_stub.create!(:settings => {:quadicons => {:ems_infra => false }})
+
+      migrate
+
+      user.reload
+
+      expect(user.settings[:quadicons][:ems_infra]).to be_nil
+      expect(user.settings[:quadicons][:ems]).to eq(false)
```
lpichler: do we have matcher for it ? like be_falsey or something like that ?

skateman:
@miq-bot assign @Fryguy

PR #215, opened 2018-06-07
Migrate Nuage CloudSubnet default type to new subclass

miha-plesko:
Nuage inventory was recently updated to fetch two types of CloudSubnets:

```
- ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L3
- ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L2
```

Before only CloudSubnet::L3 were inventoried using default type CloudSubnet. With this migration we make sure that this default type is converted to CloudSubnet::L3 type because persister won't modify STI type per se.

Related PR where we updated parser: https://github.com/ManageIQ/manageiq-providers-nuage/pull/94
@Ladas thanks for pointing out that STI type won't be updated by refresher!

@miq-bot add_label data
@miq-bot assign @Ladas

### DIFF
--- a/db/migrate/20180607084710_nuage_subclass_l3_cloud_subnet.rb
+++ b/db/migrate/20180607084710_nuage_subclass_l3_cloud_subnet.rb
@@ -1,0 +1,24 @@
+class NuageSubclassL3CloudSubnet < ActiveRecord::Migration[5.0]
+  class CloudSubnet < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time('Migrating Nuage default subnet type to L3 subnet subclass') do
+      CloudSubnet.where(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet')
+                 .update_all(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L3')
+    end
+  end
+
+  def down
+    say_with_time('Reverting Nuage L3 subnet subclass to default subnet type') do
+      CloudSubnet.where(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L3')
+                 .update_all(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet')
+    end
+
+    say_with_time('Reverting Nuage L2 subnet subclass to default subnet type') do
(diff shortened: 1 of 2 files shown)


miha-plesko:
@gtanzillo @agrare so this is the tiny schema PR that we were talking about in Berlin 2/2

agrare:
@miq-bot assign @Fryguy

miha-plesko:
@Ladas can you please take a look again here, I've applied two changes:

- converted to `::CloudSubnet::L3` instead `::CloudSubnetL3` as per https://github.com/ManageIQ/manageiq-providers-nuage/pull/105
- enhanced `down` migration so that it converts both `::CloudSubnet::L3` and `::CloudSubnet::L2` into `::CloudSubnet`. Before it was converting `L3` only.

Ladas:
@Fryguy can you review? https://github.com/ManageIQ/manageiq-providers-nuage/pull/105 was merged

db/migrate/20180607084710_nuage_subclass_l3_cloud_subnet.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def down
+    say_with_time('Reverting Nuage L3 subnet subclass to default subnet type') do
+      CloudSubnet.where(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L3')
+                 .update_all(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet')
+    end
+
+    say_with_time('Reverting Nuage L2 subnet subclass to default subnet type') do
+      CloudSubnet.where(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L2')
+        .update_all(:type => 'ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet')
```
Ladas: @miha-plesko  looks like rubocop wants this also aligned

miha-plesko: Miscusi, fixed. I knew all those "Avoid using update_all" are legit but didn't notice the Align one for some reason.

miha-plesko:
At some point, we're gonna need to merge this @Fryguy or else user will have to remove and re-add Nuage provider to get the right types in DB.

PR #217, opened 2018-06-07
Allow FloatingIp to relate directly to NetworkRouter

miha-plesko:
With this commit we add foreign key `network_router_id` on FloatingIp model in order to be able to connect FloatingIp with NetworkRouter. Currently such connection is only possible through CloudNetworks which does not fit well in Nuage inventoring.

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574922

@miq-bot assign @Fryguy 
@miq-bot add_label schema

/cc @Ladas

### DIFF
--- a/db/migrate/20180607134817_add_network_router_id_to_floating_ip.rb
+++ b/db/migrate/20180607134817_add_network_router_id_to_floating_ip.rb
@@ -1,0 +1,5 @@
+class AddNetworkRouterIdToFloatingIp < ActiveRecord::Migration[5.0]
+  def change
+    add_reference(:floating_ips, :network_router, :type => :bigint, :index => true)
+  end
+end


miha-plesko:
PR that needs this new field: https://github.com/ManageIQ/manageiq-providers-nuage/pull/97

miha-plesko:
@agrare I see you're usually pinged for first review on schema change PRs. So here it is, the ping, if you please.

db/migrate/20180607134817_add_network_router_id_to_floating_ip.rb
```diff
@@ -0,0 +1,5 @@
+class AddNetworkRouterIdToFloatingIp < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :floating_ips, :network_router, :type => :bigint, index: true, foreign_key: true
```
agrare: We don't use any foreign key constraints so can you remove that?  Otherwise looks good to me.

miha-plesko:
Thanks @agrare , I've removed the foreign key constraint.

@Ladas on gitter you suggested that we connect FloatingIp to NetworkRouter via NetworkPort which would work in case when FloatingIp is assigned to a NetworkPort. However, this is not always the case: FloatingIp can be in "unassigned" state meaning it's not related to any NetworkPort yet. With this commit we'd like to make sure that relation between NetworkRouter and FloatingIp exists even in such case.

Ladas:
@miha-plesko do you have a drawing somewhere of what the router connections are? The current schema comes mainly from OpenStack, where you route CloudSubnets to CloudNetwork (as gateway), then I think the floating_ips is a list of available floating ips you can assign to VMs connected to the CloudSubnets

miha-plesko:
@Ladas so our current inventoring plan is as follows:

capture

On Nuage GUI it looks like this (example of L3 network):
capture

From left to right:

- L3 Domain aka `NetworkRouter` (yellow)
- Zone (not planned to be inventoried in MIQ for now) (red)
- L3 Subnet aka `CloudSubnetL3` (blue)
- vPort aka `NetworkPort`
- interface (not planned to be inventoried in MIQ for now)

And example of L2 network:

capture

This time, left-most element is L2 Subnet:

- L2 Subnet aka `CloudSubnetL2` (green)
- vPort aka `NetworkPort`
- interface (not planned to be inventoried in MIQ for now)

My thinking is that we should probably inventory "Zone" as `CloudNetwork`, but I'm not sure if it would be the right thing to do. So for now I'd just allow for FloatingIp to be related to NetworkRouter.

Looking forward to your comment!

miha-plesko:
@Ladas looking forward to your opinion ^

Ladas:
@miha-plesko sorry, I've been traveling on Friday

ok, so for OpenStack having ```network_router :belongs_to cloud_network```, the CloudNetwork is the network we can get the FloatingIps from (having CIDR). 

So the question is, how is this modeled in Nuage. It looks like the ```L3 Domain aka NetworkRouter``` holds the CIDR? Could we maybe split the model, that the network part of the router would be in CloudNetwork? And the FloatingIps would be associated to that?

miha-plesko:
Closing after discussion with @Ladas. We'll rather inventory also CloudNetworks and connect FloatingIps to them, then things will appear as they should.

miha-plesko:
Reopening the PR. In fact it turns out we do need to explicitly inventory connection between FloatingIp and Network router to match how Nuage models relations. See [here](https://github.com/ManageIQ/manageiq-providers-nuage/pull/97#issuecomment-396620719). @Ladas I kindly ask for merge.

miha-plesko:
@Fryguy can we merge this?

miha-plesko:
@gtanzillo @agrare so this is the tiny schema PR that we were talking about in Berlin 1/2

Loicavenel:
@Fryguy HELP please

PR #218, opened 2018-06-08
Move ansible secrets from OpenShift to the database

carbonin:
In previous versions, the secret values for the embedded ansible pod
lived in an OpenShift secret object named "ansible-secrets".In newer
versions these values are managed using the Authentication model
against MiqDatabase.

This migration retrieves the values from the secret and saves them
as "Authentication" objects if we see that we are running in OpenShift
and the secret object is present.

https://www.pivotaltracker.com/story/show/157287338

@miq-bot assign @Fryguy 

/cc @bdunne


carbonin:
As requested in https://github.com/ManageIQ/manageiq-pods/pull/295/files#r193231863

carbonin:
Adding WIP until I can build this into an image and test the container upgrade end-to-end

carbonin:
@miq-bot add_label wip

PR #219, opened 2018-06-08
Add VMware folder types.

lfu:
Part of https://github.com/ManageIQ/manageiq/pull/17519.

https://bugzilla.redhat.com/show_bug.cgi?id=1090957

@miq-bot add_label gaprindashvili/no, enhancement

### DIFF
--- a/db/migrate/20180608174354_update_vmware_folder_types.rb
+++ b/db/migrate/20180608174354_update_vmware_folder_types.rb
@@ -1,0 +1,35 @@
+class UpdateVmwareFolderTypes < ActiveRecord::Migration[5.0]
+  VMWARE_FOLDER = "ManageIQ::Providers::Vmware::InfraManager::Folder".freeze
+  VMWARE_DC = "ManageIQ::Providers::Vmware::InfraManager::Datacenter".freeze
+
+  class EmsFolder < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class Datacenter < EmsFolder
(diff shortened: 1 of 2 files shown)


Fryguy:
@agrare Can you verify this?  Do other providers use folder STI?  I would think we'd have to fix them all at once, but then I'm confused about which ones do and don't use STI.

agrare:
@Fryguy The other STI used for folders are `Datacenter < EmsFolder` and `StorageCluster < EmsFolder`.  (And `ManageIQ::Providers::AutomationManager::InventoryGroup < EmsFolder` which is a new one to me).

Ovirt uses the base `EmsFolder` type: https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/refresh/parse/parser.rb#L709-L710
Same with SCVMM: https://github.com/ManageIQ/manageiq-providers-scvmm/blob/master/app/models/manageiq/providers/microsoft/infra_manager/refresh_parser.rb#L593

I'll test this out with all three providers added.

agrare:
@miq-bot assign @Fryguy

gmcculloug:
@agrare Is this data migration really needed or will it happen naturally as part of a provider refresh?  The feature that is using this data will be part of the next release so I would think a full provider refresh would happen as part of the upgrade to the next version.

agrare:
@gmcculloug we don't update the type as part of EmsRefresh so existing folders will not have their types updated.

lfu:
Add new method to ```VM``` instead of ```EmsFolder```. Keep ```EmsFolder``` untouched.

PR #221, opened 2018-06-13
Add Firmware Type to Hardwares

agrare:
Add a column to store if a VM/Host is using BIOS or EFI as their
firmware type.


agrare:
@miq-bot assign @Fryguy

PR #222, opened 2018-06-18
Migrations for suspending provider

slemrmartin:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1455145

**Issue**: https://github.com/ManageIQ/manageiq/issues/17489

- [ ] (**dependent**) https://github.com/ManageIQ/manageiq/pull/17452

Adds Zone visibility and EMS's backup zone relation

### DIFF
--- a/db/migrate/20180618083035_add_visible_to_zone.rb
+++ b/db/migrate/20180618083035_add_visible_to_zone.rb
@@ -3,8 +3,8 @@ class AddVisibleToZone < ActiveRecord::Migration[5.0]
     add_column :zones, :visible, :boolean, :default => true
     reversible do |dir|
       dir.up do
-        Zone.reset_column_information
-        Zone.seed
+        ::Zone.reset_column_information
+        ::Zone.seed
       end
     end
   end
--- a/db/migrate/20180618083035_add_visible_to_zone.rb
+++ b/db/migrate/20180618083035_add_visible_to_zone.rb
@@ -1,0 +1,11 @@
+class AddVisibleToZone < ActiveRecord::Migration[5.0]
+  def change
+    add_column :zones, :visible, :boolean, :default => true
+    reversible do |dir|
+      dir.up do
+        Zone.reset_column_information
+        Zone.seed
+      end
+    end
+  end
+end
--- a/db/migrate/20180618083035_add_visible_to_zone.rb
+++ b/db/migrate/20180618083035_add_visible_to_zone.rb
@@ -1,5 +1,5 @@
 class AddVisibleToZone < ActiveRecord::Migration[5.0]
   def change
-    add_column :zones, :visible, :boolean, :default => true
+    add_column :zones, :visible, :boolean
   end
 end
--- a/db/migrate/20180618083035_add_visible_to_zone.rb
+++ b/db/migrate/20180618083035_add_visible_to_zone.rb
@@ -1,11 +1,5 @@
 class AddVisibleToZone < ActiveRecord::Migration[5.0]
   def change
     add_column :zones, :visible, :boolean, :default => true
-    reversible do |dir|
-      dir.up do
-        ::Zone.reset_column_information
-        ::Zone.seed
-      end
-    end
   end
 end
--- a/db/migrate/20180618084054_init_zones_visibility.rb
+++ b/db/migrate/20180618084054_init_zones_visibility.rb
@@ -1,0 +1,16 @@
+class InitZonesVisibility < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Updating all zones to visible") do
+      Zone.update_all(:visible => true)
+    end
+  end
+
+  def down
+    say_with_time("Resetting zone visibility") do
+      Zone.update_all(:visible => nil)
+    end
+  end
+end
--- a/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
+++ b/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
@@ -1,0 +1,5 @@
+class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
+  def change
+    add_column :ext_management_systems, :backup_zone_id, :integer, :after => :zone_id, :index => true
+  end
+end
--- a/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
+++ b/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
@@ -1,5 +1,5 @@
 class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
   def change
-    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true, :after => :zone_id
+    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true
   end
 end
--- a/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
+++ b/db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
@@ -1,5 +1,5 @@
 class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
   def change
-    add_column :ext_management_systems, :backup_zone_id, :integer, :after => :zone_id, :index => true
+    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true, :after => :zone_id
   end
 end
--- a/db/migrate/20180626125654_make_maintenance_zone_record.rb
+++ b/db/migrate/20180626125654_make_maintenance_zone_record.rb
@@ -1,16 +1,32 @@
 class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
   class Zone < ActiveRecord::Base
+    validates_presence_of   :name
+    validates_uniqueness_of :name
+
+    MAINTENANCE_ZONE_NAME = "__maintenance__".freeze
   end
 
   def up
     say_with_time("Creating Maintenance Zone") do
-      Zone.create_with(:description => "Maintenance Zone", :visible => false).find_or_create_by!(:name => 'maintenance')
+      zone = Zone.where(:name => Zone::MAINTENANCE_ZONE_NAME).first
+      if zone.present?
+        zone.name = "#{zone.name}_0"
+        zone.save
+      end
+
+      Zone.create_with(:description => "Maintenance Zone", :visible => false).find_or_create_by!(:name => Zone::MAINTENANCE_ZONE_NAME)
     end
   end
 
   def down
     say_with_time("Deleting Maintenance Zone") do
-      Zone.where(:name => 'maintenance').where(:visible => false).destroy_all
+      Zone.where(:name => Zone::MAINTENANCE_ZONE_NAME).where(:visible => false).destroy_all
+
+      orig = Zone.where(:name => "#{Zone::MAINTENANCE_ZONE_NAME}_0").first
+      if orig.present?
+        orig.name = orig.name[0..orig.name.size - 3]
+        orig.save
+      end
     end
   end
 end
--- a/db/migrate/20180626125654_make_maintenance_zone_record.rb
+++ b/db/migrate/20180626125654_make_maintenance_zone_record.rb
@@ -1,5 +1,7 @@
 class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
   class Zone < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
     validates_presence_of   :name
 
     MAINTENANCE_ZONE_NAME = "__maintenance__".freeze
@@ -7,7 +9,7 @@ class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
 
   def up
     say_with_time("Creating Maintenance Zone") do
-      zone = Zone.where(:name => Zone::MAINTENANCE_ZONE_NAME).first
+      zone = Zone.in_my_region.where(:name => Zone::MAINTENANCE_ZONE_NAME).first
       if zone.present?
         zone.name = "#{zone.name}_0"
         zone.save
@@ -19,9 +21,9 @@ class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
 
   def down
     say_with_time("Deleting Maintenance Zone") do
-      Zone.where(:name => Zone::MAINTENANCE_ZONE_NAME).where(:visible => false).destroy_all
+      Zone.in_my_region.where(:name => Zone::MAINTENANCE_ZONE_NAME).where(:visible => false).destroy_all
 
-      orig = Zone.where(:name => "#{Zone::MAINTENANCE_ZONE_NAME}_0").first
+      orig = Zone.in_my_region.where(:name => "#{Zone::MAINTENANCE_ZONE_NAME}_0").first
       if orig.present?
         orig.name = orig.name[0..orig.name.size - 3]
         orig.save
--- a/db/migrate/20180626125654_make_maintenance_zone_record.rb
+++ b/db/migrate/20180626125654_make_maintenance_zone_record.rb
@@ -1,0 +1,16 @@
+class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Creating Maintenance Zone") do
+      Zone.create_with(:description => "Maintenance Zone", :visible => false).find_or_create_by!(:name => 'maintenance')
+    end
+  end
+
+  def down
+    say_with_time("Deleting Maintenance Zone") do
+      Zone.where(:name => 'maintenance').where(:visible => false).destroy_all
+    end
+  end
+end
--- a/db/migrate/20180626125654_make_maintenance_zone_record.rb
+++ b/db/migrate/20180626125654_make_maintenance_zone_record.rb
@@ -1,7 +1,6 @@
 class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
   class Zone < ActiveRecord::Base
     validates_presence_of   :name
-    validates_uniqueness_of :name
 
     MAINTENANCE_ZONE_NAME = "__maintenance__".freeze
   end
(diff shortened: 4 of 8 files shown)


slemrmartin:
Cc @Fryguy

db/migrate/20180618083035_add_visible_to_zone.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisibleToZone < ActiveRecord::Migration[5.0]
+  def change
+    add_column :zones, :visible, :boolean, :default => true
```
Fryguy: Prefer `default_value_for`  at the model level over a database-level default.

Fryguy: That being said, you will need a data migration to update all existing zones to visible => true.

db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
```diff
@@ -0,0 +1,5 @@
+class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true, :after => :zone_id
```
Fryguy: Remove the `:after` bit...it's not necessary.

db/migrate/20180626125654_make_maintenance_zone_record.rb
```diff
@@ -0,0 +1,16 @@
...
+      Zone.create_with(:description => "Maintenance Zone", :visible => false).find_or_create_by!(:name => 'maintenance')
+    end
+  end
+
+  def down
+    say_with_time("Deleting Maintenance Zone") do
+      Zone.where(:name => 'maintenance').where(:visible => false).destroy_all
+    end
+  end
+end
```
Fryguy: This entire migration is unnecessary, because this can be handled in Zone.seed.

db/migrate/20180618083035_add_visible_to_zone.rb
slemrmartin: @Fryguy exactly, so why is preferred more difficult solution? database-level default, at least for boolean, was standard for every project I seen before.

db/migrate/20180618084757_add_backup_zone_id_to_ext_management_system.rb
slemrmartin: it makes db structure more clear, if you group logical related columns together

db/migrate/20180626125654_make_maintenance_zone_record.rb
slemrmartin: So do I have to add Zone.seed to every method which uses maintenance zone? 
I'm not sure how find_or_create in every method call is better than one data migration

slemrmartin: also this call should be ensured in Zone.maintenance_zone but I follower implementation of Zone.default_zone, where it is not. 
Rely on all developers, that they will call Zone.seed every time when they want to use one of these methods doesn't seem to be good idea

db/migrate/20180618083035_add_visible_to_zone.rb
slemrmartin: I thought about it, what about compromise solution? keep `:default => true` as better alternative to data migration and add `default_value_for` to keep your code standards

slemrmartin: ...ok, I disagree with your approach, but we need to finish this RFE, so as you wish

db/migrate/20180626125654_make_maintenance_zone_record.rb
slemrmartin: @fryguy - are you sure that `Zone.seed` is called after migrations in all existing installations?

db/migrate/20180626125654_make_maintenance_zone_record.rb
```diff
@@ -0,0 +1,32 @@
+class MakeMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+    validates_presence_of   :name
+    validates_uniqueness_of :name
```
Fryguy: These must be removed.  In a global region, since zones are transferred from other regions, there are multiple "default" zones, for example.  As such, this migration will fail to migrate a global database.

db/migrate/20180626125654_make_maintenance_zone_record.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    MAINTENANCE_ZONE_NAME = "__maintenance__".freeze
+  end
+
+  def up
+    say_with_time("Creating Maintenance Zone") do
+      zone = Zone.where(:name => Zone::MAINTENANCE_ZONE_NAME).first
```
Fryguy: It just occurred to me that this is not region aware.  Please change this to `Zone.in_my_region.where(...)`, to allow migrating a global database.  You can also create a spec showing this, but will need to give each zone an id within a region.  There should be other migrations that have examples showing how to do that.

slemrmartin:
@Fryguy I'm not sure, how exactly migrations works on global database so please look if `in_my_region` occurences are enough

spec/migrations/20180626125654_make_maintenance_zone_record_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:zone_stub) { migration_stub(:Zone) }
+
+  let(:remote_region_start) do
+    anonymous_class_with_id_regions.rails_sequence_start +
+      anonymous_class_with_id_regions.rails_sequence_factor
+  end
+
+  migration_context :up do
+    before(:each) do
+      zone_stub.delete_all
```
Fryguy: Not sure why you need this...the database *should* be empty.

slemrmartin: it is not cleaned between specs in `migration_context` when I use `zone_stub.create!`. I don't know if it is some bad local settings, but specs are failing without this line.

spec/migrations/20180626125655_rename_user_maintenance_zone_record_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  let(:zone_stub) { migration_stub(:Zone) }
+
+  let(:remote_region_start) do
+    anonymous_class_with_id_regions.rails_sequence_start +
+      anonymous_class_with_id_regions.rails_sequence_factor
+  end
+
+  migration_context :up do
+    before(:each) do
+      zone_stub.delete_all
```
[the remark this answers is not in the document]
Ladas: hm, this should not happen, something must be off, because it should all just a transaction that gets rolled back

Delete like this will not work if the specs will run in parallel. So we need to find a way to to properly fix this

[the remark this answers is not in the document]
slemrmartin: @Ladas probably some bad `before :all` in the past in my local db so it work without it normally.

blomquisg:
@miq-bot assign @agrare

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  def down
+    say_with_time("Renaming user-defined Maintenance Zone") do
+      Zone.in_my_region.where(:name => Zone::MAINTENANCE_ZONE_NAME).where(:visible => false).destroy_all
+
+      orig = Zone.in_my_region.where(:name => "#{Zone::MAINTENANCE_ZONE_NAME}_0").first
+      if orig.present?
+        orig.name = orig.name[0..orig.name.size - 3]
```
agrare: Since you know the name is always going to be `"#{Zone::MAINTENANCE_ZONE_NAME}_0"` if you're here you don't need the dynamic string building, just call it `Zone::MAINTENANCE_ZONE_NAME`.

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,31 @@
...
+    include ActiveRecord::IdRegions
+
+    validates_presence_of   :name
+
+    MAINTENANCE_ZONE_NAME = "__maintenance__".freeze
+  end
+
+  def up
+    say_with_time("Renaming user-defined Maintenance Zone") do
+      zone = Zone.in_my_region.where(:name => Zone::MAINTENANCE_ZONE_NAME).first
```
agrare: If the name is unique can't you just do `.find_by()` instead of `.where().first`?

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,31 @@
...
+      if zone.present?
+        zone.name = "#{zone.name}_0"
```
agrare: Getting into some real edge cases now, but if they had a `__maintenance___0` zone already this will fail.

agrare:
@slemrmartin most of the complexity seems to be around the name of the maintenance zone which must be known so that e.g. [this](https://github.com/ManageIQ/manageiq/pull/17452/files#diff-1e1418a05e4c9116c15a7b073e0b00bbR91) can work.

What if instead of a `:visible` column we just go straight for a `:maintenance` boolean column, that way we can create the maintenance zone with whatever name we want if it conflicts, even a UUID or something.  Then the check is just `zone.maintenance?` instead of `zone.name == "__maintenance__"`?

Otherwise we're stuck with all this name conflict complexity.

agrare:
Ohh now I remember why we had to do that, MiqQueue.put only has the zone name not the zone object [ref](https://github.com/ManageIQ/manageiq/pull/17452/files#diff-d8c450b508868941c2b0c5884e718324R129)

slemrmartin:
@agrare `maintenance` boolean column seems confusing to me, it tells that every zone can be in maintenance mode, but it can't. 
With `__maintenance__` unique name I followed `default` zone, which is also identified by unique name (so there is common approach for both unique records, 2 - the reason you mentioned )

On the other side `visible` mode is more generic and can be used by another RFE for different purposes. UI and API hide zones which are not visible, this can be used in the future without new special columns

agrare:
Yeah I think `__maintenance__` is unlikely enough to have a collision, `__maintenance___0` is even less likely.

@carbonin @bdunne what do you guys think?  If we're really paranoid we could do what we used to do for [duplicate hostnames](https://github.com/ManageIQ/manageiq/blob/53b733b7a254f5677cd26ee7ccd5a7c3c879a63a/app/models/ems_refresh/save_inventory_infra.rb#L176-L184)

carbonin:
@agrare Maybe it would be best to just fail the migration if they have a zone called `__maintenance__` in their environment already. I feel like trying to deal with this special zone name by renaming something that the user set up may cause more problems than it would solve.

carbonin:
Plus you could then include remediation steps directly in the migration error message because we would know exactly what the problem was.

agrare:
@carbonin :+1: works for me

slemrmartin:
@carbonin I'm not sure what you mean by it - what does it mean "just fail the migration"? 
Migration only renames possible existing zone, it doesn't create our `__maintenance__` zone (because this code is in `Zone.seed`) - so this conflict can happen in runtime.

And creating of zone in migration doesn't want @Fryguy, so it seems to me that we are in loop :)

agrare:
@slemrmartin Just leave the rename as is but add an exception handler that prints a message like "Failed to rename __maintenance__, check for zone conflicts" and re-raise the exception.  It will fail the upgrade and the customer will have to remediate it.

db/migrate/20180618084054_init_zones_visibility.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def up
+    say_with_time("Updating all zones to visible") do
+      Zone.update_all(:visible => true)
+    end
+  end
+
+  def down
+    say_with_time("Resetting zone visibility") do
```
bdunne: For things like this I think it should be merged with the `add_column` migration.  Then the `up` adds the column and updates the existing records and the `down` just drops the column.  Updating the column on every record before dropping the column seems pointless to me.

@carbonin @agrare Thoughts?

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,31 @@
+class RenameUserMaintenanceZoneRecord < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    validates_presence_of   :name
```
bdunne: Is this necessary?

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
bdunne: Should we instead generate a unique zone name for the region and store it somewhere in the database (maybe MiqRegion) so that we don't run into naming conflicts?

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,31 @@
...
+      if zone.present?
+        zone.name = "#{zone.name}_0"
+        zone.save
+      end
+    end
+  end
+
+  def down
+    say_with_time("Renaming user-defined Maintenance Zone") do
+      Zone.in_my_region.where(:name => Zone::MAINTENANCE_ZONE_NAME).where(:visible => false).destroy_all
```
bdunne: The `where`s can be merged

bdunne:
> "just fail the migration"

Doesn't seem acceptable to me.  If it was only developers I could be okay with it, but not for customers.  I know it's an edge case, but they're not going to know what went wrong or how to fix it and it's going to lead to an unnecessary support call on a weekend (since that's when they do upgrades).  Maybe we could talk about something like https://github.com/ManageIQ/manageiq-schema/pull/222#discussion_r213321488 instead?

carbonin:
@bdunne they should know exactly what to do to fix it if we include the remediation steps in the failure message. It should be straight forward enough.

bdunne:
Okay, so @carbonin and I were just discussing this some more...

I think renaming an existing zone in a migration is just out of the question.  Too many things don't use real relations and point to zones by name (mistakes were made), that's why there is a uniqueness validation on Zone#name.

That said, it really doesn't matter what the name of our maintenance zone is as long as we can find it when we want to use it.  I think we should generate a random zone name and create that as our maintenance zone (and if it conflicts with an existing one, generate a new one until it doesn't conflict).  Then we can put a reference to that zone on MiqRegion so that we can find it in the future.

Is there any case where we would want multiple maintenance zones?

slemrmartin:
@bdunne good point
maintenance zone is identified by unique name (just like default zone, I wanted to be consistent) and created outside migration in `Zone.seed`

Answer is: No, we don't want multiple maintenance zones - that's another reason why it isn't identified by some boolean flag. 

I consider your solution as more robust, but it requires new migration to MiqRegion - I hope for choosing any solution, it was customer request (urgent, I suppose) 3 months ago, I think ;))

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
Fryguy: If it is, it need to be multi-region aware.

carbonin:
@Fryguy For what it's worth, we don't show the region accordion for anything other than the current region. So we don't really display the zone anywhere in the UI. 

But while thinking about this, I'm realizing that the provider (and its relation to its zone) would be replicated. That makes me think that the global will need the information about which zone is is the maintenance zone for each of the remote regions to display the provider correctly (as suspended).

slemrmartin:
I would prefer to keep identification by name `__maintenance__`. We searched for all occurences in UI and API and hide this zone (and others) identified by `visible?` boolean.
But there could be always some random place, of course.

If we need to rename existing `__maintenance__`, I can search schema.rb for all "zone" columns which are string and add it to this migration. This can solve remaining problems, right?

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
slemrmartin: Ok, I'll use "#{Zone::MAINTENANCE_ZONE_NAME}_#{Zone.my_region_number}"
Problem is the same, but risk of fail is extremely low, I think

db/migrate/20180618084054_init_zones_visibility.rb
slemrmartin: @bdunne - there is a rule "Don't mix structural and data migration". So I followed it. 
If there are lots of reviewers they cannot want opposite things

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  class MiqQueue < ActiveRecord::Base
+    belongs_to :miq_task
+  end
+
+  def up
+    say_with_time("Renaming user-defined Maintenance Zone") do
+      zone = Zone.in_my_region.find_by(:name => Zone::MAINTENANCE_ZONE_NAME)
+      if zone.present?
+        # STEP 1 - rename zone
```
bdunne: I still think this is the wrong approach

slemrmartin: @bdunne you were worried about string associations
1) possibility that user has `__maintenance__` as name is almost zero
2) if so, zone and associations are renamed to name with possibility of duplication zero^2.

could you be more specific? When this fails (with significant probability). 
Is it not acceptable for @Fryguy, @agrare, @carbonin too?

slemrmartin:
I prepared alternative PR with region association:
https://github.com/ManageIQ/manageiq-schema/pull/275

db/migrate/20180626125655_rename_user_maintenance_zone_record.rb
bdunne: > 1. possibility that user has __maintenance__ as name is almost zero

I agree, but while unlikely, not impossible

> 2. if so, zone and associations are renamed to name with possibility of duplication zero^2.

Right now you're only updating `Job` and `MiqQueue`.  The following models have a `zone` column:
```
=> ["MiqQueue", "Job", "MiqTask", "ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job", "ManageIQ::Providers::EmbeddedAnsible::AutomationManager::PlaybookRunner", "ManageIQ::Providers::EmsRefreshWorkflow", "ManageIQ::Providers::AnsibleRunnerWorkflow", "VmScan", "ManageIQ::Providers::NativeOperationWorkflow", "Service::LinkingWorkflow"]
```
I also believe it's in Settings and who knows how many other places it's buried in a serialized column.  I just don't think it's practical to try renaming a zone

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-schema/pull/275 now that it is merged

PR #223, opened 2018-06-19
Create table for Cloud Volume Types

mansam:
http://bugzilla.redhat.com/show_bug.cgi?id=1592900
Cloud Volume Types are needed for mapping new Cinder volumes to specific backends.

### DIFF
--- a/db/migrate/20180618212608_create_cloud_volume_types.rb
+++ b/db/migrate/20180618212608_create_cloud_volume_types.rb
@@ -7,7 +7,7 @@ class CreateCloudVolumeTypes < ActiveRecord::Migration[5.0]
       t.string :backend_name
       t.string :ems_ref
       t.bigint :ems_id
-      t.boolean :is_public
+      t.boolean :public
 
       t.timestamps
     end
--- a/db/migrate/20180618212608_create_cloud_volume_types.rb
+++ b/db/migrate/20180618212608_create_cloud_volume_types.rb
@@ -1,0 +1,15 @@
+class CreateCloudVolumeTypes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :cloud_volume_types do |t|
+      t.text :description
+      t.string :name
+      t.string :type
+      t.string :backend_name
+      t.string :ems_ref
+      t.bigint :ems_id
+      t.boolean :is_public
+
+      t.timestamps
+    end
+  end
+end


mansam:
@agrare, could you please direct this to someone who can review it? Thanks!

lpichler:
@mansam just question, is it somehow related to `CloudVolume#volume_type` (CloudVolume#volume_types - method) ? (I am using it in chargeback)

agrare:
@mansam could you post some examples of what the different backings look like?  Should we associate this to the underlying cinder/etc... storages?

mansam:
@lpichler At least for Cinder volumes, the volume type field that is collected during inventory refresh would correspond to the name of one of these user defined types. I didn't touch that field though, since the association is a string rather than an ID, and it seems dangerous to try and retrofit in an association that isn't necessary for the feature I need to support.

@agrare Volume Type backend_names from Cinder are basically just a string identifying a volume backend that's specified in the user's Cinder configuration, something like "LVM_iSCSI". The name of the type object itself ("mytype" or etc) can then be provided to the Cinder API when creating a volume to instruct it to be built with a particular backend.

It's probably possible to associate these objects to Disks/Storages, but I'm not 100% on the details of those models and their relationships, or what other providers might be doing with the concept. I didn't want to add any extra moving parts than I had to.

Example of volume type output from Cinder:
```
    "volume_types": [
        {
            "name": "SSD",
            "qos_specs_id": null,
            "extra_specs": {
                "volume_backend_name": "lvmdriver-1"
            },
            "os-volume-type-access:is_public": true,
            "is_public": true,
            "id": "6685584b-1eac-4da6-b5c3-555430cf68ff",
            "description": null
        },
        {
            "name": "SATA",
            "qos_specs_id": null,
            "extra_specs": {
                "volume_backend_name": "lvmdriver-1"
            },
            "os-volume-type-access:is_public": true,
            "is_public": true,
            "id": "8eb69a46-df97-4e41-9586-9a40a7533803",
            "description": null
        }
    ]
}
```

mansam:
@agrare @lpichler any further thoughts?

agrare:
@mansam looks good to me
@miq-bot assign @Fryguy

aufi:
:+1:

mansam:
Hi @Fryguy, have you had a chance to look over this?

db/migrate/20180618212608_create_cloud_volume_types.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCloudVolumeTypes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :cloud_volume_types do |t|
+      t.text :description
+      t.string :name
+      t.string :type
+      t.string :backend_name
+      t.string :ems_ref
+      t.bigint :ems_id
+      t.boolean :is_public
```
Fryguy: prefer `public` over `is_public`

db/migrate/20180618212608_create_cloud_volume_types.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCloudVolumeTypes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :cloud_volume_types do |t|
+      t.text :description
+      t.string :name
+      t.string :type
```
Fryguy: Is this a literal STI type or some cloud-specific string?

mansam: Intended to be STI type

db/migrate/20180618212608_create_cloud_volume_types.rb
mansam: :+1: changing

Fryguy:
Once the is_public thing is fixed, this LGTM

mansam:
@Fryguy That field has been renamed

mansam:
@Fryguy can this be merged?

PR #226, opened 2018-06-25
Add ems_ref to lans to help with unique listing

agrare:
Add an ems_ref column to the lans table so that it can be used to assist
in getting a unique list of lans across hosts in cases where the uid_ems
is not unique i.e. VMware where the uid_ems is the name of the lan.

By adding the ems_ref which is unique across the ems we are able to get
a truly unique list of networks when showing a list across hosts.

### DIFF
--- a/db/migrate/20180625120055_add_ems_ref_to_lans.rb
+++ b/db/migrate/20180625120055_add_ems_ref_to_lans.rb
@@ -1,0 +1,5 @@
+class AddEmsRefToLans < ActiveRecord::Migration[5.0]
+  def change
+    add_column :lans, :ems_ref, :string
+  end
+end


agrare:
@miq-bot assign @Fryguy 

agrare:
@Fryguy this will allow the v2v UI to move a lot of logic out of the UI and in to the backend for listing vmware source networks.

Fryguy:
@agrare If you use ems_ref, we just have to be careful that the next refresh doesn't create brand new objects (or if it does, if that's ok...i.e. we don't lose tags or anything).  Not sure this will need a data migration or not.

agrare:
@Fryguy for now we're still going to use the same uid_ems for finding lans so we won't create/delete any existing lans, this is just to help list unique networks.

Fryguy:
@agrare Then, my only concern is that we will have an ems_ref, and not use it for what it's for, which will confuse people.   ¯\\\_(ツ)_/¯

PR #227, opened 2018-06-26
Enable S3 DB Backups

jerryk55:
Add the AWS region to the file depot so S3 depots contain a region value.

For RFE referenced in https://bugzilla.redhat.com/show_bug.cgi?id=1513520

@roliveri please review.  Also requires https://github.com/ManageIQ/manageiq-gems-pending/pull/355
as well as a forthcoming ManageIQ core PR and a TBD manageiq-ui-classic PR.  

@Fryguy please review and merge.  Thanks.

### DIFF
--- a/db/migrate/20180620170052_add_region_to_file_depot.rb
+++ b/db/migrate/20180620170052_add_region_to_file_depot.rb
@@ -1,5 +1,0 @@
-class AddRegionToFileDepot < ActiveRecord::Migration[5.0]
-  def change
-    add_column :file_depots, :region, :string
-  end
-end
--- a/db/migrate/20180620170052_add_region_to_file_depot.rb
+++ b/db/migrate/20180620170052_add_region_to_file_depot.rb
@@ -1,0 +1,5 @@
+class AddRegionToFileDepot < ActiveRecord::Migration[5.0]
+  def change
+    add_column :file_depots, :region, :string
+  end
+end
--- a/db/migrate/20180620170052_add_aws_region_to_file_depot.rb
+++ b/db/migrate/20180620170052_add_aws_region_to_file_depot.rb
@@ -1,0 +1,5 @@
+class AddAwsRegionToFileDepot < ActiveRecord::Migration[5.0]
+  def change
+    add_column :file_depots, :aws_region, :string
+  end
+end


blomquisg:
@miq-bot assign @Fryguy

db/migrate/20180620170052_add_region_to_file_depot.rb
```diff
@@ -0,0 +1,5 @@
+class AddRegionToFileDepot < ActiveRecord::Migration[5.0]
+  def change
+    add_column :file_depots, :region, :string
```
jrafanie: A region column feels specific to amazon and not to depots.  It's also a string... like the mistake we made with using zone name's in the queue instead of id's referencing rows in the zone table.  If we intend to use depots in other clouds, it might make sense to consider using a polymorphic relationship to that provider's storage location table.

This might be out of scope of this PR but I see the zone mistake we made with the queue when I look at this.

jerryk55: Yes, it is really just for the S3 File Depot, but there is not a separate S3 File Depot table.  There will also be a depot for Swift.  Can we look at your suggestion at a later point so that we can get this merged for the upcoming Sprint?  There are 4 other PRs that are waiting like dominoes on this one to fall... (poor choice of words, I know)

jrafanie: Sure, unless we know how it will look for other systems, it's probably better to wait until we know how it will look.  That is, unless the migration would be painful.

jerryk55: No we don't know at this point.  Thanks for your help.

jerryk55:
@Fryguy are you talking about the *name* region?  Can it be called something like "object_region" or "depot_region"?

Fryguy:
I'm fine if this is AWS specific and just named aws_region it something like that

Fryguy:
Yes the name `region` would conflict with our general region concept (every model has region_number, in_region, etc methods).  For ext_management_systems we chose `provider_region`, but I'm not sure that would make sense here.

Fryguy:
Reference, if you're curious https://github.com/ManageIQ/activerecord-id_regions/blob/master/lib/active_record/id_regions.rb

jerryk55:
@Fryguy, renamed the column "aws_region" and re-pushed.  Let me know if this looks good.  Thanks!

jerryk55:
@Fryguy can this be merged now?  Thanks!

PR #231, opened 2018-07-06
Adding Location LED name to asset_details table

douglasgabriel:
It is required to send LED action to physical components, once that the name of the LED can vary to each component.

### DIFF
--- a/db/migrate/20180706115011_add_loc_led_name_asset_details.rb
+++ b/db/migrate/20180706115011_add_loc_led_name_asset_details.rb
@@ -1,0 +1,5 @@
+class AddLocLedNameAssetDetails < ActiveRecord::Migration[5.0]
+  def change
+    add_column :asset_details, :location_led_ems_ref, :string
+  end
+end


douglasgabriel:
@miq-bot assign @agrare

agrare:
@douglasgabriel I'd recommend calling this something other than `_name`, it sounds like this is more of a unique identifier sent to the API to target a specific LED which we would call an `_ems_ref`.

xClarity may use a name for this but other physical providers might use something else.

douglasgabriel:
Good point @agrare. Updated.

agrare:
@miq-bot assign @Fryguy

Fryguy:
In general I'm ok with this, but I'm curious why this is not a foreign key reference to the table with the location_led itself instead?

Fryguy:
Discussed with @douglasgabriel and the long term goal is to put location_led_status into asset_details, so that should keep them together and consistent, so I am good with this. :+1:

PR #232, opened 2018-07-12
Add missing ems_id to switch

Ladas:
Add missing ems_id to switch, so we can continue with proper
refresh and cascade delete.

Dependents:
- [ ] https://github.com/ManageIQ/manageiq/pull/18427
- [ ] https://github.com/ManageIQ/manageiq-providers-vmware/pull/364


Ladas:
@agrare here it is, do you think it looks right?

db/migrate/20180711120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def up
+    say_with_time("Add missing ems_id to switch") do
+      connection.execute <<-SQL
+        UPDATE switches
+          SET ems_id=hosts.ems_id
+          FROM host_switches
+          INNER JOIN hosts ON hosts.id = host_switches.host_id
+          WHERE host_switches.switch_id = switches.id AND hosts.ems_id IS NOT NULL
```
agrare: Should we explicitly skip switches with an ems_id already not-null here?  We're not touching physical switches implicitly because they don't have entries in the host_switches table but that feels a little weak

Ladas: I've added ```AND switches.type = 'ManageIQ::Providers::Vmware::InfraManager::DistributedVirtualSwitch'``` so that should be probably enough?

db/migrate/20180711120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,42 @@
...
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => "ems_id"
+
+    has_many :host_switches
```
Fryguy: Any relationship needs to fully qualify the stubs.. e.g.  `:class_name => "AddMissingEmsIdToSwitch::HostSwitch"`

db/migrate/20180711120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,42 @@
...
+    belongs_to :host
+    belongs_to :switch
+  end
+
+  class Switch < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => "ems_id"
+
+    has_many :hosts, :through => :host_switches
```
Fryguy: This can get really nasty in migrations because Rails doesn't properly fully-qualify the class for this relationship.

Ladas: this is used only in spec, the migration is using raw SQL

db/migrate/20180711120130_add_missing_ems_id_to_switch.rb
Ladas: this is used only in spec, the migration is using raw SQL

db/migrate/20180925120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,52 @@
+class AddMissingEmsIdToSwitch < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => "ems_id"
```
Fryguy: All relationships in stubs must use fully qualified class names

Ladas: ok, added

db/migrate/20180925120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,52 @@
...
+             AND switches.type = 'ManageIQ::Providers::Vmware::InfraManager::DistributedVirtualSwitch'
+      SQL
+
+      connection.execute <<-SQL
+        UPDATE switches
+          SET host_id=hosts.id
+          FROM host_switches
+          INNER JOIN hosts ON hosts.id = host_switches.host_id
+          WHERE host_switches.switch_id = switches.id
+             AND switches.type = 'ManageIQ::Providers::Vmware::InfraManager::DistributedVirtualSwitch'
```
agrare: This should be `AND switches.type = 'ManageIQ::Providers::Vmware::InfraManager::HostVirtualSwitch'

db/migrate/20180925120130_add_missing_ems_id_to_switch.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  class Switch < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => "ems_id", :class_name => "AddMissingEmsIdToSwitch::ExtManagementSystem"
+    belongs_to :host, :class_name => "AddMissingEmsIdToSwitch::Host"
+
+    has_many :hosts, :through => :host_switches, :class_name => "AddMissingEmsIdToSwitch::Host"
+    has_many :host_switches, :class_name => "AddMissingEmsIdToSwitch::HostSwitch"
+  end
```
agrare: If these aren't used in the migration can you move to the spec files?

agrare:
I pushed some commits to this so I'll unassign myself.

@carbonin @bdunne can you review?  Setting a host_id and ems_id on the switches table will let us do proper cascade deletes and get rid of this really hacky way we currently handle deletions (https://github.com/ManageIQ/manageiq/pull/18427/files#diff-50f9feed046a3410ac5531492059a6fbL308) in save_inventory.

I tried to get this done with AR in a single query but wasn't able to, maybe you can think of a better way of doing this without raw SQL.

PR #233, opened 2018-07-12
Create Physical Disks migration

EsdrasVP:
This PR is able to:
- Add PhysicalDisks migration into schema with reference to PhysicalStorage

### DIFF
--- a/db/migrate/20180726142030_create_physical_disk.rb
+++ b/db/migrate/20180726142030_create_physical_disk.rb
@@ -1,0 +1,18 @@
+class CreatePhysicalDisk < ActiveRecord::Migration[5.0]
+  def change
+    create_table :physical_disks do |t|
+      t.string :model
+      t.string :vendor
+      t.string :status
+      t.string :location
+      t.index  :location
+      t.string :serial_number
+      t.string :health_state
+      t.string :type
+      t.string :disk_size
+      t.bigint :physical_storage_id
+      t.index  :physical_storage_id
+      t.timestamps
+    end
+  end
+end


EsdrasVP:
@miq-bot add_reviewer @Fryguy

agrare:
@EsdrasVP what is a storage driver exactly in this context?  Can you give some examples?

EsdrasVP:
@agrare A storage driver is a disk driver inside a physical storage. It has its own health state (being something like: "Ok" or "Degraded"), size, model, serial number, etc. Initially I thought making Physical Storages have Storages as guest devices, since it is an existing component in ManageIQ, but I noticed they are way more complex than this ones. What do you think of this? We should still consider it as a Storage even when it have less information?

agrare:
@EsdrasVP So is it a physical disk drive, a group of drives on a controller, a LUN, a single read/write head?  "A storage driver is a disk driver inside a physical storage" is still a little vague.

EsdrasVP:
> @EsdrasVP So is it a physical disk drive, a group of drives on a controller, a LUN, a single read/write head? "A storage driver is a disk driver inside a physical storage" is still a little vague.

@agrare It is a group of physical disk drivers in a controller.

agrare:
> @agrare It is a group of physical disk drivers in a controller.

Okay thanks, I don't want to spend too much time on a name but IMO `storage_driver` isn't descriptive enough.  We should also try to make sure that we're modeling something that can be realistically applied to other vendor's physical storage arrays.

agrare:
@EsdrasVP offline corrected me that a "storage_drive" is actually a single disk, so I proposed "physical_disks" as the table name

EsdrasVP:
@agrare Are the changes ok now?

db/migrate/20180720121837_add_port_status_to_physical_network_ports.rb
```diff
@@ -0,0 +1,5 @@
+class AddPortStatusToPhysicalNetworkPorts < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_network_ports, :port_status, :string
```
agrare: Is this related to this change or did this just sneak in?

EsdrasVP: Sorry for it, this should be in another PR of mine. I'm removing it from here.

EsdrasVP:
ping @Fryguy

agrare:
@EsdrasVP can you add an index for `:physical_storage_id` and `:location`?  Them I'm :+1: on this

agrare:
@bdunne a lot/most of the properties in the Disk model only apply to virtual/cloud envs:
- filename
- mode
- size_on_disk
- start_connected
- backing

Plus Storage and PhysicalStorage are already two different things for similar reasons.

EsdrasVP:
> Is this significantly different from the existing Disk model?

@bdunne Not that much, but there are particular attributes for the physical storage on Physical Infrastructure. Also, this physical disks are less complex than the existing ones that are linked to the Storage model.

db/migrate/20180726142030_create_physical_disk.rb
```diff
@@ -0,0 +1,18 @@
...
+  def change
+    create_table :physical_disks do |t|
+      t.string :model
+      t.string :vendor
+      t.string :status
+      t.string :location
+      t.index  :location
+      t.string :serial_number
+      t.string :health_state
+      t.string :type
```
slemrmartin: @EsdrasVP what is this `type` column? It seems it's not related to definition of STI subclass, but some custom value...
When I ran specs, there is for example "SAS" value

agrare: Ah you're right I assumed it was going to be used for STI.  We'll need to rename this something like disk_type or actually do STI

EsdrasVP: > Ah you're right I assumed it was going to be used for STI. We'll need to rename this something like disk_type or actually do STI

I think we can go for `disk_type` then, since this has nothing to do with existing STI classes.

PR #234, opened 2018-07-13
Create Canister migration

EsdrasVP:
This PR is able to:
- Create a canister migration
- Add its reference to hardwares table

A Canister is a controller inside the Physical Storage (which may have more than one Canister). It contains its own firmware, ports, network, power state, etc.

Depends on:
- [~ManageIQ/manageiq-schema#233 - Merged~](https://github.com/ManageIQ/manageiq-schema/pull/233)
- [ManageIQ/manageiq-schema#269](https://github.com/ManageIQ/manageiq-schema/pull/269)
- [~ManageIQ/manageiq#17700 - Merged~](https://github.com/ManageIQ/manageiq/pull/17700)

### DIFF
--- a/db/migrate/20180713194201_create_canister.rb
+++ b/db/migrate/20180713194201_create_canister.rb
@@ -1,0 +1,23 @@
+class CreateCanister < ActiveRecord::Migration[5.0]
+  def change
+    create_table :canisters do |t|
+      t.string :ems_ref
+      t.string :serial_number
+      t.string :name
+      t.string :position
+      t.string :status
+      t.string :health_state
+      t.string :disk_bus_type
+      t.string :phy_isolation
+      t.string :controller_redundancy_status
+      t.integer :disks
+      t.integer :disk_channel
+      t.integer :system_cache_memory
+      t.string :power_state
+      t.string :host_ports
+      t.string :hardware_version
+      t.bigint :physical_storage_id
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20180713194229_add_canister_id_to_hardwares.rb
+++ b/db/migrate/20180713194229_add_canister_id_to_hardwares.rb
@@ -1,0 +1,5 @@
+class AddCanisterIdToHardwares < ActiveRecord::Migration[5.0]
+  def change
+    add_column :hardwares, :canister_id, :bigint
+  end
+end


EsdrasVP:
@miq-bot assign @agrare

agrare:
@EsdrasVP can you add to the PR description what a canister is?


> Create a canister migration

Isn't enough for me to understand what is being added

EsdrasVP:
> @EsdrasVP can you add to the PR description what a canister is?
> 
> Create a canister migration
> 
> Isn't enough for me to understand what is being added

@agrare I updated the description explaining what a Canister is. Can you see if it's more understandable?

agrare:
@EsdrasVP is Canister an industry standard term for this type of object?  (A quick google search for "storage canister" yielded a lot of amazon results for air-tight coffee bean containers which might just be my recommended search results :laughing:)

If it has its own controller it almost sounds like a caniter is its own self-contained storage?

EsdrasVP:
> @EsdrasVP is Canister an industry standard term for this type of object? (A quick google search for "storage canister" yielded a lot of amazon results for air-tight coffee bean containers which might just be my recommended search results )

@agrare Other companies might call it by a different name, so I think that's why it's only appearing as coffee bean containers on search results hahaha.

> If it has its own controller it almost sounds like a caniter is its own self-contained storage?

Yes, all the work will be done by the Canisters, which will exist inside of a PhysicalStorage. This distinction is important because a Lenovo PhysicalStorage will contain more than one Canister inside it.

agrare:
Okay canister seems like a decent enough name don't want to dwell on that.

db/migrate/20180713194201_create_canister.rb
```diff
@@ -0,0 +1,21 @@
+class CreateCanister < ActiveRecord::Migration[5.0]
+  def change
+    create_table :canisters do |t|
+      t.string :serial_number
```
agrare: 2 things here, can you add ems_ref for a unique reference?  Even if e.g. position is unique for lenovo we can use ems_ref generally.
Also I assume this belongs to a physical_storage but you don't have a reference to physical_storage here.

db/migrate/20180730134426_add_port_status_to_physical_network_ports.rb
```diff
@@ -0,0 +1,5 @@
+class AddPortStatusToPhysicalNetworkPorts < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_network_ports, :port_status, :string
```
agrare: This looks unrelated to the rest of this PR

db/migrate/20180713194201_create_canister.rb
EsdrasVP: @agrare I added now `ems_ref` and `physical_storage_id` to the migration.

db/migrate/20180730134426_add_port_status_to_physical_network_ports.rb
EsdrasVP: @agrare It is, kinda. Because Canister will have some physical ports in it, and I didn't found any reference to this, so I created it. If you want to, I can put it in another PR.

agrare: Yeah lets do this in a new PR since port_status can apply to anything not just canisters.  So a canister will have hardware -> guest_devices -> physical_network_ports?

EsdrasVP: > So a canister will have hardware -> guest_devices -> physical_network_ports?

@agrare Exactly, this relationship it's currently on the core PR.

EsdrasVP: > Yeah lets do this in a new PR since port_status can apply to anything not just canisters.

@agrare I've done it and put as dependency on this PR, could you take a look there?

agrare:
@EsdrasVP should a physical_disk have a canister_id then?

EsdrasVP:
> @EsdrasVP should a physical_disk have a canister_id then?

@agrare No, because the PhysicalDisks are on the same level as Canisters on a PhysicalStorage. It's quite bad actually, because doesn't allow us to see which canister a disk is inside.

PR #237, opened 2018-07-19
Add Owner/Group/Tenant to Authentication

agrare:
Add foreign keys for evm_owner, miq_group, and tenant to the
authentications table to allow ownership of key pairs by users, groups,
and tenants.

https://bugzilla.redhat.com/show_bug.cgi?id=1589766

### DIFF
--- a/db/migrate/20180719162710_add_owner_and_group_to_auth.rb
+++ b/db/migrate/20180719162710_add_owner_and_group_to_auth.rb
@@ -1,0 +1,13 @@
+class AddOwnerAndGroupToAuth < ActiveRecord::Migration[5.0]
+  def up
+    add_column :authentications, :evm_owner_id, :bigint
+    add_column :authentications, :miq_group_id, :bigint
+    add_column :authentications, :tenant_id, :bigint
+  end
+
+  def down
+    remove_column :authentications, :evm_owner_id
+    remove_column :authentications, :miq_group_id
+    remove_column :authentications, :tenant_id
+  end
+end


db/migrate/20180719162710_add_owner_and_group_to_auth.rb
```diff
@@ -0,0 +1,35 @@
...
+    include MigrationStubHelper
+  end
+
+  def up
+    add_column :authentications, :evm_owner_id, :bigint
+    add_column :authentications, :miq_group_id, :bigint
+    add_column :authentications, :tenant_id, :bigint
+
+    say_with_time("Migrate evm_owner and miq_group from reserved table to authentications") do
+      Authentication.includes(:reserved_rec).each do |auth|
```
agrare: TODO: thinking about using the userid string field to store the evm_owner_userid name, if so we will need to clear it here

JPrause:
@miq-bot add_label blocker

JPrause:
@agrare if this can be backported please add the gaprindashvili/yes label

agrare:
@JPrause schema changes cannot be backported
I added code to migrate out of the reserve table so that the model changes can be backported.

db/migrate/20180719162710_add_owner_and_group_to_auth.rb
```diff
@@ -0,0 +1,35 @@
...
+    self.inheritance_column = :_type_disabled
+
+    include ReservedMigrationMixin
+    include MigrationStubHelper
+  end
+
+  def up
+    add_column :authentications, :evm_owner_id, :bigint
+    add_column :authentications, :miq_group_id, :bigint
+    add_column :authentications, :tenant_id, :bigint
```
Ladas: @gtanzillo @lpichler hm, I think that for the rest of the cloud objects, we assign ```cloud_tenant_id```, then the rbac goes through Tenant -> mapped CloudTenant -> cloud objects, do I remember it right?

agrare:
@miq-bot assign @Fryguy 
/cc @gtanzillo @bdunne

agrare:
@bdunne can you take a look?

db/migrate/20180719162710_add_owner_and_group_to_auth.rb
gtanzillo: @agrare Sort of... Once the `CloudTenant` is mapped to a ManageIQ `Tenant` RBAC will use that for tenancy.

db/migrate/20180719162710_add_owner_and_group_to_auth.rb
gtanzillo: I don't think that's necessary because it will get cleared (return nil) when `evm_owner_id` is cleared

spec/migrations/20180719162710_add_owner_and_group_to_auth_spec.rb
```diff
@@ -0,0 +1,63 @@
+require_migration
+
+describe AddOwnerAndGroupToAuth do
+  class User < ActiveRecord::Base; end
```
bdunne: We don't usually define AR::Base classes in the specs.  Does this wind up nested under some kind of dynamically created RSpec ExampleGroup class?  If not, these should probably be moved into the migration and use a `migration_stub` here to avoid namespace collision.

agrare:
@gtanzillo @bdunne looks like the target release for this BZ changed so we don't need to backport.  Removing all of the reserved mixin stuff.
@bdunne IDK the specifics of ownership e.g. if it can have a user/group/tenant set at the same time just following the example of the other models.

PR #239, opened 2018-07-30
[WIP] Adding refresh_queue table

Ladas:
Adding refresh_queue table, as a temporary solution for refresh,
before we have a normal queue. We will be using this table as a
normal queue, which will be more efficient, even if the DB is
not really good option for this.

### DIFF
--- a/db/migrate/20180730092457_create_refresh_queue_table.rb
+++ b/db/migrate/20180730092457_create_refresh_queue_table.rb
@@ -1,0 +1,16 @@
+class CreateRefreshQueueTable < ActiveRecord::Migration[5.0]
+  def change
+    create_table :refresh_queue do |t|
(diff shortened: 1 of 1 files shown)


Ladas:
@miq-bot assign @agrare

Fryguy:
Not sure why we wouldn't just use an actual queue?

agrare:
@Ladas I think for classic MIQ we can just do streaming stuff in a single worker and not need a queue.  Anything actually _requiring_ a queue (e.g. bring your own image) we can just do on the new platform only

agrare:
Given that I don't think we'll be splitting the collectors and persisters without also introducing an actual queue system I'm going to close this.  We can revisit if it turns out that we actually need this.

PR #242, opened 2018-08-07
Create a ConversionHost table

agrare:
This adds a table for conversion hosts which can be polymorphically related to other inventory items i.e. Vms/Hosts.

### DIFF
--- a/db/migrate/20180807153714_add_conversion_host_table.rb
+++ b/db/migrate/20180807153714_add_conversion_host_table.rb
@@ -8,6 +8,7 @@ class AddConversionHostTable < ActiveRecord::Migration[5.0]
       t.bigint  :resource_id
       t.string  :version
       t.integer :max_concurrent_tasks
+      t.text    :capabilities
       t.timestamps
       t.index %w(resource_id resource_type)
     end
--- a/db/migrate/20180807153714_add_conversion_host_table.rb
+++ b/db/migrate/20180807153714_add_conversion_host_table.rb
@@ -1,14 +1,13 @@
 class AddConversionHostTable < ActiveRecord::Migration[5.0]
   def change
     create_table :conversion_hosts do |t|
-      t.string  :name
-      t.string  :address
-      t.string  :type
-      t.string  :resource_type
-      t.bigint  :resource_id
-      t.string  :version
-      t.integer :max_concurrent_tasks
-      t.text    :capabilities
+      t.string     :name
+      t.string     :address
+      t.string     :type
+      t.references :resource, :type => :bigint, :polymorphic => true
+      t.string     :version
+      t.integer    :max_concurrent_tasks
+      t.text       :capabilities
       t.timestamps
       t.index %w(resource_id resource_type)
     end
--- a/db/migrate/20180807153714_add_conversion_host_table.rb
+++ b/db/migrate/20180807153714_add_conversion_host_table.rb
@@ -1,0 +1,13 @@
+class AddConversionHostTable < ActiveRecord::Migration[5.0]
+  def change
+    create_table :conversion_hosts do |t|
+      t.string :name
+      t.string :address
+      t.string :type
+      t.string :resource_type
+      t.bigint :resource_id
+      t.timestamps
+      t.index %w(resource_id resource_type)
+    end
+  end
+end
--- a/db/migrate/20180807153714_add_conversion_host_table.rb
+++ b/db/migrate/20180807153714_add_conversion_host_table.rb
@@ -1,11 +1,13 @@
 class AddConversionHostTable < ActiveRecord::Migration[5.0]
   def change
     create_table :conversion_hosts do |t|
-      t.string :name
-      t.string :address
-      t.string :type
-      t.string :resource_type
-      t.bigint :resource_id
+      t.string  :name
+      t.string  :address
+      t.string  :type
+      t.string  :resource_type
+      t.bigint  :resource_id
+      t.string  :version
+      t.integer :max_concurrent_tasks
       t.timestamps
       t.index %w(resource_id resource_type)
     end
--- a/db/migrate/20180807153714_add_conversion_host_table.rb
+++ b/db/migrate/20180807153714_add_conversion_host_table.rb
@@ -7,7 +7,8 @@ class AddConversionHostTable < ActiveRecord::Migration[5.0]
       t.references :resource, :type => :bigint, :polymorphic => true
       t.string     :version
       t.integer    :max_concurrent_tasks
-      t.text       :capabilities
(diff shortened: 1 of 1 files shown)


agrare:
@fdupont-redhat What other properties do we need to add to a conversion host?

fdupont-redhat:
@agrare If it's a VM, how do we get the credentials ? For the host it's already there, but AFAIK it's not for the VM. Or at least it's not visible in the UI.

agrare:
@fdupont-redhat that's a different issue :)  Lets tackle that in a different PR/Trello card.

fdupont-redhat:
@agrare LGTM, then.

fdupont-redhat:
We don't really have a notion of version, but capabilities might be good. Currently, we use tags to store the transport method, but it could be an attribute. Could it be a JSON field ? I don't know what we could store in there: virt-v2v capabilities, transport methods, other things that we don't know yet.

@nyoxi @aufi any idea ?

aufi:
E.g. some Conversion host build number would be nice to detect old ones..

nyoxi:
I can imagine storing various capabilities and versions could make sense. But that brings the question how will you keep them up-to-date with what's really on the host?

agrare:
@nyoxi that would have to be done by whoever communicates with the code on the conversion host.  This won't be able to be updated by inventory refresh because it won't be based on properties available to the RefreshWorker in provider inventory.

agrare:
Okay @nyoxi I added a version string and a serialized capabilities so we can have some flexibility in what we store there.  As long as we're not planning on doing any queries based on capabilities we should be good with that.

fdupont-redhat:
@agrare we do requests based on capabilities. Currently, the only capability we use in transport_method, which can be vddk, ssh, whatever we come up with. IIUC, it might be a good idea to have a specific column for that one, isn't it ?

agrare:
@bdunne can you take a look?

agrare:
> Currently, the only capability we use in transport_method, which can be vddk, ssh, whatever we come up with. IIUC, it might be a good idea to have a specific column for that one, isn't it ?

Yes it would, I was going off of:
> I can imagine storing various capabilities

which seemed pretty vague :smile:

If a transport_method boolean column is all that you need then I'll add that.

fdupont-redhat:
That's all I need. It's the only attribute that is used during host selection, apart from current number of migrations which is calculated.

agrare:
@fdupont-redhat is that going to be a single value or is it possible for multiple to be supported?

fdupont-redhat:
In the current implementation, we use a single value as it also decides the transport method to use for the migration. But IMO, it should be a per-mapping/plan/vm choice. So, being able to use the same host for many transport methods should be possible, as I don't see why a conversion wouldn't be able to use more than one transport method.

Please make it a multiple value attribute.

agrare:
> Please make it a multiple value attribute.

Going to need multiple columns then, don't want to be doing LIKE queries on a comma separated list :/

fdupont-redhat:
You mean one boolean column per transport method, right ?

agrare:
@fdupont-redhat correct, is this something that can be inferred from the version?  Or is it something that could be different for conversion hosts on the same version?

db/migrate/20180807153714_add_conversion_host_table.rb
```diff
@@ -0,0 +1,16 @@
+class AddConversionHostTable < ActiveRecord::Migration[5.0]
+  def change
+    create_table :conversion_hosts do |t|
+      t.string  :name
+      t.string  :address
+      t.string  :type
+      t.string  :resource_type
```
bdunne: `t.references :resource, :type => :bigint, :polymorphic => true`

agrare: :+1: that's slick

fdupont-redhat:
@agrare It's not related to the version. Version will most likely refer to ovirt-ansible-v2v-conversion-host package version.

agrare:
@fdupont-redhat okay I added two columns to represent if vddk and/or ssh transport is supported, PTAL

db/migrate/20180809155959_add_conversion_host_id_to_miq_tasks.rb
```diff
@@ -0,0 +1,6 @@
+class AddConversionHostIdToMiqTasks < ActiveRecord::Migration[5.0]
+  def change
+    add_column :miq_tasks, :conversion_host_id, :bigint
```
bdunne: Is this necessary?  I'm assuming that the `Host` would be the resource on the `ConversionHost`, right?

Something like:
```ruby
class ConversionHost < ApplicationRecord
  belongs_to :resource, :polymorphic => true
  ...
end

class Host < ApplicationRecord
  has_one :conversion_host, :as => :resource
  ...
end
```

fdupont-redhat:
:+1:

db/migrate/20180809155959_add_conversion_host_id_to_miq_tasks.rb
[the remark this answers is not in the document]
agrare: We wanted to add a method for throttling how many tasks run on a conversion host without having to loop through all active tasks and pull the conversion host out of the task options.

I don't see a trello card yet (/cc @fdupont-redhat) but there is an ongoing design doc here https://docs.google.com/document/d/1ANBPzDLF-id3MS9_8ztmrNjNwh3FJ6yGCvA2BUE808c/edit#heading=h.lzsppid3wjo9

Since this is still in design maybe we should hold off on this and come back and add this later.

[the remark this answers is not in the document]
fdupont-redhat: Here's the Trello card: https://trello.com/c/DOClRvfD/189-migration-throttling-cf-model

fdupont-redhat:
@agrare As the `resource` attribute is polymorphic, what is the purpose of `address` ? I'd expect it to be the IP address to connect to the conversion host, but for an oVirt host it maps to `resource.ipaddress`, while for an OpenStack VM it could map to `resource.ipaddresses.first` or `resource.ipv4_address`...

How would you implement the polymorphism ? In the ConversionHost class using `send` and `resource_type` ?

agrare:
@fdupont-redhat it was for the case where the conversion host is not in MIQ inventory e.g. a bare metal conversion host.

It looks like both the host and the vm models have an `#ipaddresses` method that you can use.  I'd recommend a method on `ConversionHost#ipaddress` which could do `address || resource.ipaddresses.first` that should cover all cases

fdupont-redhat:
@agrare Thanks for the refresh. I'll create IP address methods in ConversionHost class.

PR #243, opened 2018-08-07
Add an empty database spec

bdunne:
An empty database (migrated form 0) should remain empty, only the schema itself should change.

Adding this as an automated code review to address https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-411142335


spec/automated_review/empty_database_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe "Empty Database" do
+  let(:connection) { ApplicationRecord.connection }
```
carbonin: I know we use `ActiveRecord::Base` in the actual migrations. Should we use it here too?

spec/automated_review/empty_database_spec.rb
```diff
@@ -0,0 +1,11 @@
+describe "Empty Database" do
+  it "after migrated remains empty" do
+    counts = ApplicationRecord.connection.tables.each_with_object([]) do |t, array|
```
carbonin: Should we use `ActiveRecord::Base.connection` here? I know we use it in the migrations, but wasn't sure if we should in the specs as well.

spec/automated_review/empty_database_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+      next if ManageIQ::Schema::SYSTEM_TABLES.include?(t)
+      count = ApplicationRecord.connection.execute("SELECT COUNT(*) FROM #{t};").getvalue(0, 0)
+      array << "#{t}: #{count}" if count.positive?
```
carbonin: What about negative row counts :scream: :trollface:

PR #245, opened 2018-08-08
Add a STI disabled spec

bdunne:
Adding this as an automated code review to address https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-310643185

Also fixing the current failures

### DIFF
--- a/spec/support/migration_helper.rb
+++ b/spec/support/migration_helper.rb
@@ -15,6 +15,22 @@ module Spec
 
             it("with empty tables") { migrate }
 
+            it "STI is disabled for all involved ActiveRecord::Base descendants" do
+              classes = described_class.constants.collect do |symbol|
+                const = described_class.const_get(symbol)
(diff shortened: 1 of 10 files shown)


Fryguy:
So, one issue I see is that this uses the current schema to check "old" migrations, which is invalid.  You need to use the schema at the time of the migration to determine if the inheritance disablement is necessary.

bdunne:
@Fryguy Updated, please take a look

spec/support/migration_helper.rb
```diff
@@ -15,6 +15,19 @@ def migration_context(direction, &block)
 
             it("with empty tables") { migrate }
 
+            it "STI is disabled for all involved ActiveRecord::Base descendants" do
+              classes = ActiveRecord::Base.descendants.select { |klass| klass.namespace.first == described_class.to_s }.compact
```
Fryguy: Might be better to flip this around to something like

```ruby
described_class.constants.select { |c| c < ActiveRecord::Base }
```

PR #247, opened 2018-08-08
Add a test requiring data migrations to have an associated spec file

bdunne:
We can determine which files are data migrations since they will define a class that inherits from ActiveRecord::Base.  If we hook inherited in AR::Base, we can get the constant and the file in which it is defined (the data migration).

Adding this as an automated code review to address https://github.com/ManageIQ/manageiq-schema/issues/4#issue-237082801

### DIFF
--- a/spec/automated_review/data_migrations_need_tests_spec.rb
+++ b/spec/automated_review/data_migrations_need_tests_spec.rb
@@ -1,0 +1,29 @@
+describe "Data migrations" do
+  KNOWN_MISSING_DATA_MIGRATION_SPEC_FILES = [
+    "20151021174044_add_tenant_default_group_spec.rb",
+    "20160317041206_add_maintenance_to_host_spec.rb",
+    "20160428215808_add_filters_to_entitlements_spec.rb",
+    "20160713130940_remove_type_template_and_vms_filters_from_miq_search_spec.rb",
+    "20170419154137_remove_deleted_migration_timestamps_spec.rb",
+    "20171011180000_move_openstack_refresher_settings_spec.rb"
+  ].freeze
+
+  def data_migration_files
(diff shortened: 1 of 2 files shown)


spec/automated_review/data_migrations_need_tests_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  def data_migration_files
+    Spec::Support::ConstantWatcher.classes_by_file.keys.select { |file| file.include?("db/migrate") }
+  end
+
+  it "need tests" do
+    Dir.glob(ManageIQ::Schema::Engine.root.join("db", "migrate", "*.rb")).each { |i| require i }
+
+    data_migration_files.each do |data_migration_file|
+      spec_file_basename = File.basename(data_migration_file).sub(".rb", "_spec.rb")
+      next if spec_file_basename.in?(KNOWN_MISSING_DATA_MIGRATION_SPEC_FILES)
```
Fryguy: I think move this down and XOR the comparison, such that if someone adds a test that's in the missing list, they are forced to remove it.

spec/support/constant_watcher.rb
```diff
@@ -0,0 +1,21 @@
...
+        path = caller_locations(2..2).first.path
+        self.classes_by_file ||= {}
+        self.classes_by_file[path] ||= []
+        self.classes_by_file[path] << const.name
+      end
+    end
+  end
+end
+
+ActiveRecord::Base.singleton_class.prepend(Spec::Support::ConstantWatcher)
```
Fryguy: Can you verify this works locally if you do `bundle exec rspec spec/specific/spec/file.rb`

Fryguy:
#249 is merged, so please update the KNOWN list.

spec/support/constant_watcher.rb
bdunne: Yes, it's loaded by spec_helper

PR #248, opened 2018-08-09
Ensure id and *_id columns are bigint

bdunne:
Based on changes in https://github.com/ManageIQ/manageiq-schema/pull/245

Adding this as an automated code review to address https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-380508905

### DIFF
--- a/spec/automated_review/id_columns_are_bigint_spec.rb
+++ b/spec/automated_review/id_columns_are_bigint_spec.rb
@@ -1,0 +1,41 @@
+require 'set'
+describe "*_id columns" do
+  EXCEPTIONS_LIST = Set.new(
+    [
+      "authentications.ldap_id",
+      "authentications.rhsm_pool_id",
+      "cloud_networks.provider_segmentation_id",
+      "container_volumes.common_volume_id",
+      "miq_queue.task_id",
+      "network_ports.binding_host_id",
+      "scan_histories.task_id",
+      "sessions.session_id"
+    ]
+  ).freeze
+
+  it "should be type bigint" do
+    ActiveRecord::Base.connection.tables.each do |table|
+      next if ManageIQ::Schema::SYSTEM_TABLES.include?(table)
+      DatabaseHelper.columns_for_table(table).each do |column|
+        next unless column.name.end_with?("_id")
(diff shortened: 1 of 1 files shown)


spec/automated_review/id_columns_are_bigint_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    exceptions_list = [
+      "authentications.ldap_id",
+      "authentications.rhsm_pool_id",
+      "cloud_networks.provider_segmentation_id",
+      "container_volumes.common_volume_id",
+      "miq_queue.task_id", # <--WAT? Expected miq_queue.task_id to be type 'bigint', got: character varying
+      "network_ports.binding_host_id",
+      "scan_histories.task_id", # <--WAT?
+      "sessions.session_id"
+    ].freeze
```
Fryguy: Minor, but maybe make this a Set, since we are doing continual lookups

Fryguy:
As discussed, the tests fail and tell the user to remove the column from the exclusion list when they fix it.

Also, the exclusions_list should be a constant like in the other PR>

spec/automated_review/id_columns_are_bigint_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      "network_ports.binding_host_id",
+      "scan_histories.task_id",
+      "sessions.session_id"
+    ]
+  ).freeze
+
+  it "should be type bigint" do
+    ActiveRecord::Base.connection.tables.each do |table|
+      next if ManageIQ::Schema::SYSTEM_TABLES.include?(table)
+      DatabaseHelper.columns_for_table(table).each do |column|
```
Fryguy: Moving forward, perhaps this should move into a DatabaseHelper method...

```ruby
def DatabaseHelper.each_table
  ActiveRecord::Base.connection.tables.each do |table|
    next if ManageIQ::Schema::SYSTEM_TABLES.include?(table)
    yield table
  end
end

def DatabaseHelper.each_column
  each_table do |table|
    columns_for_table(table).each do |column|
      yield table, column
    end
  end
end
```

Then caller just does

```ruby
DatabaseHelper.each_column do |table, column|
   ...
end
```

PR #250, opened 2018-08-10
Skip updating rows with vm_ems_ref=nil for EventStream

lpichler:
During upgrade customer's db to latest it spent some time on this query.

And then I saw that `EventStream#vm_ems_ref` was nil. So we can skip them in updating process.
(Not sure what does it mean)

### DIFF
--- a/db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
+++ b/db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
@@ -67,7 +67,7 @@ class AzureBackslashToForwardSlash < ActiveRecord::Migration[5.0]
     end
 
     say_with_time("Updating Azure event stream ems_ref") do
-      EventStream.where(:source => "AZURE").update_all("vm_ems_ref = replace(vm_ems_ref, '\\', '/')")
+      EventStream.where(:source => "AZURE").where.not(:vm_ems_ref => nil).update_all("vm_ems_ref = replace(vm_ems_ref, '\\', '/')")
     end
   end
 end


lpichler:
There are more similar queries in this migration, @djberg96 do you think that it worth to add such condition for them as well ? 
Thanks!

lpichler:
@miq-bot assign @Fryguy

djberg96:
@lpichler If `vm_ems_ref` is null then there's no need to modify it, so this should be fine, and it looks like the specs are passing. Please modify any other code that would benefit from this as well.

lpichler:
@djberg96 I updated rest, thanks!

djberg96:
@lpichler Actually, maybe we should add a check for `uid_ems` as well. I don't *think* you would ever have one without the other, but just in case.

Also, the current failure appears to be unrelated, as it is coming from a picture blob migration.

db/migrate/20180424141617_azure_backslash_to_forward_slash.rb
```diff
@@ -26,6 +26,7 @@ class EventStream < ActiveRecord::Base
   def up
     say_with_time("Updating Azure VM ems_ref, uid_ems and description") do
       Vm.where(:type => "ManageIQ::Providers::Azure::CloudManager::Vm")
+        .where.not(:ems_ref => nil)
```
agrare: The `EventStream.vm_ems_ref` being nil doesn't mean that the VM's ems_ref is nil...just means that the event wasn't about a VM.

I don't think it is possible (or at least it really shouldn't be) to have a vm with a nil ems_ref.

djberg96: @agrare so then we should just stick with the original PR for event streams I think.

lpichler:
@djberg96 @agrare Ok, I leave it only for EventStream, thanks for your feedback!

djberg96:
:+1:

PR #254, opened 2018-08-14
Add cancelation_status column in MiqRequest and MiqRequestTask

hsong-rh:
Add `cancelation_status` to enable cancel operation for V2V migration.

This PR is related with:
https://github.com/ManageIQ/manageiq/pull/17825
https://github.com/ManageIQ/manageiq-automation_engine/pull/212

### DIFF
--- a/db/migrate/20180813141056_add_cancelation_status_to_miq_request.rb
+++ b/db/migrate/20180813141056_add_cancelation_status_to_miq_request.rb
@@ -1,0 +1,6 @@
+class AddCancelationStatusToMiqRequest < ActiveRecord::Migration[5.0]
+  def change
+    add_column :miq_requests,      :cancelation_status, :string
+    add_column :miq_request_tasks, :cancelation_status, :string
+  end
+end


agrare:
@hsong-rh couldn't the cancel states just be additional types of request_state ?

hsong-rh:
@agrare We follow the same logic as `approval_state` and feel that cancel state is different from request states. It may occur at any request_state and may take some time to finish the cancellation.

agrare:
I'll defer to @gmcculloug on this one, but what happens when we add "pause", are we going to add a "pause_state" column?  How about "resume", ..."abort"?  These all seem like general "state" and do not require dedicated columns IMO.

gmcculloug:
I am setting up a meeting for the end of today to get the relevant people together to work through the details of this feature.  Stay tuned.

bdunne:
@gmcculloug Any update?  Should this be marked as WIP in the mean time?

gmcculloug:
Yes, we are working on changes that do not require schema updates.  We should mark this as WIP for now and hopefully close once we feel the approach will work.

@miq-bot add_label wip

hsong-rh:
After discussion, we still want to introduce new column for cancel request. No lock and backport are needed so far.

@agrare @Fryguy @gmcculloug please review and merge.

bdunne:
Can anyone elaborate on this and how it will be used?  Also, why would `cancel_request` be a string?  It sounds like a boolean to me, but I don't have anything to base this off of...

hsong-rh:
@bdunne `cancel_request` will have 3 different values: `cancel_requested`, `canceling` and `canceled`, means cancellation is pending, processing and done. This is the reason we didn't choose to use boolean.

bdunne:
Maybe `cancelation_status` would be a better name? What do you think @gmcculloug ?

carbonin:
:+1: On `cancelation_status`

`cancel_request` sounds like it should be used like "Should this request be canceled" rather than anything having to do with status. Or if you're basing the logic on `approval_state` maybe this should be `cancelation_state`?

hsong-rh:
@bdunne @carbonin Changed name to `cancelation_status`, please review and merge it.

PR #258, opened 2018-08-20
Model relation "SecurityGroup belongs to NetworkRouter or CloudSubnet"

miha-plesko:
With this commit we add a foreign key on SecurityGroup to follow what Nuage needs. In Nuage, security group is something that only exists in a scope of either L3 domain (NetworkRouter) or L2 domain (CloudSubnet::L2).

image

MIQ is currently not able to model such relation yet, hence this PR.

@miq-bot assign @Fryguy 

@Ladas @agrare looking forward to your opinion. Need for this was revealed when enhancing targeted refresh where we would need something like `network_router.security_groups`, but there is no such relation yet... Is schema already frozen for Hammer?

### DIFF
--- a/db/migrate/20180817152200_add_network_router_id_to_security_group.rb
+++ b/db/migrate/20180817152200_add_network_router_id_to_security_group.rb
@@ -1,0 +1,5 @@
+class AddNetworkRouterIdToSecurityGroup < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :security_groups, :network_router, :type => :bigint, :index => true
+  end
+end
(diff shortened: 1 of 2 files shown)


Ladas:
@miq-bot assign @agrare

miha-plesko:
@agrare are you okay with this?

miha-plesko:
Kindly ask either @Fryguy or @agrare to get this one in. Need to do a bunch of followup PRs, namely:

- inform Rails about the two foreign keys in miq-core repo
- update ui-classic to render relation
- update Nuage targeted refresh that relies on this relation

And would't want to get too close to code-freeze.

agrare:
@miha-plesko you don't need the parens for `add_reference` no matter what miq-bot says :)

miha-plesko:
@agrare so no parens shall be.

PR #262, opened 2018-08-23
Moving location_led_state to asset_details table

douglasgabriel:
__This PR is able to:__
- Move the `location_led_state` column from `physical_chassis` and `physical_servers` to `asset_details`

__Goal__
The goal is to put together ems_ref and state information of the location led.

It is a complement of https://github.com/ManageIQ/manageiq-schema/pull/231

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -1,4 +1,8 @@
 ActiveSupport::Inflector.inflections do |inflect|
+  inflect.singular(/Chassis$/, "Chassis")
+  inflect.plural(/Chassis$/, "Chassis")
+  inflect.singular(/chassis$/, "chassis")
+  inflect.plural(/chassis$/, "chassis")
   inflect.singular(/Queue$/, "Queue")
   inflect.plural(/Queue$/, "Queue")
   inflect.singular(/queue$/, "queue")
--- a/db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
+++ b/db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
@@ -1,0 +1,44 @@
+class MoveLocationLedStateToAssetDetailsTable < ActiveRecord::Migration[5.0]
+  class AssetDetail < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  class PhysicalChassis < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  class PhysicalServer < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :asset_details, :location_led_state, :string
+
+    PhysicalChassis.in_my_region.find_each { |chassis| move_location_led_state_to_asset_detail("PhysicalChassis", chassis) }
+    PhysicalServer.in_my_region.find_each { |server| move_location_led_state_to_asset_detail("PhysicalServer", server) }
+
+    remove_column :physical_chassis, :location_led_state
+    remove_column :physical_servers, :location_led_state
+  end
+
+  def down
+    add_column :physical_chassis, :location_led_state, :string
+    add_column :physical_servers, :location_led_state, :string
+
+    AssetDetail.where(:resource_type => %w(PhysicalChassis PhysicalServer)).in_my_region.find_each do |asset_detail|
+      self.class.const_get(asset_detail.resource_type).update(asset_detail.resource_id, :location_led_state => asset_detail.location_led_state)
+    end
+
+    remove_column :asset_details, :location_led_state
+  end
+
+  private
+
+  def move_location_led_state_to_asset_detail(class_name, asset)
+    AssetDetail.in_my_region.find_or_initialize_by(:resource_type => class_name, :resource_id => asset.id).update_attributes!(:location_led_state => asset.location_led_state)
+  end
+end
--- a/spec/migrations/20180823111741_move_location_led_state_to_asset_details_table_spec.rb
+++ b/spec/migrations/20180823111741_move_location_led_state_to_asset_details_table_spec.rb
@@ -1,0 +1,53 @@
+require_migration
+
+describe MoveLocationLedStateToAssetDetailsTable do
+  let(:asset_detail_stub)     { migration_stub(:AssetDetail) }
+  let(:physical_chassis_stub) { migration_stub(:PhysicalChassis) }
+  let(:physical_server_stub)  { migration_stub(:PhysicalServer) }
+
+  context "resources with asset detail" do
+    migration_context :up do
+      let!(:physical_chassis) { physical_chassis_stub.create!(:location_led_state => 'Blinking') }
+      let!(:physical_server)  { physical_server_stub.create!(:location_led_state => 'Off') }
+
+      it "should migrate the data from physical_chassis and physical_servers to asset_details table" do
+        asset_detail_chassis = asset_detail_stub.create!(:resource_type => "PhysicalChassis", :resource_id => physical_chassis.id)
+        asset_detail_server  = asset_detail_stub.create!(:resource_type => "PhysicalServer", :resource_id => physical_server.id)
+
+        migrate
+
+        expect(asset_detail_chassis.reload.location_led_state).to eq('Blinking')
+        expect(asset_detail_server.reload.location_led_state).to  eq('Off')
+      end
+
+      it "should create associated asset_detail and migrate the data if it doesn't exist" do
+        migrate
+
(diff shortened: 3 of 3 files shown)


douglasgabriel:
@miq-bot assign @Fryguy 
@miq-bot add_reviewer @agrare

carbonin:
Seems likely @agrare. I'm assuming #183 was not released in gaprindashvili?

agrare:
@carbonin correct, tag --contains shows some 5.10 tags but no 5.9

config/initializers/inflections.rb
```diff
@@ -7,6 +7,10 @@
   inflect.plural(/quota$/, "quotas")
   inflect.singular(/Quota$/, "Quota")
   inflect.plural(/Quota$/, "Quotas")
+  inflect.singular(/Chassis$/, "Chassis")
```
bdunne: Please sort alphabetically

db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
```diff
@@ -0,0 +1,65 @@
...
+    self.inheritance_column = :_type_disabled
+
+    has_one :asset_detail, :as => :resource, :dependent => :destroy
+  end
+
+  def up
+    add_column :asset_details, :location_led_state, :string
+
+    PhysicalServer.includes(:asset_detail).each do |server|
+      asset_detail = AssetDetail.where('resource_id = ? and resource_type like ?', server.id, "%PhysicalServer")&.first
```
bdunne: Any reason not to use `server.asset_detail` here?

bdunne: This could be one-lined as:
```ruby
server.asset_detail&.update_attributes!(:location_led_state => server.location_led_state)
```

bdunne: Shouldn't we create the related `AssetDetail` if it doesn't exist so that we don't lose data?

db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
```diff
@@ -0,0 +1,65 @@
...
+      if asset_detail
+        asset_detail.location_led_state = server.location_led_state
+        asset_detail.save
+      end
+    end
+
+    PhysicalChassis.all.each do |chassis|
+      asset_detail = AssetDetail.where('resource_id = ? and resource_type like ?', chassis.id, "%PhysicalChassis")&.first
```
bdunne: same comments here

db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
```diff
@@ -0,0 +1,65 @@
...
+    remove_column :physical_servers, :location_led_state
+    remove_column :physical_chassis, :location_led_state
+  end
+
+  def down
+    add_column :physical_servers, :location_led_state, :string
+    add_column :physical_chassis, :location_led_state, :string
+
+    PhysicalServer.all.each do |server|
+      asset_detail = AssetDetail.where('resource_id = ? and resource_type like ?', server.id, "%PhysicalServer")&.first
```
bdunne: same comments here

db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
```diff
@@ -0,0 +1,65 @@
...
+      if asset_detail
+        server.location_led_state = asset_detail.location_led_state
+        server.save
+      end
+    end
+
+    PhysicalChassis.all.each do |chassis|
+      asset_detail = AssetDetail.where('resource_id = ? and resource_type like ?', chassis.id, "%PhysicalChassis")&.first
```
bdunne: same comments here

spec/migrations/20180823111741_move_location_led_state_to_asset_details_table_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        :asset_detail       => asset_detail_chassis,
+        :location_led_state => 'Blinking'
+      )
+
+      migrate
+
+      expect(asset_detail_server.reload.location_led_state).to  eq('Off')
+      expect(asset_detail_chassis.reload.location_led_state).to eq('Blinking')
+    end
+  end
```
bdunne: I think there should also be a test where there is a `PhysicalServer` and `PhysicalChassis` and no `AssetDetails`.

bdunne: (and same for down)

carbonin:
Okay, so this makes me feel like we'll have to go the more invasive route to ensure that changes like this don't break replication.

But I would also like to point out that this type of flip-floping on schema has not occurred since the introduction of pglogical replication in the darga release. I want to take this opportunity to push toward more thoughtful schema design so that we don't have as much column churn. It seems like we could have avoided all of this if the design for the `switches` and `physical_servers`/`physical_chasis` were better planned to begin with.

db/migrate/20180823111741_move_location_led_state_to_asset_details_table.rb
douglasgabriel: Actually I had some problems with `server.asset_detail` so I code this way, but now I did this change and everything looks fine, so thanks for the tip :+1: . About create the related `AssetDetail` I was concerned about those resources that, for some reason, did not have its `AssetDetail` filled, so I decide to respect that :smile: . I'm sure that Lenovo's resources always will have `AssetDetail` but maybe other vendor could not fill this information. Do you think that it is a valid concern :thinking: ?

douglasgabriel: Oh, I think that I got you concern. If we don't create the relation, so the data will be loose. Therefore, agreed, we should create the relation

spec/migrations/20180823111741_move_location_led_state_to_asset_details_table_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          :location_led_state => 'Off'
+        )
+
+        chassis = physical_chassis_stub.create!(
+          :location_led_state => 'Blinking'
+        )
+
+        migrate
+
+        expect(server.reload.asset_detail.location_led_state).to  eq('Off')
```
bdunne: Can you verify that the `asset_detail.resource_type` is `"PhysicalServer"` not `"MoveLocationLedStateToAssetDetailsTable::PhysicalServer"`

douglasgabriel: But, it is `MoveLocationLedStateToAssetDetailsTable::PhysicalServer`, once I'm using the class defined inside `MoveLocationLedStateToAssetDetailsTable`. I thought that it is the expected result :thinking:. Anyway, is there something that I'm missing here to make the asset_detail.resource_type point to "PhysicalServer" ?

bdunne: The problem is that `MoveLocationLedStateToAssetDetailsTable::PhysicalServer` only exists during migration.  For production use, only `PhysicalServer` will exist.

douglasgabriel: I'm not totally familiarized with the schema tests environment, please, correct me if I'm wrong. As far as I know, the tests suite doesn't has access to ManageIQ models, so we need to define those models like I'm doing in [here](https://github.com/ManageIQ/manageiq-schema/pull/262/files#diff-a18ede530fc209a9f9daed548de9f3b1R8), so when I create the relationship between server and asset_detail inside the specs the `asset_detail.resource_type` will be `MoveLocationLedStateToAssetDetailsTable::PhysicalServer`. Anyway, if I run this migration, everything works fine, once it is changing the right tables and tuples.

If you insist that I must do this test, I will need to configure some more thing to make `asset_detail.resource_type` point to `"PhysicalServer"`, could you tell me what is this configuration that I'm missing here?

douglasgabriel:
@bdunne @carbonin and @agrare , Sorry but, we're in a bit of a hurry to get this merged. Do you think that will be better just add the new column in asset_details table and just update our parser to put the information in the right place, without migrate the data here?

Once we have a dependency on what @carbonin reported before?

spec/migrations/20180823111741_move_location_led_state_to_asset_details_table_spec.rb
bdunne: Here's how I would write it to avoid the incorrect class names... https://github.com/douglasgabriel/manageiq-schema/pull/1

douglasgabriel:
@bdunne , I did the changes that you recommended and everything looks fine now. Although, I don't understood the Hakiri complaint :thinking:

bdunne:
@carbonin can you also review this?

I'm not sure what's wrong with Hakiri.

carbonin:
Looks like github is having some issues this morning. Probably causing the Hakiri failures as it's failing when trying to clone the repo.

image

douglasgabriel:
Done @bdunne . Thanks for the tips here :smile:

PR #264, opened 2018-08-28
[WIP] Combine switches column migrations

carbonin:
This causes an issue with upgrading replicated environments

When the first migration is run, data is generated which references
the new columns in the switches table, but those columns are then removed.

When replication sees data which references a column not in the global
schema it exits and tries again later (typically within a few minutes).
If we then migrate the global database to add then quickly remove the
column in question the data from the remote stream will never be
able to be successfully consumed.

This fixes the issue by not adding the column to the wrong table
to start with.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1601015

This is one option to fix the issue, but it does edit existing migrations which I acknowledge is something we don't want to do lightly. Additionally this makes no attempt to detect or fix this kind of issue for future releases.

### DIFF
--- a/db/migrate/20180126144529_add_extra_information_to_switches.rb
+++ b/db/migrate/20180126144529_add_extra_information_to_switches.rb
@@ -4,10 +4,5 @@ class AddExtraInformationToSwitches < ActiveRecord::Migration[5.0]
     add_column :switches, :type, :string
     add_column :switches, :health_state, :string
     add_column :switches, :power_state, :string
-    add_column :switches, :product_name, :string
-    add_column :switches, :part_number, :string
-    add_column :switches, :serial_number, :string
-    add_column :switches, :description, :string
-    add_column :switches, :manufacturer, :string
   end
 end
--- a/db/migrate/20180323204821_remove_switch_details_from_switches.rb
+++ b/db/migrate/20180323204821_remove_switch_details_from_switches.rb
@@ -1,9 +1,0 @@
-class RemoveSwitchDetailsFromSwitches < ActiveRecord::Migration[5.0]
-  def change
-    remove_column :switches, :product_name,  :string
-    remove_column :switches, :part_number,   :string
-    remove_column :switches, :serial_number, :string
-    remove_column :switches, :description,   :string
-    remove_column :switches, :manufacturer,  :string
-  end
-end


carbonin:
I have an alternative, more invasive, PR to compare with this one on the way.

carbonin:
The alternate solution would be the pair of PRs:

https://github.com/ManageIQ/manageiq-schema/pull/266
https://github.com/ManageIQ/manageiq/pull/17919

agrare:
@carbonin this makes sense as a fix for this specific issue, your other PRs sound good to resolve this generally though.

carbonin:
Yeah, that's what we're trying to resolve. Do we add code which could disrupt a vitally important part of activerecord (other route) or deal with this possibly happening again (this route).

I, personally, would vote for making the targeted fix here as these migrations haven't been released yet and leave `ActiveRecord::Migration#migrate` alone. But I wanted to at least POC both changes so we could have a better discussion.

Also, the full explanation of the issue is here https://bugzilla.redhat.com/show_bug.cgi?id=1601015#c10 if anyone wants to read the whole write-up.

gtanzillo:
>👎 Two reasons:
>...

Afaik, this is the first time we've seen this. Given that the alternative approach in the other PRs create actual risk because we're adding logic to a critical area of code. The risks of this approach are only perceived risks because the migration in question is only on master (may disrupt developers but not customers) and I think we can avoid this situation if it happens again in the future.

agrare:
So the crux of the problem is adding a column and removing it in two different migrations in the same release?  Could we add a spec for this to prevent this?

bdunne:
In this case, the remote region runs the first migration which adds the new column (generating changes for all existing records).  Replication gets stuck here because the global DB doesn't have the new column yet.  Then the remote runs the second migration.  Once the global runs the first migration to add the column, replication can be manually restarted and get unstuck (but we don't currently do this in any automated way).  So, in this case, the second migration ran on the global before all of those changes from the remote region had been replicated and at this point they will be stuck forever and the only resolution is to delete all data for the remote region from the global database and re-replicate everything.

agrare:
@bdunne what if we left the old columns but kept the data migration and added a deprecation warning, then after the release we delete the old columns?  That way it should allow the migration on the global/remove regions to run independently.

We could still run into an issue if you're upgrading across multiple versions but its simpler than patching `#migrate` and cleaner than editing old migrations.

carbonin:
Seeing as we now have three places where an issue like this has happened, I'm going to close this in favor of the migrate patch approach.

PR #266, opened 2018-08-28
Add a "migrations_ran" column to miq_regions

carbonin:
This will allow the global to track the changes generated by each
migration as they are replicated.

https://bugzilla.redhat.com/show_bug.cgi?id=1601015

Dependent of: https://github.com/ManageIQ/manageiq/pull/17919
Alternative to: https://github.com/ManageIQ/manageiq-schema/pull/264


db/migrate/20180817133610_add_migrations_ran_to_miq_regions.rb
```diff
@@ -0,0 +1,18 @@
...
+  class MiqRegion < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    add_column :miq_regions, :migrations_ran, :string, :array => true
+    MiqRegion.reset_column_information
+    r = MiqRegion.in_my_region.first
+    return unless r
+    r.migrations_ran = ActiveRecord::SchemaMigration.normalized_versions << version
```
bdunne: `update_attributes!` instead?

bdunne: Also, I don't think this version should be added yet, should it?

carbonin:
I'm fairly sure I will need to fudge the migration timestamp here to fix the original issue.

With that in mind, I'm thinking of just making this the first migration in the `hammer` release. Thoughts @Fryguy @bdunne @gtanzillo ?

carbonin:
So to add some real info to my previous question:

- The earliest migration timestamp in master which isn't in Gaprindashvili is `20171031200739`
- The latest migration in Gaprindashvili is `20180118131417` - just for reference, I don't think this is a barrier we would consider for this case.
- The switches migration which adds the column that caused this bug is `20180126144529`

So as I see it our options are the following:
1. Super super early (after collapsed migration)
2. Just before `20171031200739` (first migration in Hammer)
3. Just before `20180126144529` (just fix the bug)

I think I'd prefer option 2. I think option 1 is a little extreme and won't let us do another collapsed migration (probably ever) and option 3 leaves us more open to someone having some very early migration timestamp in a branch that could cause the same issue we're trying to solve.

For now I'll rename this migration according to option 2 and un-wip. Still definitely open for discussion on this point though,

bdunne:
I'm fine with option 2.

carbonin:
@Fryguy @gtanzillo any objections here?

PR #267, opened 2018-08-29
Add table for service instances

Ladas:
Add table for service instances. We need the service instances
modeled to be able to do service retirement.


Ladas:
@miq-bot assign @agrare 

Ladas:
@gmcculloug @gtanzillo @agrare so we've talked with @syncrou and without this modeled, we are not able to retire the services.

gmcculloug:
@Ladas Totally understand, the current effort we are working on is purely focused on ordering.  While we do not have an immediate need for this data I suspect we (or someone) will in the future.  Likely, the near future.

Ladas:
so, do we have better naming than service_instance?

agrare:
@syncrou what do you think of the name?

Ladas:
@bdunne can you add the API also for ```service_instances``` ?

PR #268, opened 2018-08-30
Add parent physical chassis id to physical chassis

tadeboro:
This additional field is used to model chassis nesting. Example would be installation that uses Intel's Rack Scale Design (pod containing multiple racks containing multiple blocks that in turn contain multiple sleds ...).


tadeboro:
@miq-bot assign @agrare

spec/migrations/20180828092111_add_parent_physical_chassis_id_to_physical_chassis_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_migration
```
lpichler: we don't need specs for structural migrations, they are tested by rails's specs, we need it only for migration with data manipulation, so you can remove it, thanks!

tadeboro:
@lpichler Thanks for your input. I removed the unnecessary tests as suggested.

PR #270, opened 2018-09-05
Remove column report_base_model from ChargebackRate table

hstastna:
**Related to:** 
https://github.com/ManageIQ/manageiq-schema/pull/209
https://bugzilla.redhat.com/show_bug.cgi?id=1581817

**What:**
We need to store the type of the Rate which is related to type of Report: Chargeback for VMs/Images/Projects, but not in `ChargebackRate` table. This was the first idea presented in https://github.com/ManageIQ/manageiq-schema/pull/209 but we decided that storing report type in `ChargeableField` is the best option (another migration will be needed).

**Details:**
We need to store the type of the Rate for adding new drop down to editing screen for Rate to be able to choose the type of the rate and then to display only appropriate columns of the rate, based on the chosen type of the rate (VMs, Images, Projects).

### DIFF
--- a/db/migrate/20180905144610_remove_report_base_model_from_chargeback_rate.rb
+++ b/db/migrate/20180905144610_remove_report_base_model_from_chargeback_rate.rb
@@ -1,0 +1,5 @@
+class RemoveReportBaseModelFromChargebackRate < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :chargeback_rates, :report_base_model, :string
+  end
+end


hstastna:
ping @lpichler, @Fryguy, @gtanzillo ;)

carbonin:
There's nothing that depends on this column yet, right? Do we want to also add the "right" solution to this PR?

hstastna:
@carbonin Right and yes, we want the "right" solution but I need to discuss it with @lpichler as I was told that it may not be as simple. So I would better create a new PR for the right solution so we don't need to wait for it here.

PR #272, opened 2018-09-17
Add limits colums to conversion_hosts table

fdupont-redhat:
This adds colums to model V2V throttling limits at conversion host level.

### DIFF
--- a/db/migrate/20180917151300_add_limits_to_conversion_host.rb
+++ b/db/migrate/20180917151300_add_limits_to_conversion_host.rb
@@ -1,0 +1,9 @@
+class AddLimitsToConversionHost < ActiveRecord::Migration[5.0]
+  def change
+    add_column :conversion_hosts, :concurrent_transformation_limit, :string
+    add_column :conversion_hosts, :cpu_limit, :string
+    add_column :conversion_hosts, :memory_limit, :string
+    add_column :conversion_hosts, :network_limit, :string
+    add_column :conversion_hosts, :blockio_limit, :string
+  end
+end


fdupont-redhat:
@miq-bot add-label gaprindashvili/no, schema, transformation
@miq-bot add-reviewer agrare

db/migrate/20180917082341_add_conversion_host_limits.rb
```diff
@@ -0,0 +1,11 @@
+class AddConversionHostLimitsColumns < ActiveRecord::Migration[5.0]
```
agrare: This class doesn't match the name of the file, did you create this with `rails generate migration AddConversionHostLimits` or create the file manually?

db/migrate/20180917082341_add_conversion_host_limits.rb
```diff
@@ -0,0 +1,11 @@
...
+  def change
+    change_table :conversion_hosts do |t|
+      t.string :limits_max_concurrent_transformations
```
carbonin: I'm not sure about the naming and plualization here, also `limit` + `max` seems redundant. Wouldn't this be clearer as `concurrent_transformation_limit`?

db/migrate/20180917082341_add_conversion_host_limits.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :limits_max_cpu
```
carbonin: Same for these, I think they would read better as `<metric>_limit` or `max_used_<metric>`.

Also, just for my own curiosity, are these metrics on the conversion host or limits on the VMs this host can convert?

db/migrate/20180917082341_add_conversion_host_limits.rb
fdupont-redhat: I've redone the migration using rails generator. Now tests are green.

agrare: :+1: recommend squashing those commits since the first doesn't add any value to the history

fdupont-redhat:
@agrare Commit squashed.

PR #274, opened 2018-09-18
Rename cancellation status

d-m-u:
It'd be good if this were spelled right. 

Merging this will break the back-end, I was hoping @hsong-rh could help fix when that becomes an issue because it was his original PR: https://github.com/ManageIQ/manageiq-schema/pull/254

### DIFF
--- a/db/migrate/20180918130338_rename_cancelation_status_in_miq_request.rb
+++ b/db/migrate/20180918130338_rename_cancelation_status_in_miq_request.rb
@@ -1,0 +1,6 @@
+class RenameCancelationStatusInMiqRequest < ActiveRecord::Migration[5.0]
+  def change
+    rename_column :miq_requests,      :cancelation_status, :cancellation_status
+    rename_column :miq_request_tasks, :cancelation_status, :cancellation_status
+  end
+end


d-m-u:
I don't even know. 
@miq-bot assign @Fryguy 
... cause why not.

carbonin:
Looks like both spellings are acceptable and ["Canceled (with one “L”) is the preferred choice in American English"](https://writingexplained.org/canceled-or-cancelled-difference)

If we're splitting hairs another quote seems to say that this particular word is a special case where two L's are preferred, but I'd prefer to not add an additional migration over something that isn't actually an incorrect spelling.

> Likewise, The AP Stylebook prefers the use of cancel, canceled, and canceling, but it favors cancellation over cancelation

Here are some references:
https://en.wiktionary.org/wiki/cancelation
https://writingexplained.org/canceled-or-cancelled-difference

@Fryguy @bdunne I recall actually having this conversation with at least one of you. Thoughts?

Fryguy:
both spellings are acceptable, but I personally would prefer two L's.  @hsong-rh, is there a lot of code targets to change?

Fryguy:
> It'd be good if this were spelled right.

You mean "spelt"? 🇬🇧 :joy:

(cc @juliancheal)

juliancheal:
@Fryguy spelt is a type of flour. 🍞

bdunne:
I like the single 'l' spelling for consistency (most of the product is already using American English), but if everyone else has a problem with that... ¯\\\_(ツ)_/¯

jrafanie:
I'd prefer one L to keep it 🇺🇸 English and dump more 🍵 in the harbor. 

Just kidding, I don't care.  I just wanted to get a reaction from @juliancheal.  I don't think this is a typo if it's acceptable 🇺🇸 English or 🇬🇧 English.  I honestly didn't realize it was one L for 🇬🇧  and two for 🇺🇸.  

I wouldn't want to have a deal with a migration for this situation.  Now, if someone tries to add a `colour` or `initialised_on` column, then 😡 

🤣

juliancheal:
> I honestly didn't realize it was one L for 🇺🇸 and two for 🇬🇧.

Realise.

hsong-rh:
I just noticed this PR. Actually when I named them, I did the same research on web and on purpose chose single L for `canceled` and `canceling`, double L for `cancellation`. I am ok with both of them. If you want to change, I can change. 😀

d-m-u:
Reallise.

AllenBW:
🍿

AllenBW:
no quibbling over the spelling of emojis 😏

d-m-u:
Okay, **fine**.

PR #275, opened 2018-09-20
Maintenance zone support for suspending provider

slemrmartin:
**BZ**: https://bugzilla.redhat.com/show_bug.cgi?id=1455145
**Issue**: https://github.com/ManageIQ/manageiq/issues/17489

Suspending provider consists of moving provider (`ExtManagementSystem`) to maintenance zone.

Maintenance zone will be created in `Zone.seed` and identified by association to region.
Name of zone will be generated as `maintenance_#{uuid}`. This zone will have `visible` flag set to false, which will be handled by PRs in UI and API.

---

This is an alternative to https://github.com/ManageIQ/manageiq-schema/pull/222, where zone is supposed to be identified by unique name.

- [ ] (**dependent**) https://github.com/ManageIQ/manageiq/pull/17452

### DIFF
--- a/spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
+++ b/spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
@@ -1,0 +1,5 @@
+class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true
+  end
+end
--- a/spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
+++ b/spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
@@ -1,5 +1,0 @@
-class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
-  def change
-    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true
-  end
-end
--- a/db/migrate/20180618083035_add_visible_to_zone.rb
+++ b/db/migrate/20180618083035_add_visible_to_zone.rb
@@ -1,0 +1,5 @@
+class AddVisibleToZone < ActiveRecord::Migration[5.0]
+  def change
+    add_column :zones, :visible, :boolean
+  end
+end
--- a/db/migrate/20180618084054_init_zones_visibility.rb
+++ b/db/migrate/20180618084054_init_zones_visibility.rb
@@ -1,0 +1,16 @@
+class InitZonesVisibility < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Updating all zones to visible") do
+      Zone.update_all(:visible => true)
+    end
+  end
+
+  def down
+    say_with_time("Resetting zone visibility") do
+      Zone.update_all(:visible => nil)
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


slemrmartin:
cc @agrare @Fryguy @bdunne @Ladas @carbonin

Ladas:
@bdunne I believe this is the last day we can merge it? So can you? (btw. I believe it was already commented that we should not mix data and schema migrations together)

agrare:
@slemrmartin what does the corresponding core PR look like if the maintenance zone name isn't constant?
e.g. here https://github.com/ManageIQ/manageiq/pull/17452/files#diff-d8c450b508868941c2b0c5884e718324R129 where all we know is the maintenance zone name.
Are you going to have to do some query for `Zone.where(:visible => :false)` and cache that name?

bdunne:
We should probably implement a `Zone.maintenance_zone` method that caches on the class

spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
```diff
@@ -0,0 +1,5 @@
+class AddBackupZoneIdToExtManagementSystem < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :ext_management_systems, :backup_zone, :type => :bigint, :index => true
```
carbonin: I know this was here from the original PR, but I'm not a fan of the name of this column.

`backup` makes it sound like it will get used if there is a problem rather than as a part of normal operation (i.e. resume).

I haven't been able to come up with anything I'm really happy with, but here are some of my thoughts:

- `runtime_zone`
- `zone_before_pause`

Any other ideas? @agrare @Fryguy ?

carbonin:
Generally, I'm with @bdunne. This looks like a much better approach than a rename. I'll :+1: if we can come up with a better name for the `backup_zone` column.

slemrmartin:
> @slemrmartin what does the corresponding core PR look like if the maintenance zone name isn't constant?
> e.g. here https://github.com/ManageIQ/manageiq/pull/17452/files#diff-d8c450b508868941c2b0c5884e718324R129 where all we know is the maintenance zone name.
> Are you going to have to do some query for `Zone.where(:visible => :false)` and cache that name?

@agrare there can be used `Zone.maintenance_zone&.name` (which will be changed to find zone by region.maintenance_zone).
Flag "visible" will be used by UI and API and can be more used for hide it from user's eyes.

slemrmartin:
> We should probably implement a `Zone.maintenance_zone` method that caches on the class
It's is implemented in dependent PR, only has to be changed according to this PR

spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
slemrmartin: `original_zone`?

agrare:
@slemrmartin @bdunne :+1:

spec/migrations/20180618084757_add_backup_zone_id_to_ext_management_system.rb
carbonin: I also considered `original_zone` or `last_zone`, but really this is something specific to pausing an ems so I would rather make that obvious from the attribute name.

Spoke about it for a bit with @jrafanie, our "Master Namer of Things" and we came up with `zone_when_paused`. How about that?

jrafanie: Naming is hard

agrare: Don't want to bikeshed too much about the name, but `zone_when_paused` would be the maintenance zone, I'd think `zone_before_pause` would be better than `zone_when_paused`

agrare:
LGTM, @carbonin/@bdunne thoughts?

PR #278, opened 2018-09-26
Fix incorrectly tenant default groups

kbrock:
When creating a tenant, if there are already groups in the database,
it would assign a random group.
This would manifest by having a seemingly random group assigned to newly
discovered VMs/Hosts. There can then be privilege issues as the admins
possibly don't have access to the group.

This only happens for customers who had a database before tenancy
was introduced in 2015. It incorrectly assigned during the migration
process in 20151021174140_assign_tenant_default_group

This also will probably only happen for the default tenant.

The fix: When a tenant points to a group that is not a tenant group,
change the default group to point to a new tenant group.
The old group was an incorrect group.

https://bugzilla.redhat.com/show_bug.cgi?id=1625788

I also changed the MiqGroup code to choose the correct group.

see also https://github.com/ManageIQ/manageiq/pull/18025

@miq-bot add_label bug hammer/yes
@miq-bot assign @Fryguy

### DIFF
--- a/db/migrate/20180926152238_fix_default_tenant_group.rb
+++ b/db/migrate/20180926152238_fix_default_tenant_group.rb
@@ -1,0 +1,57 @@
+class FixDefaultTenantGroup < ActiveRecord::Migration[5.0]
+  class Tenant < ActiveRecord::Base
+    belongs_to :default_miq_group, :class_name => "::FixDefaultTenantGroup::MiqGroup"
+
+    def add_default_miq_group
+      tenant_group = ::FixDefaultTenantGroup::MiqGroup.create_tenant_group(self)
+      update_attributes!(:default_miq_group_id => tenant_group.id)
+    end
+
+    def root?
+      ancestry.nil?
+    end
+  end
+
+  class MiqUserRole < ActiveRecord::Base
+    DEFAULT_TENANT_ROLE_NAME = "EvmRole-tenant_administrator".freeze
+
+    # if there is no role, that is ok
+    # MiqGroup.seed will populate
+
+    def self.default_tenant_role
+      @default_role ||= find_by(:name => DEFAULT_TENANT_ROLE_NAME)
+    end
+  end
+
+  class MiqGroup < ActiveRecord::Base
+    TENANT_GROUP = "tenant".freeze
+    USER_GROUP   = "user".freeze
+
+    def tenant_group?
+      group_type == TENANT_GROUP
+    end
+
+    def self.create_tenant_group(tenant)
+      role = ::FixDefaultTenantGroup::MiqUserRole.default_tenant_role
+      create_with(
+        :description      => "Tenant #{tenant.name} #{tenant.id} access",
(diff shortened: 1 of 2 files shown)


spec/migrations/20180926152238_fix_default_tenant_group_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      it "creates a new group" do
+        g = group_stub.create!(:group_type => group_stub::USER_GROUP)
+        t = tenant_stub.create!(:default_miq_group_id => g.id)
+        expect(t.default_miq_group).not_to be_tenant_group
+
+        migrate
+
+        t.reload
+        expect(t.default_miq_group_id).to be_truthy
```
bdunne: This feels like an unnecessary expectation since you're going to use that column in the next expectation

db/migrate/20180926152238_fix_default_tenant_group.rb
```diff
@@ -0,0 +1,56 @@
+class FixDefaultTenantGroup < ActiveRecord::Migration[5.0]
+  class Tenant < ActiveRecord::Base
+    belongs_to :default_miq_group, :class_name => "::FixDefaultTenantGroup::MiqGroup", :dependent => :destroy
```
bdunne: I don't see the `:dependent => :destroy` used anywhere in the migration

kbrock:
thnx @bdunne Made those changes

db/migrate/20180926152238_fix_default_tenant_group.rb
```diff
@@ -0,0 +1,56 @@
...
+      ).find_or_create_by!(
+        :tenant_id  => tenant.id,
+        :group_type => TENANT_GROUP,
+      )
+    end
+  end
+
+  def up
+    say_with_time "adding default tenant groups" do
+      Tenant.all.reject { |t| t.default_miq_group&.tenant_group? }.each do |t|
```
bdunne: Should we avoid looping through the list twice with something like this?
```ruby
Tenant.find_each do |t|
  next if t.default_miq_group&.tenant_group?
  t.default_miq_group_id = nil
  t.add_default_miq_group
end
```

kbrock: thnx, I'll fix this

kbrock:
@miq-bot add_label Hammer/yes
This is fixing data for all customers who use manageiq since before 2016

Thanks for the code review @bdunne  - made that last change

simaishi:
Hammer backport details:

```
$ git log -1
commit 4b5752c00b8f6b3853800973d90f5ba5c6b7c11f
Date:   Fri Sep 28 16:42:15 2018 -0400

    Merge pull request #278 from kbrock/bz_1625788
    
    Fix incorrectly tenant default groups
    
    (cherry picked from commit 46cc3f254368ccab8a6ec68b99188b3fffd322ab)
```

PR #281, opened 2018-10-01
Add conversion_host_id to miq_request_task

fdupont-redhat:
The `ServiceTemplateTransformationPlanTask` class relies on a conversion host to actually run the migration. The conversion host is modeled in `ConversionHost` class. From a database point of view, the `ServiceTemplateTransformationPlanTask` class is stored in `miq_request_tasks` table, because it's a sub-sub-class of `MiqRequestTask`. This PR adds a `conversion_host_id` column to `miq_request_tasks` to set the relationship.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1634029

### DIFF
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -34,7 +34,7 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
 
   def down
     conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
-      return if task.conversion_host.nil?
+      next if task.conversion_host.nil?
       task.options[:transformation_host_id] = task.conversion_host.resource.id
       task.save!
       task.conversion_host.id
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -21,13 +21,14 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
       host_id = task.options[:transformation_host_id]
       next unless host_id
       host = Host.find_by(:id => host_id)
-      next if host.nil?
-      task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
-        ch.name                     = host.name
-        ch.vddk_transport_supported = true
-        ch.ssh_transport_supported  = false
-      end
       task.options.delete(:transformation_host_id)
+      if host.present?
+        task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
+          ch.name                     = host.name
+          ch.vddk_transport_supported = true
+          ch.ssh_transport_supported  = false
+        end
+      end
       task.save!
     end
   end
@@ -38,7 +39,7 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
       task.save!
       task.conversion_host.id
     end
-    ConversionHost.destroy(conversion_host_ids)
+    ConversionHost.destroy(conversion_host_ids.compact)
 
     remove_column :miq_request_tasks, :conversion_host_id
   end
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -1,0 +1,48 @@
+class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
+  class MiqRequestTask < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    serialize :options, Hash
+    belongs_to :conversion_host, :class_name => "AddConversionHostIdToMiqRequestTasks::ConversionHost"
+  end
+
+  class ServiceTemplateTransformationPlanTask < MiqRequestTask
+  end
+
+  class ConversionHost < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :service_template_transformation_plan_tasks, :class_name => "AddConversionHostIdToMiqRequestTasks::MiqRequestTask"
+  end
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :tags, :class_name => "AddConversionHostIdToMiqRequestTasks::Tag"
+  end
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    ServiceTemplateTransformationPlanTask.all.reject { |task| task.options[:transformation_host_id].nil? }.each do |task|
+      host = Host.find_by(:id => task.options[:transformation_host_id])
+      task.conversion_host = ConversionHost.where(:id => task.options[:transformation_host_id]).first_or_create do |ch|
+        ch.name                     = host.name
+        ch.resource_type            = host.type
+        ch.resource_id              = host.id
+        ch.address                  = host.ipaddress
+        ch.vddk_transport_supported = true
+        ch.ssh_transport_supported  = false
+      end
+      task.options.delete(:transformation_host_id)
+      task.save!
+    end
+  end
+
+  def down
+    ServiceTemplateTransformationPlanTask.all.select { |task| task.conversion_host.present? }.each do |task|
+      task.options[:transformation_host_id] = task.conversion_host.id
+      ConversionHost.find(task.conversion_host.id).destroy!
+      task.save!
+    end
+
+    remove_column :miq_request_tasks, :conversion_host_id
+  end
+end
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -18,10 +18,9 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
     add_column :miq_request_tasks, :conversion_host_id, :bigint
 
     MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
-      host_id = task.options[:transformation_host_id]
+      host_id = task.options.delete(:transformation_host_id)
       next unless host_id
       host = Host.find_by(:id => host_id)
-      task.options.delete(:transformation_host_id)
       if host.present?
         task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
           ch.name                     = host.name
@@ -35,11 +34,12 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
 
   def down
     conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      return if task.conversion_host.nil?
       task.options[:transformation_host_id] = task.conversion_host.resource.id
       task.save!
       task.conversion_host.id
-    end
-    ConversionHost.destroy(conversion_host_ids.compact)
+    end.uniq.compact
+    ConversionHost.destroy(conversion_host_ids)
 
     remove_column :miq_request_tasks, :conversion_host_id
   end
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -5,29 +5,24 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
     belongs_to :conversion_host, :class_name => "AddConversionHostIdToMiqRequestTasks::ConversionHost"
   end
 
-  class ServiceTemplateTransformationPlanTask < MiqRequestTask
-  end
-
   class ConversionHost < ActiveRecord::Base
     self.inheritance_column = :_type_disabled
-    has_many :service_template_transformation_plan_tasks, :class_name => "AddConversionHostIdToMiqRequestTasks::MiqRequestTask"
   end
 
   class Host < ActiveRecord::Base
     self.inheritance_column = :_type_disabled
-    has_many :tags, :class_name => "AddConversionHostIdToMiqRequestTasks::Tag"
   end
 
   def up
     add_column :miq_request_tasks, :conversion_host_id, :bigint
 
-    ServiceTemplateTransformationPlanTask.all.reject { |task| task.options[:transformation_host_id].nil? }.each do |task|
-      host = Host.find_by(:id => task.options[:transformation_host_id])
-      task.conversion_host = ConversionHost.where(:id => task.options[:transformation_host_id]).first_or_create do |ch|
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
+      host_id = task.options[:transformation_host_id]
+      next unless host_id
+      host = Host.find_by(:id => host_id)
+      next if host.nil?
+      task.conversion_host = ConversionHost.find_or_create_by!(:resource_id => host.id, :resource_type => host.type) do |ch|
         ch.name                     = host.name
-        ch.resource_type            = host.type
-        ch.resource_id              = host.id
-        ch.address                  = host.ipaddress
         ch.vddk_transport_supported = true
         ch.ssh_transport_supported  = false
       end
@@ -37,8 +32,9 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
   end
 
   def down
-    ServiceTemplateTransformationPlanTask.all.select { |task| task.conversion_host.present? }.each do |task|
-      task.options[:transformation_host_id] = task.conversion_host.id
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
+      next unless task.conversion_host
+      task.options[:transformation_host_id] = task.conversion_host.resource_id
       ConversionHost.find(task.conversion_host.id).destroy!
       task.save!
     end
--- a/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
+++ b/db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
@@ -7,6 +7,7 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
 
   class ConversionHost < ActiveRecord::Base
     self.inheritance_column = :_type_disabled
+    belongs_to :resource, :polymorphic => true
   end
 
   class Host < ActiveRecord::Base
@@ -21,7 +22,7 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
       next unless host_id
       host = Host.find_by(:id => host_id)
       next if host.nil?
-      task.conversion_host = ConversionHost.find_or_create_by!(:resource_id => host.id, :resource_type => host.type) do |ch|
+      task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
         ch.name                     = host.name
         ch.vddk_transport_supported = true
         ch.ssh_transport_supported  = false
@@ -32,12 +33,12 @@ class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
   end
 
   def down
-    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
-      next unless task.conversion_host
-      task.options[:transformation_host_id] = task.conversion_host.resource_id
-      ConversionHost.find(task.conversion_host.id).destroy!
+    conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      task.options[:transformation_host_id] = task.conversion_host.resource.id
       task.save!
+      task.conversion_host.id
     end
+    ConversionHost.destroy(conversion_host_ids)
 
     remove_column :miq_request_tasks, :conversion_host_id
   end
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -6,7 +6,7 @@ describe AddConversionHostIdToMiqRequestTasks do
   let(:conversion_host_stub) { migration_stub(:ConversionHost) }
 
   migration_context :up do
-    it "when host doesn't exist" do
+    it "doesn't set the conversion host when the host doesn't exists" do
       host = host_stub.create!
       conversion_host_id = host.id
       task = task_stub.create!(
@@ -19,11 +19,11 @@ describe AddConversionHostIdToMiqRequestTasks do
       task.reload
 
       expect(task.options).to eq(:dummy_key => 'dummy_value')
-      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => conversion_host_id)).to be_nil
+      expect(conversion_host_stub.find_by(:resource_id => conversion_host_id)).to be_nil
       expect(task.conversion_host).to be_nil
     end
 
-    it "when host exist" do
+    it "creates a conversion host and updates the task when the host exists" do
       host = host_stub.create!
       task = task_stub.create!(
         :type    => 'ServiceTemplateTransformationPlanTask',
@@ -34,8 +34,8 @@ describe AddConversionHostIdToMiqRequestTasks do
       task.reload
 
       expect(task.options).to eq(:dummy_key => 'dummy_value')
-      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host)).not_to be_nil
-      expect(task.conversion_host).to eq(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host))
+      expect(conversion_host_stub.find_by(:resource => host)).not_to be_nil
+      expect(task.conversion_host).to eq(conversion_host_stub.find_by(:resource => host))
     end
   end
 
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -4,29 +4,31 @@ describe AddConversionHostIdToMiqRequestTasks do
   let(:task_stub) { migration_stub(:MiqRequestTask) }
   let(:host_stub) { migration_stub(:Host) }
   let(:conversion_host_stub) { migration_stub(:ConversionHost) }
-  let(:host_id) { 42 }
 
   migration_context :up do
     it "when host doesn't exist" do
+      host = host_stub.create!
+      conversion_host_id = host.id
       task = task_stub.create!(
         :type    => 'ServiceTemplateTransformationPlanTask',
-        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => conversion_host_id }
       )
+      host.destroy!
 
       migrate
       task.reload
 
-      expect(task.options).to eq({ :dummy_key => 'dummy_value', :transformation_host_id => host_id })
-      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host_id)).to be_nil
+      expect(task.options).to eq({ :dummy_key => 'dummy_value' })
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => conversion_host_id)).to be_nil
       expect(task.conversion_host).to be_nil
     end
 
     it "when host exist" do
+      host = host_stub.create!
       task = task_stub.create!(
         :type    => 'ServiceTemplateTransformationPlanTask',
-        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host.id }
       )
-      host = host_stub.create!(:id => host_id)
 
       migrate
       task.reload
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -18,7 +18,7 @@ describe AddConversionHostIdToMiqRequestTasks do
       migrate
       task.reload
 
-      expect(task.options).to eq({ :dummy_key => 'dummy_value' })
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
       expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => conversion_host_id)).to be_nil
       expect(task.conversion_host).to be_nil
     end
@@ -33,7 +33,7 @@ describe AddConversionHostIdToMiqRequestTasks do
       migrate
       task.reload
 
-      expect(task.options).to eq({ :dummy_key => 'dummy_value' })
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
       expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host)).not_to be_nil
       expect(task.conversion_host).to eq(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host))
     end
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -4,19 +4,34 @@ describe AddConversionHostIdToMiqRequestTasks do
   let(:task_stub) { migration_stub(:MiqRequestTask) }
   let(:host_stub) { migration_stub(:Host) }
   let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+  let(:host_id) { 42 }
 
   migration_context :up do
-    it "creates conversion host" do
-      host = host_stub.create!
+    it "when host doesn't exist" do
       task = task_stub.create!(
         :type    => 'ServiceTemplateTransformationPlanTask',
-        :options => { :transformation_host_id => host.id }
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
       )
 
       migrate
-
       task.reload
-      expect(task.options).to eq({})
+
+      expect(task.options).to eq({ :dummy_key => 'dummy_value', :transformation_host_id => host_id })
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host_id)).to be_nil
+      expect(task.conversion_host).to be_nil
+    end
+
+    it "when host exist" do
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
+      )
+      host = host_stub.create!(:id => host_id)
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq({ :dummy_key => 'dummy_value' })
       expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host)).not_to be_nil
       expect(task.conversion_host).to eq(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host))
     end
@@ -25,9 +40,7 @@ describe AddConversionHostIdToMiqRequestTasks do
   migration_context :down do
     it "updates task.options" do
       host = host_stub.create!
-      conversion_host = conversion_host_stub.create!(
-        :resource => host
-      )
+      conversion_host = conversion_host_stub.create!(:resource => host)
       task = task_stub.create!(
         :type            => 'ServiceTemplateTransformationPlanTask',
         :conversion_host => conversion_host
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -7,40 +7,37 @@ describe AddConversionHostIdToMiqRequestTasks do
 
   migration_context :up do
     it "creates conversion host" do
-      task = task_stub.create!
       host = host_stub.create!
-      conversion_host = conversion_host_stub.create!(
-        :resource_id   => host.id,
-        :resource_type => host.type
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :transformation_host_id => host.id }
       )
-      task.options = { :transformation_host_id => conversion_host.id }
-      task.save
 
       migrate
 
       task.reload
       expect(task.options).to eq({})
-      expect(task.conversion_host).to eq(conversion_host)
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host.id)).not_to be_nil
+      expect(task.conversion_host).to eq(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host.id))
     end
   end
 
   migration_context :down do
     it "updates task.options" do
-      task = task_stub.create!
       host = host_stub.create!
       conversion_host = conversion_host_stub.create!(
         :resource_id   => host.id,
         :resource_type => host.type
       )
-      conversion_host.save
-      task.conversion_host = conversion_host
-      task.save
+      task = task_stub.create!(
+        :type            => 'ServiceTemplateTransformationPlanTask',
+        :conversion_host => conversion_host
+      )
 
       migrate
 
       task.reload
-      expect(task.attributes).not_to include('conversion_host_id')
-      expect(task.options[:transformation_host_id]).to eq(conversion_host.id)
+      expect(task.options[:transformation_host_id]).to eq(host.id)
       expect { conversion_host.reload }.to raise_error(ActiveRecord::RecordNotFound)
     end
   end
--- a/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
+++ b/spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
@@ -1,0 +1,47 @@
+require_migration
+
+describe AddConversionHostIdToMiqRequestTasks do
+  let(:task_stub) { migration_stub(:MiqRequestTask) }
+  let(:host_stub) { migration_stub(:Host) }
+  let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+
+  migration_context :up do
+    it "creates conversion host" do
+      task = task_stub.create!
+      host = host_stub.create!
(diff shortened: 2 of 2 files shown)


carbonin:
It looks like this information is currently being stored in the `ServiceTemplateTransformationPlanTask#options` column. I think this PR should also include a data migration to remove that value and put it in the new column for all previous tasks.

fdupont-redhat:
@carbonin That's even more complex. Today, the information stored in options[:transformation_host_id] is the id of a host (class Host). The PR stores the conversion host (class ConversionHost) id. A conversion host has a `resource` which is either a Host or a Vm, stored as `resource_type`. So current migration should create a ConversionHost with a `resource` relationship to a `resource_type` object with id `resource_id`, if it doesn't exist. Then the `conversion_host_id` can be set.

@agrare what are the possible values for `resource_type` ? `Host` and `Vm` ?

agrare:
@fdupont-redhat if `options[:transformation_host_id]` can only ever contain a `Host` object then this should create a `ConversionHost` record for every distinct `options[:transformation_host_id]` and then `conversion_host.resource = Host.find(options[:transformation_host_id])`.

It is a little more complicated than just `Host` or `Vm` because the hosts will be STI'd e.g. `ManageIQ::Providers::Redhat::InfraManager::Host`.

fdupont-redhat:
@miq-bot add-label gaprindashvili/no, schema, transformation

fdupont-redhat:
@miq-bot add-reviewer agrare, carbonin

fdupont-redhat:
@miq-bot add-reviewer agrare
@miq-bot add-reviewer carbonin

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :tags, :class_name => "AddConversionHostIdToMiqRequestTasks::Tag"
+  end
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    ServiceTemplateTransformationPlanTask.all.reject { |task| task.options[:transformation_host_id].nil? }.each do |task|
```
carbonin: A few things here ...
First, I don't think this query will work the way you're intending it to. STI is disabled so I think you would be better off using the base class and specifying the `type` in a `where` clause.

Also, instead of using `reject` here, can we make this a guard clause in the `each` block? That will simplify this and prevent us from iterating through the list of tasks twice.

So something like this:
```ruby
MiqRequestTask.where(:type => "ServiceTemplateTransformationPlanTask").each do |task|
  next unless task.options[:transformation_host_id]
  ...
end
```

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+        ch.vddk_transport_supported = true
+        ch.ssh_transport_supported  = false
+      end
+      task.options.delete(:transformation_host_id)
+      task.save!
+    end
+  end
+
+  def down
+    ServiceTemplateTransformationPlanTask.all.select { |task| task.conversion_host.present? }.each do |task|
```
carbonin: Same here, use the base class and you don't need to loop through the tasks twice.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  class ConversionHost < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :service_template_transformation_plan_tasks, :class_name => "AddConversionHostIdToMiqRequestTasks::MiqRequestTask"
+  end
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :tags, :class_name => "AddConversionHostIdToMiqRequestTasks::Tag"
```
carbonin: Why is this relation needed?

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+    serialize :options, Hash
+    belongs_to :conversion_host, :class_name => "AddConversionHostIdToMiqRequestTasks::ConversionHost"
+  end
+
+  class ServiceTemplateTransformationPlanTask < MiqRequestTask
+  end
+
+  class ConversionHost < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :service_template_transformation_plan_tasks, :class_name => "AddConversionHostIdToMiqRequestTasks::MiqRequestTask"
```
carbonin: This relation isn't being used in the migration, why define it here?

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
+class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
+  class MiqRequestTask < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    serialize :options, Hash
+    belongs_to :conversion_host, :class_name => "AddConversionHostIdToMiqRequestTasks::ConversionHost"
```
carbonin: Again, you're not using this relation, why define it?

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  class ServiceTemplateTransformationPlanTask < MiqRequestTask
```
carbonin: I think this can be removed ... See my other comments about STI and using the base class for querying.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :tags, :class_name => "AddConversionHostIdToMiqRequestTasks::Tag"
+  end
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    ServiceTemplateTransformationPlanTask.all.reject { |task| task.options[:transformation_host_id].nil? }.each do |task|
+      host = Host.find_by(:id => task.options[:transformation_host_id])
```
carbonin: I think this can use `find` rather than `find_by`. Also this entire block would be way easier to read if `task.options[:transformation_host_id]` was stored in a local variable (ideally before the guard clause).

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+      task.conversion_host = ConversionHost.where(:id => task.options[:transformation_host_id]).first_or_create do |ch|
```
agrare: `find_or_create_by!` would be better over `.where(:id => ...).first_or_create`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: Should probably throw in a `next if host.nil?` since you're using `host.name`, `host.type`, ... below just in case the host was deleted and the task options weren't updated which given they aren't kept in sync by refresh is a distinct possibility.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+        ch.name                     = host.name
+        ch.resource_type            = host.type
+        ch.resource_id              = host.id
```
agrare: Should be able to just do `ch.resource = host` here

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,48 @@
...
+        ch.address                  = host.ipaddress
```
agrare: If we're going to delegate ipaddress to the resource per https://github.com/ManageIQ/manageiq-schema/pull/242#issuecomment-426257391 then we should leave this blank in the conversion_host record.  That way if the host ip address changes it will be automatically picked up.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+describe AddConversionHostIdToMiqRequestTasks do
+  let(:task_stub) { migration_stub(:MiqRequestTask) }
+  let(:host_stub) { migration_stub(:Host) }
+  let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+
+  migration_context :up do
+    it "creates conversion host" do
+      task = task_stub.create!
+      host = host_stub.create!
+      conversion_host = conversion_host_stub.create!(
```
agrare: You can just do `.create!(:resource => host)` no need to split up the id/type

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        :resource_id   => host.id,
+        :resource_type => host.type
+      )
+      task.options = { :transformation_host_id => conversion_host.id }
```
agrare: You can pass `:options => {:trans....}` into the `create!()` and then you don't need the extra save

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  migration_context :down do
+    it "updates task.options" do
+      task = task_stub.create!
+      host = host_stub.create!
+      conversion_host = conversion_host_stub.create!(
+        :resource_id   => host.id,
+        :resource_type => host.type
```
agrare: Same as above https://github.com/ManageIQ/manageiq-schema/pull/281/files#r221943457

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      )
+      conversion_host.save
+      task.conversion_host = conversion_host
```
agrare: Likewise just pass `:conversion_host` in to the `.create!`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
carbonin: Ah, if that's a possibility @agrare then we should also keep the `find_by` (`find` raises if the record doesn't exist).

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      task.save
+
+      migrate
+
+      task.reload
+      expect(task.attributes).not_to include('conversion_host_id')
```
agrare: No need to check that ActiveRecord deleted the column, if that is broken we have bigger issues

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: :+1:

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
carbonin: Also, if `options[:transformation_host_id]` is a reference to the `Host` record `id` then it can't also be a reference to the `ConversionHost` `id`.

This `where` just looks wrong.

agrare: Oh yeah good catch, @fdupont-redhat did you mean to `ConversionHost.where(:id => host.conversion_host.id)` or something?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Good question @carbonin. Probably, because I'm dumb :smile:
Removed and spec updated.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Done.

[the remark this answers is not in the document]
fdupont-redhat: Fixed.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      task = task_stub.create!
+      host = host_stub.create!
+      conversion_host = conversion_host_stub.create!(
+        :resource_id   => host.id,
+        :resource_type => host.type
+      )
+      task.options = { :transformation_host_id => conversion_host.id }
+      task.save
+
+      migrate
```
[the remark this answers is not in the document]
fdupont-redhat: Should be done by updated spec.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: @agrare It fails if I use `:resource => host`. The error message is:

```ruby
ActiveModel::UnknownAttributeError:
       unknown attribute 'resource' for AddConversionHostIdToMiqRequestTasks::ConversionHost.
```

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Done.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Removed.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: You probably need to add the relation to the stubbed model

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,47 @@
...
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
+      host_id = task.options[:transformation_host_id]
+      if host_id
```
agrare: I know it is a style thing but I usually prefer `next if host_id.nil?` over `if host_id` and then everything else indented.  Reminds me of the java only one return style which brings back bad memories :smile:

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: You should check for the type also then, just checking for the id and not the type

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,44 @@
...
+      task.options.delete(:transformation_host_id)
+      task.save!
+    end
+  end
+
+  def down
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
+      next unless task.conversion_host
+      task.options[:transformation_host_id] = task.conversion_host.resource_id
+      ConversionHost.find(task.conversion_host.id).destroy!
```
agrare: You can just pass the id to destroy like `ConversionHost.destroy(task.conversion_host.id)`

For performance it might be better to `.map` the conversion host ids that you want to destroy and issue one query after the loop like `conversion_host_ids = MiqRequestTask.where().map { |task| }; ConversionHost.destroy(conversion_host_ids)`

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :transformation_host_id => host.id }
+      )
+
+      migrate
+
+      task.reload
+      expect(task.options).to eq({})
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host.id)).not_to be_nil
```
jameswnl: you should be able to do `AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host)`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
+      host_id = task.options[:transformation_host_id]
+      next unless host_id
+      host = Host.find_by(:id => host_id)
+      next if host.nil?
+      task.conversion_host = ConversionHost.find_or_create_by!(:resource_id => host.id, :resource_type => host.type) do |ch|
```
agrare: If you add the `belongs_to :resource, :polymorphic => true` to your [`ConversionHost` stub](https://github.com/ManageIQ/manageiq-schema/pull/281/files#diff-c1d34c28716698230c0bdaeb06a6aeeeR8) you should be able to do `find_or_create_by!(:resource => host)`, this will also help [here](https://github.com/ManageIQ/manageiq-schema/pull/281/files#diff-ad3d27814217293149c9a7bbe6240470R28) and [here](https://github.com/ManageIQ/manageiq-schema/pull/281/files#diff-ad3d27814217293149c9a7bbe6240470R21)

fdupont-redhat: Thanks. Helpful and good to memorize.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
fdupont-redhat: Done.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Done.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,43 @@
+require_migration
+
+describe AddConversionHostIdToMiqRequestTasks do
+  let(:task_stub) { migration_stub(:MiqRequestTask) }
+  let(:host_stub) { migration_stub(:Host) }
+  let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+
+  migration_context :up do
```
jameswnl: Can you add a case where `host` is not found?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      host = host_stub.create!
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :transformation_host_id => host.id }
+      )
+
+      migrate
+
+      task.reload
+      expect(task.options).to eq({})
```
jameswnl: can you add another dummy key in this `options` field and here check only that `:transformation_host_id` is removed?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Done.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Done.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "when host doesn't exist" do
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
+      )
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq({ :dummy_key => 'dummy_value', :transformation_host_id => host_id })
```
agrare: Hm is this right?  If it can't find the host it should still clear the transformation_host_id right?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host_id)).to be_nil
```
agrare: Don't think you need this test, you're already creating a bogus id so this just confirms that it is bogus

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,56 @@
+require_migration
+
+describe AddConversionHostIdToMiqRequestTasks do
+  let(:task_stub) { migration_stub(:MiqRequestTask) }
+  let(:host_stub) { migration_stub(:Host) }
+  let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+  let(:host_id) { 42 }
```
agrare: This looks weird, 42 is not going to be in the right region or worse if the region is 0 it might be a valid host?
How about getting the host_id from a host_stub then destroying the record.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Oh nevermind you're confirming that it doesn't create a conversion host for the bogus host, carry on

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => host_id)).to be_nil
+      expect(task.conversion_host).to be_nil
+    end
+
+    it "when host exist" do
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host_id }
+      )
+      host = host_stub.create!(:id => host_id)
```
agrare: Oh yeah and this definitely isn't going to work because the host_id is hard coded to be region 0.  I would say don't define the host_id in the `let()` and just create the host to be deleted in the above test

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
jameswnl: I don't think we need this check. Next `task.conversion_host` check of `nil` should be good enough

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,45 @@
...
+      host_id = task.options[:transformation_host_id]
+      next unless host_id
+      host = Host.find_by(:id => host_id)
+      next if host.nil?
+      task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
+        ch.name                     = host.name
+        ch.vddk_transport_supported = true
+        ch.ssh_transport_supported  = false
+      end
+      task.options.delete(:transformation_host_id)
```
carbonin: Doing the delete here means that we will not remove this option from the existing hash if we can't find the host in the existing inventory. Is that the behavior we are expecting?

If it's okay to remove the host id from the hash either way, then you can replace the right hand side of the `host_id` assignment with this line.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,45 @@
...
+      task.save!
+    end
+  end
+
+  def down
+    conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      task.options[:transformation_host_id] = task.conversion_host.resource.id
+      task.save!
+      task.conversion_host.id
```
carbonin: Will any of these ever be nil? Do we need a `compact` here?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
carbonin: I think this is what I was talking about in https://github.com/ManageIQ/manageiq-schema/pull/281/files#r222028373 @agrare

agrare: Ha yeah I was in the middle of linking this to your comment above

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
jameswnl: so if this `host` is gone from db, removing it from `options` seems better.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Should be good.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: @agrare @jameswnl decide what's better.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
fdupont-redhat: Should be good.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: @fdupont-redhat you can probably move it up [here](https://github.com/ManageIQ/manageiq-schema/pull/281/files#diff-c1d34c28716698230c0bdaeb06a6aeeeR21) and save a LOC

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  migration_context :up do
+    it "when host doesn't exist" do
+      host = host_stub.create!
+      conversion_host_id = host.id
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => conversion_host_id }
+      )
+      host.destroy!
```
agrare: :+1:

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+          ch.ssh_transport_supported  = false
+        end
+      end
+      task.save!
+    end
+  end
+
+  def down
+    conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      task.options[:transformation_host_id] = task.conversion_host.resource.id
```
agrare: If `task.conversion_host` is nil you should skip the rest of this

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+      task.save!
+      task.conversion_host.id
+    end
```
agrare: Should `.uniq` also since there will be many tasks with the same conversion host.  You can do `end.uniq.compact` or `ConversionHost.destroy(conversion_host_ids.uniq.compact)` your call

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -35,11 +34,12 @@ def up
 
   def down
     conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      return if task.conversion_host.nil?
```
agrare: Should be `next if`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+      if host.present?
+        task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
+          ch.name                     = host.name
+          ch.vddk_transport_supported = true
+          ch.ssh_transport_supported  = false
+        end
+      end
+      task.save!
+    end
+  end
```
agrare: the `#up` method looks good to me now

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
+require_migration
+
+describe AddConversionHostIdToMiqRequestTasks do
+  let(:task_stub) { migration_stub(:MiqRequestTask) }
+  let(:host_stub) { migration_stub(:Host) }
+  let(:conversion_host_stub) { migration_stub(:ConversionHost) }
+
+  migration_context :up do
+    it "when host doesn't exist" do
```
carbonin: Nit, but it would be great if this wrote out what we were testing for.

For example, `it "doesn't set the conversion host when the host object doesn't exist"`

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => conversion_host_id)).to be_nil
+      expect(task.conversion_host).to be_nil
+    end
+
+    it "when host exist" do
```
carbonin: Same here, `it "references the existing Host in the ConversionHost record"`

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => conversion_host_id }
+      )
+      host.destroy!
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource_id => conversion_host_id)).to be_nil
```
carbonin: I think this should be `conversion_host_stub.find_by...` right?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host.id }
+      )
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
+      expect(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host)).not_to be_nil
```
carbonin: Same here, I think this should use the stub.

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(task.conversion_host).to eq(AddConversionHostIdToMiqRequestTasks::ConversionHost.find_by(:resource => host))
```
carbonin: ditto

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Done

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Done

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Done

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Done

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
agrare: Done

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
+class AddConversionHostIdToMiqRequestTasks < ActiveRecord::Migration[5.0]
+  class MiqRequestTask < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    serialize :options, Hash
+    belongs_to :conversion_host, :class_name => "AddConversionHostIdToMiqRequestTasks::ConversionHost"
+  end
+
+  class ConversionHost < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    belongs_to :resource, :polymorphic => true
```
bdunne: I don't think you can use polymorphic relations with the stub models.  If you do, you'll get something like: `:resource_type => MigrationClass::StubModel` which won't be found in production.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :miq_request_tasks, :conversion_host_id, :bigint
+
+    MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').each do |task|
```
bdunne: Usually we prefer `find_each` to load records in batches.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+      host_id = task.options.delete(:transformation_host_id)
+      next unless host_id
+      host = Host.find_by(:id => host_id)
```
bdunne: `Host.find(host_id)`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+      if host.present?
+        task.conversion_host = ConversionHost.find_or_create_by!(:resource => host) do |ch|
```
bdunne: `ConversionHost.find_or_initialize_by(:resource_type => "Host", :resource_id => host.id).update_attributes!(.....)`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+      end
+      task.save!
+    end
+  end
+
+  def down
+    conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
+      next if task.conversion_host.nil?
+      task.options[:transformation_host_id] = task.conversion_host.resource.id
```
bdunne: Maybe `task.conversion_host.resource_id` to avoid loading `resource` (which I think would fail anyway)?

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  migration_context :up do
+    it "doesn't set the conversion host when the host doesn't exists" do
+      host = host_stub.create!
+      conversion_host_id = host.id
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => conversion_host_id }
+      )
+      host.destroy!
```
bdunne: You should be able to generate a record id (in the current region) without creating and destroying a record

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
bdunne: Also, this should probably be scoped with `in_my_region` to avoid running the migration on more records than necessary.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
```diff
@@ -0,0 +1,46 @@
...
+          ch.vddk_transport_supported = true
+          ch.ssh_transport_supported  = false
+        end
+      end
+      task.save!
+    end
+  end
+
+  def down
+    conversion_host_ids = MiqRequestTask.where(:type => 'ServiceTemplateTransformationPlanTask').map do |task|
```
bdunne: This should also be scoped with `in_my_region`

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      task = task_stub.create!(
+        :type    => 'ServiceTemplateTransformationPlanTask',
+        :options => { :dummy_key => 'dummy_value', :transformation_host_id => host.id }
+      )
+
+      migrate
+      task.reload
+
+      expect(task.options).to eq(:dummy_key => 'dummy_value')
+      expect(conversion_host_stub.find_by(:resource => host)).not_to be_nil
```
bdunne: shouldn't this `eq(something)`?

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
carbonin: We are expecting that this may be `nil` so `find_by` was chosen on purpose here to avoid the exception.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
carbonin: I don't see why we should use `find_or_initialize_by` + `update_attributes` over what's here, but yeah it does look like this will break because the host class will be incorrect.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
bdunne: Right above it is `next unless host_id`, so it will never be nil...

spec/migrations/20181001131632_add_conversion_host_id_to_miq_request_tasks_spec.rb
carbonin: I'm not sure how you could do that without querying the sequences. I think it would have been nice to create the host in a before block, but other than that I think this is fine.

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: If the host was deleted it will be nil.  Since this isn't a proper relation now there is not `:dependent => nullify`

db/migrate/20181001131632_add_conversion_host_id_to_miq_request_tasks.rb
agrare: Yeah you're right I thought in production it would use the type from the record but appears not

carbonin: Okay, who wants to take on changing this? Or should we write a follow up migration to alter all the broken resources to the "production ready" name?

agrare: I can, should be able to just edit this migration?

agrare: If we're going to update the `.where` to `.in_my_region.find_each` anyway

bdunne: I think Jason would say we need a new migration since developers may have already run this one...

agrare: Okay https://github.com/ManageIQ/manageiq-schema/pull/284 data migration to fix the resource_type

simaishi:
Backport approved by Engineering/QE.

Hammer backport details:

```
$ git log -1
commit 58bd0c05dd61a8f5597f9d49f1e4fedb39cb88ec
Author: Nick Carboni <ncarboni@redhat.com>
Date:   Tue Oct 2 14:02:07 2018 -0400

    Merge pull request #281 from fdupont-redhat/v2v_add_conversion_host_id_to_transformation_task
    
    Add conversion_host_id to miq_request_task
    
    (cherry picked from commit eb7e077f462b9ab5e1c652f8220a34d22e4c8633)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1634029
```

PR #282, opened 2018-10-02
Index vms also by `vm.uid_ems` field to speedup query

miha-plesko:
With this commit we do nothing but add another index to the vms because Nuage cross-connecting needs to cherry-pick vms by this specific filed, which is not indexed so far:

```
vm.uid_ems
```
Related: https://github.com/ManageIQ/manageiq-providers-nuage/pull/150
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1635232

@miq-bot add_label enhancement,hammer/yes
@miq-bot assign @agrare

/cc @Ladas


miha-plesko:
Turns out the index already exists but is hidden under name **index_vms_on_vmm_uuid**, closing. Not sure why I was able to migrate in my dev env without duplication error, though...

PR #283, opened 2018-10-02
Add Total Space column to Physical Storages

EsdrasVP:
This PR is able to add a new column called `total_space` that will be useful to store this information extracted from all `PhysicalDisks` inside the physical storage.

### DIFF
--- a/db/migrate/20181002123523_add_total_space_to_physical_storages.rb
+++ b/db/migrate/20181002123523_add_total_space_to_physical_storages.rb
@@ -1,0 +1,5 @@
+class AddTotalSpaceToPhysicalStorages < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_storages, :total_space, :bigint
+  end
+end


db/migrate/20181002123523_add_total_disk_size_to_physical_storages.rb
```diff
@@ -0,0 +1,5 @@
+class AddTotalDiskSizeToPhysicalStorages < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_storages, :total_disk_size, :string
```
bdunne: I assume this will be in bytes, correct?

bdunne: Why string?

agrare: From the provider PR (https://github.com/ManageIQ/manageiq-providers-lenovo/pull/233/files#diff-4d59c93e67bbc3a29593871858222736R5) it looks like this would be "100 GB"

:-1: on that this should be in bytes and the parser should do the conversion

EsdrasVP: @bdunne @agrare I don't think this should be in bytes. If this was a simple storage unit, that's fine, but a physical storage have slots for a big number of storage units and could have dozens of TB. If bytes is really needed for standard purposes, I change it, but otherwise I think is best to keep it at least in GB.

And regarding the type, I can change it to be an integer instead once we agree on a standard unit.

agrare: And if you have a storage array that big I'll happily take a test unit for testing, just saying ;)

EsdrasVP: @agrare My concern was to have an unreadable number due to its size. But as requested, I changed to be in bytes and be a `bigint`.

agrare: :+1: thanks @EsdrasVP the db doesn't have to be human readable as it'll be handled by the UI.  We do the same thing with the `Disk` model.

db/migrate/20181002123523_add_total_disk_size_to_physical_storages.rb
```diff
@@ -0,0 +1,5 @@
+class AddTotalDiskSizeToPhysicalStorages < ActiveRecord::Migration[5.0]
+  def change
+    add_column :physical_storages, :total_disk_size, :bigint
```
agrare: How about `:total_space` to stay consistent with `:storages`?

EsdrasVP: @agrare You're right, it's better for consistency. I changed it to be `:total_space`.

Fryguy:
@agrare @bdunne I believe we used decimal in other tables where we store total_space.  Should this have been decimal column instead of bigint? (I'm fine with bigint FWIW, just asking for consistency)

agrare:
@Fryguy was going for consistency with Storages.total_space [ref](https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20130923182042_collapsed_initial_migration.rb#L2857) which is a `:bigint`

PR #286, opened 2018-10-03
[HAMMER] Update i18n catalog for hammer

mzazrivec:


carbonin:
I don't understand what this is doing. This is an engine and doesn't have a component that should be rendered to the UI, so why translate anything?

mzazrivec:
Yep, this is an engine with ManageIQ schema, though every ManageIQ engine contains a string with its own name, which is now being rendered in UI in the about modal dialog. Hence the `"Schema"` string and this PR.

PR #288, opened 2018-10-10
Add evm_owner, tenant, group to orch stacks

d-m-u:
As part of the work for https://github.com/ManageIQ/manageiq/pull/17951#issuecomment-428575121, orchestration stacks need an owner in order to be able to retire them as a request. Since we have group and tenant ids along with evm_owners, they're here too...

TODO: Things n stuff associated with this that still need to happen: we need to get both pathways sorted -- stacks that we discover from the providers need the user set, and ones we create ourselves need the user set at the end of provisioning. 

As discussed with Madhu and GM, the approach of using the service owner and then the ems owner is the preferable way to do this.

### DIFF
--- a/db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
+++ b/db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
@@ -1,0 +1,7 @@
+class AddEvmOwnerToOrchestrationStacks < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :orchestration_stacks, :evm_owner, :type => :bigint, :index => true
+    add_reference :orchestration_stacks, :miq_group, :type => :bigint, :index => true
+    add_reference :orchestration_stacks, :tenant,    :type => :bigint, :index => true
+  end
+end
--- a/db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
+++ b/db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
@@ -1,0 +1,131 @@
+class MigrateOrchStacksToHaveOwnershipConcept < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    belongs_to :tenant, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::Tenant"
+
+    def tenant_identity
+      User.super_admin.tap { |u| u.current_group = tenant.default_miq_group }
+    end
+  end
+
+  class Service < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    belongs_to :evm_owner, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::User"
+
+    def root
+      ancestry.present? ? self.class.find(ancestry.split("/").first) : self
+    end
+    alias root_service root
+
+    def tenant_identity
+      user = evm_owner
+      if user.nil? || !user.miq_group_ids.include?(miq_group_id)
+        user = User.super_admin.tap { |u| u.current_group_id = miq_group_id }
+      end
+      user
+    end
+  end
+
+  class MiqGroup < ActiveRecord::Base
+    TENANT_GROUP = "tenant".freeze
+    self.inheritance_column = :_type_disabled # disable STI
+
+    belongs_to :tenant,     :class_name => "MigrateOrchStacksToHaveOwnershipConcept::Tenant"
+    has_one :miq_user_role, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqUserRole"
+
+    def self.create_tenant_group(tenant)
+      create_with(
+        :description         => "Tenant access",
+        :default_tenant_role => MiqUserRole.default_tenant_role
+      ).find_or_create_by!(
+        :group_type => TENANT_GROUP,
+        :tenant_id  => tenant.id,
+      )
+    end
+  end
+
+  class MiqUserRole < ActiveRecord::Base
+    DEFAULT_TENANT_ROLE_NAME = "EvmRole-tenant_administrator".freeze
+    self.inheritance_column = :_type_disabled # disable STI
+
+    def self.default_tenant_role
+      find_by(:name => DEFAULT_TENANT_ROLE_NAME)
+    end
+  end
+
+  class User < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    has_and_belongs_to_many :miq_groups, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+    belongs_to :current_group,           :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+
+    include ActiveRecord::IdRegions
+
+    def self.super_admin
+      in_my_region.find_by(:userid => "admin")
+    end
+
+    def current_tenant
+      current_group.tenant
+    end
+  end
+
+  class ServiceResource < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    belongs_to :service, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::Service"
+  end
+
+  class Tenant < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    belongs_to :default_miq_group, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup", :dependent => :destroy
+    after_create :create_tenant_group
+
+    def create_tenant_group
+      update_attributes!(:default_miq_group => MiqGroup.create_tenant_group(self)) unless default_miq_group_id
+      self
+    end
+  end
+
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    has_many :service_resources, :as => :resource, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::ServiceResource"
+    has_many :direct_services, :through => :service_resources, :source => :service
+    belongs_to :ext_management_system, :foreign_key => :ems_id
+
+    def root
+      ancestry.present? ? self.class.find(ancestry.split("/").first) : self
+    end
+
+    def direct_service
+      direct_services.first || (root.direct_services.first if root != self)
+    end
+
+    def service
+      direct_service.try(:root_service) || (root.direct_service.try(:root_service) if root != self)
+    end
+  end
+
+  def up
+    say_with_time("Migrating existing orchestration stacks to have direct owners, groups, tenant") do
+      OrchestrationStack.find_each do |stack|
+        user = if stack.service.present?
+                 stack.service.tenant_identity
+               elsif !stack.ems_id.nil?
+                 stack.ext_management_system.tenant_identity
+               else
+                 User.super_admin
+               end
+        stack.update_attributes(:evm_owner_id => user.id, :tenant_id => user.current_tenant.id, :miq_group_id => user.current_group.id)
+      end
+    end
+  end
+
+  def down
+    # blank because the down migration on AddEvmOwnerToOrchestrationStacks covers this
+  end
+end
--- a/spec/migrations/20181016140921_migrate_orch_stacks_to_have_ownership_concept_spec.rb
+++ b/spec/migrations/20181016140921_migrate_orch_stacks_to_have_ownership_concept_spec.rb
@@ -1,0 +1,56 @@
+require_migration
+
+describe MigrateOrchStacksToHaveOwnershipConcept do
+  let(:orchestration_stack) { migration_stub(:OrchestrationStack) }
+  let(:ems) { migration_stub(:ExtManagementSystem) }
+  let(:service) { migration_stub(:Service) }
+  let(:tenant) { migration_stub(:Tenant).create! }
+  let(:group) { migration_stub(:MiqGroup).create!(:tenant => tenant) }
+  let!(:user) { migration_stub(:User).create!(:userid => "admin", :miq_groups => [group], :current_group => group) }
+
+  migration_context :up do
+    it "sets owner, tenant, and group from the user if neither the ems and service exist" do
+      stack = orchestration_stack.create!(:ext_management_system => nil)
+      expect(orchestration_stack.count).to eq(1)
+
+      migrate
+      stack.reload
+
+      expect(stack).to have_attributes(:tenant_id => user.current_tenant.id, :evm_owner_id => user.id, :miq_group_id => user.current_group.id)
+    end
+
+    it "sets owner, tenant, and group from the ems if the ems exists and the service doesn't" do
+      ext_ms = ems.create!(:tenant => tenant)
+      stack = orchestration_stack.create!(:ext_management_system => ext_ms)
+      expect(orchestration_stack.count).to eq(1)
+
+      migrate
+      stack.reload
+
+      expect(stack).to have_attributes(:tenant_id => ext_ms.tenant_id, :evm_owner_id => ext_ms.tenant_identity.id, :miq_group_id => ext_ms.tenant_identity.current_group.id)
+    end
+
+    it "sets owner, tenant, and group from the service if the service exists and ems doesn't" do
+      svc = service.create!(:tenant_id => tenant.id, :miq_group_id => group.id)
+      stack = orchestration_stack.create!(:direct_services => [svc])
+      expect(orchestration_stack.count).to eq(1)
+
+      migrate
+      stack.reload
+
+      expect(stack).to have_attributes(:tenant_id => svc.tenant_id, :evm_owner_id => svc.tenant_identity.id, :miq_group_id => svc.tenant_identity.current_group.id)
+    end
+
+    it "sets owner, tenant, and group from the service if the service and ems exists" do
+      ext_ms = ems.create!(:tenant => tenant)
+      svc = service.create!(:tenant_id => tenant.id, :miq_group_id => group.id)
+      stack = orchestration_stack.create!(:direct_services => [svc], :ext_management_system => ext_ms)
+      expect(orchestration_stack.count).to eq(1)
+
+      migrate
+      stack.reload
+
+      expect(stack).to have_attributes(:tenant_id => svc.tenant_id, :evm_owner_id => svc.tenant_identity.id, :miq_group_id => svc.tenant_identity.current_group.id)
+    end
+  end
+end


d-m-u:
@miq-bot add_label wip, hammer/yes

d-m-u:
@miq-bot remove_label hammer/yes
good try though, drew

db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
```diff
@@ -0,0 +1,7 @@
+class AddEvmOwnerToOrchestrationStacks < ActiveRecord::Migration[5.0]
+  def change
+    add_column :orchestration_stacks, :evm_owner_id, :bigint
```
bdunne: Recent migrations have been using `add_reference` instead.

d-m-u:
@miq-bot add_reviewer @tinaafitz

d-m-u:
@miq-bot remove_label wip
@miq-bot assign @Fryguy

d-m-u:
@miq-bot add_reviewer @bdunne

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  class Service < ApplicationRecord
+    self.inheritance_column = :_type_disabled # disable STI
+    belongs_to :evm_owner, :class_name => "User"
+
+    def tenant_identity
+      user = evm_owner
+      user = User.super_admin if user.nil? || !user.miq_group_ids.include?(miq_group_id)
```
gmcculloug: Most services are created doing provisioning which would set their owner identify.  They can also be created in automate which allows user/group to be set as well.  This fallback to `User.super_admin` is really just a last ditch effort and not something I would expect to encounter very much.  🎃

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,83 @@
...
+      user
+    end
+  end
+
+  class MiqGroup < ApplicationRecord
+    belongs_to :tenant
```
gmcculloug: Add `self.inheritance_column = :_type_disabled` to protect against future sub-classing of this model.  Same for `ServiceResource`

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,83 @@
...
+    belongs_to :ext_management_system, :foreign_key => :ems_id
+
+    def service
+      direct_services.first
+    end
+  end
+
+  def up
+    say_with_time("Migrating existing orchestration stacks to have direct owners, groups, tenant") do
+      OrchestrationStack.select { |orch_stack| orch_stack.service.present? }.each do |stack|
```
gmcculloug: Since you are trying to set all OrchestrationStack instance it might be better to loop over all and handle the three cases.  With service, no service but ems, or neither.

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  class OrchestrationStack < ApplicationRecord
+    self.inheritance_column = :_type_disabled # disable STI
+    has_many :service_resources, :as => :resource
+    has_many :direct_services, :through => :service_resources, :source => :service
+    belongs_to :ext_management_system, :foreign_key => :ems_id
+
+    def service
+      direct_services.first
```
kbrock: The logic for this one is much more complex in orchestration_stack. Not sure if this is good enough.

(Maybe all services are owned by the ems tenant)

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,82 @@
...
+               else
+                 User.super_admin
+               end
+        stack.update_attributes(:evm_owner_id => user.id, :tenant_id => user.current_tenant.id, :miq_group_id => user.current_group.id)
+      end
+    end
+  end
+
+  def down
+    OrchestrationStack.update_all(:evm_owner_id => nil, :tenant_id => nil, :miq_group_id => nil)
```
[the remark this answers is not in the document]
jrafanie: I guess this makes sense since we're adding the columns in the prior migration so there's no value in updating columns that are about to be removed when doing "down"

spec/migrations/20181016140921_migrate_orch_stacks_to_have_ownership_concept_spec.rb
```diff
@@ -0,0 +1,63 @@
+require_migration
+
+describe MigrateOrchStacksToHaveOwnershipConcept do
+  let(:orchestration_stack) { migration_stub(:OrchestrationStack) }
+  let(:ems) { migration_stub(:ExtManagementSystem) }
+  let(:service) { migration_stub(:Service) }
+  let(:tenant) { migration_stub(:Tenant).create }
+  let(:group) { migration_stub(:MiqGroup).create(:tenant => tenant) }
+  let!(:user) { migration_stub(:User).create(:userid => "admin", :miq_groups => [group]) }
```
kbrock:
```suggestion
  let!(:user) { migration_stub(:User).create(:userid => "admin", :miq_groups => [group], :current_group => group) }
```

Alternative would be to add this to a `before_save` filter

```ruby
class User
  before_save :ensure_current_group

  def ensure_current_group
    current_group ||= miq_groups.first
  end
```

d-m-u:
I'd just like to point out that GM didn't get that reference either, but, anyway, thanks for the help, @kbrock

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  class MiqGroup < ApplicationRecord
+    self.inheritance_column = :_type_disabled # disable STI
+    belongs_to :tenant, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::Tenant"
+  end
+
+  class User < ApplicationRecord
+    has_and_belongs_to_many :miq_groups, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+    belongs_to :current_group, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+    self.inheritance_column = :_type_disabled # disable STI
```
jrafanie: I'm speaking up for the automated Jason when I say, can you align the class_name and make it consistent with the order of associations and the inheritance colum?  Meaning, inheritance column is after associations here, but before it in MiqGroup/ServiceResource/OrchestrationStack...   It might also help to add a line of whitespace between the association and non-association methods.

db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
```diff
@@ -0,0 +1,7 @@
+class AddEvmOwnerToOrchestrationStacks < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :orchestration_stacks, :evm_owner, :type => :bigint, :index => true
+    add_reference :orchestration_stacks, :miq_group, :type => :bigint, :index => true
+    add_reference :orchestration_stacks, :tenant, :type => :bigint, :index => true
```
jrafanie: Align type/index vertically?  👀

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,79 @@
...
+    say_with_time("Migrating existing orchestration stacks to have direct owners, groups, tenant") do
+      OrchestrationStack.all.each do |stack|
+        user = if stack.service.present?
+                 stack.service.tenant_identity
+               elsif stack.ems_id != nil
+                 stack.ext_management_system.tenant_identity
+               else
+                 User.super_admin
+               end
+        stack.update_attributes(:evm_owner_id => user.id, :tenant_id => user.current_tenant.id, :miq_group_id => user.current_group.id)
```
jrafanie: is it possible for `user` or `user.current_tenant` or `user.current_group` to be nil?

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+  end
+
+  def down
+    # blank because the down migration on AddEvmOwnerToOrchestrationStacks covers this
```
jrafanie: 👍

spec/migrations/20181016140921_migrate_orch_stacks_to_have_ownership_concept_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(orchestration_stack.count).to eq(1)
+
+      migrate
+
+      expect(stack.tenant_id).to be(nil)
+      expect(stack.evm_owner_id).to be(nil)
+      expect(stack.miq_group_id).to be(nil)
+    end
+
+    it "does something if ems exists and service doesn't" do
```
jrafanie: "sets owner, tenant, and group from the ems if the ems exists and service doesn't"

spec/migrations/20181016140921_migrate_orch_stacks_to_have_ownership_concept_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      migrate
+      stack.reload
+
+      expect(stack.evm_owner_id).to eq(ext_ms.tenant_identity.id)
+      expect(stack.tenant_id).to eq(ext_ms.tenant_id)
+      expect(stack.miq_group_id).to eq(ext_ms.tenant_identity.current_group.id)
+    end
+
+    it "does something if the service exists but no ems" do
```
jrafanie: "sets owner, tenant, and group from the service if the service exists and ems doesn't"

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
kbrock: in this test, yes `current_group is nil` - but in the real world. no

d-m-u:
WIP til tomorrow because it might be better to write the darn things in pure sql per feedback

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,85 @@
+class MigrateOrchStacksToHaveOwnershipConcept < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ApplicationRecord
```
lpichler: @d-m-u 
Migration have to be independent from our app as much as possible - so it is recommended to use `ActiveRecord::Base` instead of `ApplicationRecord`.

see https://github.com/ManageIQ/manageiq/pull/7312/files#r56356275

thanks

d-m-u:
@miq-bot remove_label wip

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,85 @@
+class MigrateOrchStacksToHaveOwnershipConcept < ActiveRecord::Migration[5.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    def tenant_identity
+      User.super_admin
```
kbrock: The user may not exist, but the group does exist. I notice that you do reference `current_group` below.

```suggestion
      User.super_admin.tap { |u| u.current_group = tenant.default_miq_group }
```

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
kbrock: If you are setting to `super_admin` because the group is off, , can you set the correct `miq_group`?

```suggestion
      if user.nil? || !user.miq_group_ids.include?(miq_group_id)
        user = User.super_admin.tap { |u| u.current_group_id = miq_group_id }
```

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,85 @@
...
+  class User < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+
+    has_and_belongs_to_many :miq_groups, :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+    belongs_to :current_group,           :class_name => "MigrateOrchStacksToHaveOwnershipConcept::MiqGroup"
+
+    include ActiveRecord::IdRegions
+
+    def self.super_admin
+      find_by(:userid => "admin")
```
kbrock: Is `in_my_region` available?

```suggestion
          in_my_region.find_by_userid("admin")
```

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
kbrock: Basically, if a user exists in this system, then they have an `miq_group`.
And if they have an `miq_group`, then they have an `current_group`. (code enforces that)

This is good as is

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
kbrock: this will work:

```ruby
class OrchestrationStack
  def root
    ancestry.present? ? self.class.find(ancestry.split("/").first) : self
  end

  def direct_service
    direct_services.first || (root.direct_services.first if root != self)
  end

  def service
    direct_service.try(:root_service) || (root.direct_service.try(:root_service) if root != self)
  end
end

class Service
  def root
    ancestry.present? ? self.class.find(ancestry.split("/").first) : self
  end
  alias root_service root
end
```

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
d-m-u: Ugh, this has been changed to reflect the last comment, it just isn't showing in github's UI as resolved...

d-m-u:
@miq-bot add_label hammer/yes

d-m-u:
@bdunne, @kbrock , @jrafanie, it's green, can you all please ... I mean I don't want to ask for any more reviews of this, I really don't ... kill this so I can stop having it haunt me?

db/migrate/20181016140921_migrate_orch_stacks_to_have_ownership_concept.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    def service
+      direct_service.try(:root_service) || (root.direct_service.try(:root_service) if root != self)
+    end
+  end
+
+  def up
+    say_with_time("Migrating existing orchestration stacks to have direct owners, groups, tenant") do
+      OrchestrationStack.all.each do |stack|
```
bdunne: Prefer `find_each` to handle batches

kbrock:
@bdunne you good with this now?

simaishi:
@bdunne @Fryguy it's ok to backport this to hammer branch?

bdunne:
No, there are schema changes.  I changed the labels accordingly.

himdel:
Not sure if this is a Problem, but, this migration fails when updating a gaprindashvili db to ivanchuk:

```
== 20181016140921 MigrateOrchStacksToHaveOwnershipConcept: migrating ==========
-- Migrating existing orchestration stacks to have direct owners, groups, tenant
rails aborted!

10000000000001 is out of range for ActiveModel::Type::Integer with limit 4 bytes
/home/himdel/manageiq-ivanchuk/lib/extensions/ar_migration.rb:88:in `migrate'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
ActiveModel::RangeError: 10000000000001 is out of range for ActiveModel::Type::Integer with limit 4 bytes
/home/himdel/manageiq-ivanchuk/lib/extensions/ar_migration.rb:88:in `migrate'
bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

bdunne:
Wow, what a coincidence, I was just looking into https://bugzilla.redhat.com/show_bug.cgi?id=1732808 which reports the same error.  @himdel Do you have a reproducer environment?

himdel:
Heheh, nice, I can share the db image :)

jrafanie:
@bdunne @himdel What version of id_regions?  maybe id_regions isn't 0.3.0?  You'll get that error if an old version migration is run with rails 5.1+.  This PR was needed to fix this:
https://github.com/ManageIQ/activerecord-id_regions/pull

himdel:
Hmmm, that *could* be it :)

I'm seeing 0.3.0 but running the migration again...

```
== 20181016140921 MigrateOrchStacksToHaveOwnershipConcept: migrating ==========
-- Migrating existing orchestration stacks to have direct owners, groups, tenant
   -> 15.2983s
== 20181016140921 MigrateOrchStacksToHaveOwnershipConcept: migrated (15.2983s) 
```

but I started my repo from hammer, so I guess the same `bin/update` that updated id_regions to 0.3.0 then ran `db:migrate`, presumably still using the previous version, or something like that?

jrafanie:
oh wow... yeah, I can see that...

```
Dir.chdir(ManageIQ::Environment::APP_ROOT) do
  ManageIQ::Environment.ensure_config_files

  puts '== Installing dependencies =='
  ManageIQ::Environment.install_bundler
  ManageIQ::Environment.bundle_update  # updates id_regions

  ManageIQ::Environment.while_updating_ui do
    unless ENV["SKIP_DATABASE_SETUP"]
      ManageIQ::Environment.migrate_database # uses old bundle that launched bin/update
      ManageIQ::Environment.seed_database
    end
```

bdunne:
The bundle update completes before the database is migrated in `bin/update`, it's just updating javascript while raking / seeding

himdel:
Well, `migrate_database` uses `run_rake_task`, which does `system!`.
So.. it is a new process.

However, bundler also rewrites `RUBYLIB`, `PATH`, `GEM_PATH` and `GEM_HOME` environment variables (and adds a bunch of `BUNDLER_*`) which do get propagated to the child process.

So... that could still be possible.. (I do remember having problems with ruby run from ruby when dealing with webpacker, so it does ring a bell, a bit :).)

Don't we have a clear bundle environment thing somewhere?

himdel:
> clear bundle environment thing

```ruby
        Bundler.with_clean_env do
          system("bundle exec rake update:actual_ui")
        end
```

I think we would need that ^^ for `db:migrate` to work during the same `bin/update`.

jrafanie:
@himdel can you help me recreate it?  Based on your information, I tried the following:

```
git checkout ivanchuk; git pull; bundle update; bin/rake db:migrate VERSION=20181023171353 
git checkout hammer; git pull; bin/update
git checkout ivanchuk; git pull; bin/update
```

The first step migrates the db back to the latest migration on hammer.
The second step gets to the dependencies at "hammer time" 🕺 
The third step tries to bin/update from ivanchuk, which should bundle and migrate together.

I can't recreate it this way.  Maybe I need something in my database in that table?

jrafanie:
Note, I put some findings into the bz... it looks like an issue of one migration running before another one polluting the memory space.  If you migrate a second time, it works.

https://bugzilla.redhat.com/show_bug.cgi?id=1732808#c8

himdel:
@jrafanie Sorry, apparenty not any more :(.

All I can tell you is that I took my manageiq hammer copy, updated to ivanchuk, and pointed it to a new database that came from a gaprindashvili copy I had lying around.

I can confirm that even if bin/update updates id.regions from 0.2.2 to 0.3.0 in the same run as running db:migrate, the migration passes. so, it was probably not that.

PR #289, opened 2018-10-12
Generate a new ansible rabbitmq password

carbonin:
If the current rabbitmq password contains special characters it
will fail a new preflight check in the setup playbook.

This is fixed for new installations by
https://github.com/ManageIQ/manageiq/pull/18092 but because we
re-run the setup playbook when we upgrade the tower version, we
also need to correct existing ones.

https://bugzilla.redhat.com/show_bug.cgi?id=1638009


db/migrate/20181012160010_remove_special_characters_from_ansible_rabbitmq_password.rb
```diff
@@ -0,0 +1,24 @@
...
+class RemoveSpecialCharactersFromAnsibleRabbitmqPassword < ActiveRecord::Migration[5.0]
+  # used only in specs
+  class MiqDatabase < ActiveRecord::Base; end
+
+  class Authentication < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    auth = Authentication.find_by(
```
Fryguy: Will there by multiple of these in the global, or do we not care?  Should we do .in_my_region?

Fryguy:
LGTM

db/migrate/20181012160010_remove_special_characters_from_ansible_rabbitmq_password.rb
carbonin: Well it shouldn't matter but we don't want these to get out of sync so I'll do the `.in_my_region`.

carbonin: It's not like the new password is deterministic so it would be particularly bad in this case

carbonin:
Damn didn't realize `String#match?` was 2.4+ ...

simaishi:
@bdunne @Fryguy Looks like no schema change involved. Please ack backport to `hammer` branch.

bdunne:
> Please ack backport to hammer branch.

:+1:

simaishi:
Hammer backport details:

```
$ git log -1
commit ff038c6575afec6cae8e1935fd725889c56b45ab
Date:   Fri Oct 12 15:25:32 2018 -0400

    Merge pull request #289 from carbonin/generate_new_ansible_rabbit_password
    
    Generate a new ansible rabbitmq password
    
    (cherry picked from commit af113564ede0f37b3d417e7b0f20760a9be29733)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1638009
```

PR #297, opened 2018-10-25
Add comments to the vms table

djberg96:
As of Rails 5 we can add comments to tables and columns within the database. This is useful for developers, both within RedHat and externally, to help develop for the application.

As per the Rails 5 documentation, "It's highly recommended to specify comments in migrations for applications with large databases as it helps people to understand data model and generate documentation."

To that end I've begun with the vms table, and documented as much as I could. However, I need your help to finish it. Anywhere that you see the text "[DEFINITION NEEDED]" means I either don't know or don't fully understand what the column does and need someone to fill in that information for me.

(Originally https://github.com/ManageIQ/manageiq-schema/pull/64, but got the rebase completely botched somehow, so resubmitting).


mwperina:
@borod108 @pkliczewski @masayag

Fryguy:
@bdunne @agrare

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :cpu_hot_add_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_hot_remove_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_limit, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve_expand, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares_level, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :created_on, "The date the VM was added to the app inventory."
+    change_column_comment :vms, :deprecated, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :description, "A custom description that users can add within the app."
```
bdunne: Isn't this refreshed from the provider?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :connection_state, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_affinity, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_hot_add_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_hot_remove_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_limit, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve_expand, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares_level, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :created_on, "The date the VM was added to the app inventory."
```
bdunne: Timestamp

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :memory_reserve_expand, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_shares_level, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :miq_group_id, "MIQ Group associated with the VM."
+    change_column_comment :vms, :name, "The name of the VM/Template."
+    change_column_comment :vms, :orchestration_stack_id, "ID of the orchestration stack associated with the VM."
+    change_column_comment :vms, :power_state, "The current power state, typically 'on' or 'off'."
+    change_column_comment :vms, :previous_state, "The previous power state change."
+    change_column_comment :vms, :publicly_available, "Indicates whether the VM is public or private."
+    change_column_comment :vms, :raw_power_state, "[DEFINITION NEEDED]"
```
bdunne: The power state refreshed from the ems

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :registered, "Indicates whether the VM registered to a Provider."
+    change_column_comment :vms, :resource_group_id, "ID of the resource group that the VM/Template belongs to."
+    change_column_comment :vms, :retired, "Indicates whether or not the VM is retired."
+    change_column_comment :vms, :retirement_last_warn, "Date of last retirement warning."
+    change_column_comment :vms, :retirement_requester, "Userid of the user who requested retirement."
+    change_column_comment :vms, :retirement_state, "The current retirement state."
+    change_column_comment :vms, :retirement_warn, "Warning message when VM was put into retirement."
+    change_column_comment :vms, :retires_on, "Date the VM is to be retired."
```
bdunne: Timestamp?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :smart, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :standby_action, "Action taken when VM is put into standby mode."
+    change_column_comment :vms, :storage_id, "ID of the storage object associated with the VM."
+    change_column_comment :vms, :storage_profile_id, "ID of the storage profile associated with the VM."
+    change_column_comment :vms, :state_changed_on, "The date of the last power state change."
```
bdunne: Timestamp?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :template, "Indicates if a template was used to provision the VM or not."
```
bdunne: Indicates whether this is a VM or a template.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :tenant_id, "ID of the tenant associated with the VM."
+    change_column_comment :vms, :tools_status, "Status of VMWare tools."
```
bdunne: Guest tools (not vmware specific)

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :type, "An internal class name that includes the provider type."
+    change_column_comment :vms, :uid_ems, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :updated_on, "The last date the VM was modified within the app."
```
bdunne: Timestamp

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :vendor, "The vendor for the VM, e.g. Azure, Amazon, etc."
+    change_column_comment :vms, :version, "[DEFINITION NEEDED]"
```
bdunne: Version of the VM definition

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :guid, "A value used to uniquely identify the VM internally."
+    change_column_comment :vms, :id, "The internal database ID."
+    change_column_comment :vms, :host_id, "ID of the host that the VM/Template resides on."
+    change_column_comment :vms, :last_perf_capture_on, "Timestamp for last performance metrics capture."
+    change_column_comment :vms, :last_scan_on, "Timestamp for last SmartState Analysis."
+    change_column_comment :vms, :last_scan_attempt_on, "Timestamp for last attempt to perform SmartState Analysis."
+    change_column_comment :vms, :last_sync_on, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :linked_clone, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :location, "The location or region of the VM."
+    change_column_comment :vms, :memory_hot_add_enabled, "[DEFINITION NEEDED]"
```
bdunne: Indicates whether or not memory can be hot added to the running vm

bdunne:
:+1: I like the idea, I'd like to populate as many of the definitions as possible before merge.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :cpu_shares_level, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :created_on, "The date the VM was added to the app inventory."
+    change_column_comment :vms, :deprecated, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :description, "A custom description that users can add within the app."
+    change_column_comment :vms, :ems_created_on, "Date the associated EMS was created."
+    change_column_comment :vms, :ems_id, "ID of the ExtManagementSystem associated with the VM."
+    change_column_comment :vms, :evm_owner_id, "ID of VM owner from product database."
+    change_column_comment :vms, :ems_ref_obj, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :ems_cluster_id, "ID of associated cluster."
+    change_column_comment :vms, :fault_tolerance, "[DEFINITION NEEDED]"
```
agrare: Boolean indicating if a VM is fault tolerant aka if one host crashes it will immediately pick up on another host in the cluster

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :flavor_id, "ID of associated flavor."
+    change_column_comment :vms, :format, "The format of the VM's disk, such as VMDK."
+    change_column_comment :vms, :guid, "A value used to uniquely identify the VM internally."
+    change_column_comment :vms, :id, "The internal database ID."
+    change_column_comment :vms, :host_id, "ID of the host that the VM/Template resides on."
+    change_column_comment :vms, :last_perf_capture_on, "Timestamp for last performance metrics capture."
+    change_column_comment :vms, :last_scan_on, "Timestamp for last SmartState Analysis."
+    change_column_comment :vms, :last_scan_attempt_on, "Timestamp for last attempt to perform SmartState Analysis."
+    change_column_comment :vms, :last_sync_on, "[DEFINITION NEEDED]"
```
agrare: This looks like the last time the smartstate info was updated by the scanning mixin, not sure how this is different from last_scan_on, @roliveri ?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :linked_clone, "[DEFINITION NEEDED]"
```
agrare: A boolean indicating if a VM is a "linked clone" aka its base disk is a differencing disk pointing back to a snapshot of another VM.  This is a way of creating a high number of clones without using much storage space.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :smart, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :standby_action, "Action taken when VM is put into standby mode."
+    change_column_comment :vms, :storage_id, "ID of the storage object associated with the VM."
+    change_column_comment :vms, :storage_profile_id, "ID of the storage profile associated with the VM."
+    change_column_comment :vms, :state_changed_on, "The date of the last power state change."
+    change_column_comment :vms, :template, "Indicates if a template was used to provision the VM or not."
+    change_column_comment :vms, :tenant_id, "ID of the tenant associated with the VM."
+    change_column_comment :vms, :tools_status, "Status of VMWare tools."
+    change_column_comment :vms, :type, "An internal class name that includes the provider type."
+    change_column_comment :vms, :uid_ems, "[DEFINITION NEEDED]"
```
agrare: The globally unique reference used to reconnect VMs disconnected from other providers.  Most providers set this equal to the ems_ref (either because most have globally unique ems_refs anyway, or because there isn't anything better available).

For vmware this is the bios uuid which comes from the vmx file, so if you move this vm to another vcenter the uid_ems will stay the same even if the ems_ref is different.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
+class AddCommentsToVmsTable < ActiveRecord::Migration[5.0]
+  def change
+    change_table_comment :vms, "Virtual Machines and Templates"
+
+    change_column_comment :vms, :autostart, "Indicates whether or not the VM is set to autostart."
+    change_column_comment :vms, :availability_zone_id, "ID of the availability zone that the VM is part of."
+    change_column_comment :vms, :busy, "[DEFINITION NEEDED]"
```
agrare: I don't actually know what this one is and I can't find anyone who sets it, need to do more investigation on this one...

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :boot_time, "The last time the VM was started."
+    change_column_comment :vms, :cloud, "Indicates whether or not the VM/Template is cloud or infra."
```
agrare: We really have a column for this?  I don't see amazon setting this, need to see if anyone else does.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :config_xml, "The VM or Template configuration in XML format. Deprecated."
+    change_column_comment :vms, :connection_state, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_affinity, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_hot_add_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_hot_remove_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_limit, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_reserve_expand, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :cpu_shares_level, "[DEFINITION NEEDED]"
```
agrare: These are all related to resourcePool settings, probably easiest to read up on https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.ResourceAllocationInfo.html

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :linked_clone, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :location, "The location or region of the VM."
+    change_column_comment :vms, :memory_hot_add_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_hot_add_increment, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_hot_add_limit, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_limit, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_reserve, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_reserve_expand, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_shares_level, "[DEFINITION NEEDED]"
```
agrare: Same as https://github.com/ManageIQ/manageiq-schema/pull/297/files#r228541679 but for memory

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :id, "The internal database ID."
+    change_column_comment :vms, :host_id, "ID of the host that the VM/Template resides on."
+    change_column_comment :vms, :last_perf_capture_on, "Timestamp for last performance metrics capture."
+    change_column_comment :vms, :last_scan_on, "Timestamp for last SmartState Analysis."
+    change_column_comment :vms, :last_scan_attempt_on, "Timestamp for last attempt to perform SmartState Analysis."
+    change_column_comment :vms, :last_sync_on, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :linked_clone, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :location, "The location or region of the VM."
+    change_column_comment :vms, :memory_hot_add_enabled, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_hot_add_increment, "[DEFINITION NEEDED]"
```
agrare: Any memory hot-added to a VM must be a multiple of this value:
> Memory, in MB that can be added to a running virtual machine must be in increments of this value and needs be a multiple of this value. This value is determined by the virtual machine and is specified only if memoryHotAddEnabled has been set to true.

From https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.ConfigInfo.html

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :memory_hot_add_limit, "[DEFINITION NEEDED]"
```
agrare: The max memory that can be hot-added to a VM

> The maximum amount of memory, in MB, than can be added to a running virtual machine. This value is determined by the virtual machine and is specified only if memoryHotAddEnabled is set to true.

From https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.ConfigInfo.html

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :memory_shares, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :memory_shares_level, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :miq_group_id, "MIQ Group associated with the VM."
+    change_column_comment :vms, :name, "The name of the VM/Template."
+    change_column_comment :vms, :orchestration_stack_id, "ID of the orchestration stack associated with the VM."
+    change_column_comment :vms, :power_state, "The current power state, typically 'on' or 'off'."
+    change_column_comment :vms, :previous_state, "The previous power state change."
+    change_column_comment :vms, :publicly_available, "Indicates whether the VM is public or private."
+    change_column_comment :vms, :raw_power_state, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :registered, "Indicates whether the VM registered to a Provider."
```
agrare: There's a column for this?  We have a method [`VmOrTemplate#registered?`](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L311-L315) which checks template, ems_id, host_id

agrare: And I don't see any of the parsers setting this.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :resource_group_id, "ID of the resource group that the VM/Template belongs to."
+    change_column_comment :vms, :retired, "Indicates whether or not the VM is retired."
+    change_column_comment :vms, :retirement_last_warn, "Date of last retirement warning."
+    change_column_comment :vms, :retirement_requester, "Userid of the user who requested retirement."
+    change_column_comment :vms, :retirement_state, "The current retirement state."
+    change_column_comment :vms, :retirement_warn, "Warning message when VM was put into retirement."
+    change_column_comment :vms, :retires_on, "Date the VM is to be retired."
+    change_column_comment :vms, :smart, "[DEFINITION NEEDED]"
```
agrare: I don't see this being set or used anywhere, @Fryguy do you know what this is?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,81 @@
...
+    change_column_comment :vms, :autostart, "Indicates whether or not the VM is set to autostart."
+    change_column_comment :vms, :availability_zone_id, "ID of the availability zone that the VM is part of."
+    change_column_comment :vms, :busy, "[DEFINITION NEEDED]"
+    change_column_comment :vms, :boot_time, "The last time the VM was started."
+    change_column_comment :vms, :cloud, "Indicates whether or not the VM/Template is cloud or infra."
+    change_column_comment :vms, :cloud_network_id, "ID of the cloud (virtual) network the VM/Template is associated with."
+    change_column_comment :vms, :cloud_subnet_id, "Subnet ID associated with the VM."
+    change_column_comment :vms, :cloud_tenant_id, "ID of the cloud tenant associated with the VM."
+    change_column_comment :vms, :config_xml, "The VM or Template configuration in XML format. Deprecated."
+    change_column_comment :vms, :connection_state, "[DEFINITION NEEDED]"
```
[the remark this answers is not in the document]
djberg96: Ok, though I noticed it's a text field. I guess there are more than 2 possible values then?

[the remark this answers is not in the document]
agrare: https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/index.html <- all the possible values

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: Should I mark it deprecated then?

djberg96:
@agrare @bdunne Ok, made a bunch of updates.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: Hm, I want to say this had something to do with smart proxy. @gmcculloug I think we talked about this a long time ago, but my memory is hazy.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: We set it in Azure, though I don't remember how it's used, if at all.

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: That's correct, it is very old and no longer used.  It use to indicate that we added a file in with the other VM files to help identify the VM if it moved between systems.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: I think it depends on the provider. It could be, but it's also an editable field. I'll update it.

db/migrate/20181025230931_add_comments_to_vms_table.rb
bdunne: @gtanzillo @Fryguy ?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :cloud_tenant_id, "ID of the cloud tenant associated with the VM."
+    change_column_comment :vms, :config_xml, "The VM or Template configuration in XML format. Deprecated."
+    change_column_comment :vms, :connection_state, "Indicates the connection state of the VM, e.g. connected, disconnected, etc."
+    change_column_comment :vms, :cpu_affinity, "A comma separated list of host CPUs to schedule the VM on. VMware only."
+    change_column_comment :vms, :cpu_hot_add_enabled, "Indicates if CPUs can be added to the VM while it is powered on."
+    change_column_comment :vms, :cpu_hot_remove_enabled, "Indicates if a CPU can be removed from a VM while powered on."
+    change_column_comment :vms, :cpu_limit, "The CPU utilization limit for the VM. VMWare only."
+    change_column_comment :vms, :cpu_reserve, "Amount of CPU that is guaranteed available for the VM. VMWare only."
+
+    change_column_comment :vms, :cpu_reserve_expand, "Amount the CPU can grow beyond its normal limits if there "\
```
bdunne: Any reason to multi-line these strings?

db/migrate/20181025230931_add_comments_to_vms_table.rb
bdunne: Sounds like we should drop the column

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: They were getting kind of ridiculously long.

agrare:
Actually I think I found what it was used for, it was removed recently from the ScanningOperationsMixin: https://github.com/ManageIQ/manageiq/pull/15896/files#diff-0d6cda428e9b79334ae1d321fafcbf94L55

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: I want to say this is old (circa '07) SSA stuff.  It is should not be needed anymore.

djberg96:
@agrare @bdunne Ok, updated that column, please let me know if the description I gave it is acceptable.

djberg96:
@gmcculloug Ok, updated the comment to mark it deprecated, thanks.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :storage_id, "ID of the storage object associated with the VM."
+    change_column_comment :vms, :storage_profile_id, "ID of the storage profile associated with the VM."
+    change_column_comment :vms, :state_changed_on, "The timestamp of the last power state change."
+    change_column_comment :vms, :template, "Indicates whether this is a VM or template."
+    change_column_comment :vms, :tenant_id, "ID of the tenant associated with the VM."
+    change_column_comment :vms, :tools_status, "Status of guest tools."
+    change_column_comment :vms, :type, "An internal class name that includes the provider type."
+    change_column_comment :vms, :uid_ems, "A globally unique reference for the VM across EMS."
+    change_column_comment :vms, :updated_on, "The last timestamp the VM was modified within the app."
+    change_column_comment :vms, :vendor, "The vendor for the VM, e.g. Azure, Amazon, etc."
```
Fryguy: vendor values are lowercased, so the examples are incorrect. See https://github.com/ManageIQ/manageiq/blob/242bd55197020baf8c385dc35110051cc2947378/app/models/vm_or_template.rb#L40-L53

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    change_column_comment :vms, :memory_reserve_expand, "Amount the memory can grow beyond its normal limits if there "\
+                                                     "are unreserved resources available. VMWare only."
+
+    change_column_comment :vms, :memory_shares, "The number of memory shares allocated. Used to determine resource allocation "\
+                                             "in case of resource contention. VMWare only."
+
+    change_column_comment :vms, :memory_shares_level, "Memory shares allocation level. VMWare only."
+
+    change_column_comment :vms, :miq_group_id, "MIQ Group associated with the VM."
```
Fryguy: Should this be similar to evm_owner_id, since this is about ownership (I think...@gtanzillo?)?  So maybe "ID of MiqGroup owner." or something?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :cpu_shares, "The number of CPU shares allocated. Used to determine resource allocation "\
+                                             "in case of resource contention. VMWare only."
+
+    change_column_comment :vms, :cpu_shares_level, "CPU shares allocation level. VMWare only."
+    change_column_comment :vms, :created_on, "The timestamp the VM was added to the app inventory."
+    change_column_comment :vms, :deprecated, "Indicates whether or not a template is deprecated. Google only."
+    change_column_comment :vms, :description, "A description of the VM that is typically set by the provider, but may also be edited."
+    change_column_comment :vms, :ems_created_on, "Date the associated EMS was created."
+    change_column_comment :vms, :ems_id, "ID of the ExtManagementSystem associated with the VM."
+    change_column_comment :vms, :evm_owner_id, "ID of VM owner from product database."
```
Fryguy: I dont understand the "from product database" part...

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    change_column_comment :vms, :cpu_reserve_expand, "Amount the CPU can grow beyond its normal limits if there "\
+                                                     "are unreserved resources available. VMWare only."
+
+    change_column_comment :vms, :cpu_shares, "The number of CPU shares allocated. Used to determine resource allocation "\
+                                             "in case of resource contention. VMWare only."
+
+    change_column_comment :vms, :cpu_shares_level, "CPU shares allocation level. VMWare only."
+    change_column_comment :vms, :created_on, "The timestamp the VM was added to the app inventory."
+    change_column_comment :vms, :deprecated, "Indicates whether or not a template is deprecated. Google only."
```
Fryguy: We shouldn't put provider specifics into the database layer...tomorrow Azure could add deprecated templates, and then this comment is hard baked into the schema requiring a data migraiton.

Fryguy: Same goes for the places where we say "VMware only".

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :description, "A description of the VM that is typically set by the provider, but may also be edited."
+    change_column_comment :vms, :ems_created_on, "Date the associated EMS was created."
```
Fryguy: Perhaps "Date the VM or Templated was created on the EMS itself".  The current text sounds like it's when the EMS itself was created

db/migrate/20181025230931_add_comments_to_vms_table.rb
lpichler: yes it is about group ownership. (screen "set ownership")

db/migrate/20181025230931_add_comments_to_vms_table.rb
Fryguy: Or perhaps start with "The timestamp", since you've used that wording for other datetime columnes

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :evm_owner_id, "ID of VM owner from product database."
+    change_column_comment :vms, :ems_ref, "A native unique reference for the VM within the EMS."
+    change_column_comment :vms, :ems_ref_obj, "A YAML serialized version of the ems_ref."
+    change_column_comment :vms, :ems_cluster_id, "ID of associated cluster."
+
+    change_column_comment :vms, :fault_tolerance, "Indicates whether or not the VM is fault tolerant, i.e. if one host "\
+                                                  "crashes it will immediately pick up on another host in the cluster."
+
+    change_column_comment :vms, :flavor_id, "ID of associated flavor."
+    change_column_comment :vms, :format, "The format of the VM's disk, such as VMDK."
```
Fryguy: Not sure if this is lower case or not, but the example needs to be accurate.

db/migrate/20181025230931_add_comments_to_vms_table.rb
agrare: More specifically this is the ID of the `User` which provisioned the VM if it was provisioned through MIQ.

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+    change_column_comment :vms, :guid, "A value used to uniquely identify the VM internally."
+    change_column_comment :vms, :id, "The internal database ID."
+    change_column_comment :vms, :host_id, "ID of the host that the VM/Template resides on."
+    change_column_comment :vms, :last_perf_capture_on, "Timestamp for last performance metrics capture."
+    change_column_comment :vms, :last_scan_on, "Timestamp for last SmartState analysis."
+    change_column_comment :vms, :last_scan_attempt_on, "Timestamp for last attempt to perform SmartState Analysis."
+    change_column_comment :vms, :last_sync_on, "Timestamp for last SmartState analysis."
```
agrare: We should differentiate this from [+47](https://github.com/ManageIQ/manageiq-schema/pull/297/files#diff-77e2987773d5e054cf6983b73575d32fR47)

@roliveri what is the difference?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    change_column_comment :vms, :linked_clone, "Indicates whether the VM is a linked clone, i.e its disk is a referencing "\
+                                               "disk pointing to snapshot of another VM."
+
+    change_column_comment :vms, :location, "The location or region of the VM."
```
agrare: For infra VMs this is the storage location e.g. `[datastore 1] vm/vm.vmx` if you want to include an example

db/migrate/20181025230931_add_comments_to_vms_table.rb
agrare: Yes

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: Owner can be set through inventory, provisioning, UI or Automate.  Let's not pick only provisioning.  😉

djberg96: @gmcculloug Updated, thanks.

db/migrate/20181025230931_add_comments_to_vms_table.rb
roliveri: Not sure if this was answered yet, Greg M might know better. Scanning is acquiring the data from the VM, sync'ing is saving that data to the database. When we used a remote smartproxy, the time gap between those operations could be significant.

db/migrate/20181025230931_add_comments_to_vms_table.rb
Fryguy: I think we had added it for the UI so they could more easily differentiate.  We'd have to go into the history otherwise.

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: The answer @roliveri provided is correct.  The scanning and synchronizing of data did not always have to happen together.  The scan could happen multiple times local to the hypervisor and cached, then a single sync would transfer and process all the data.

Today I believe they are always performed together.

Either way I assume this is enough to differentiate:
`change_column_comment :vms, :last_scan_on, "Timestamp for last SmartState analysis scan."`
and
`change_column_comment :vms, :last_sync_on, "Timestamp for last SmartState analysis synchronization."`

djberg96: @gmcculloug @roliveri Ok, updated as per Greg's suggestion.

db/migrate/20181025230931_add_comments_to_vms_table.rb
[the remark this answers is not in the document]
djberg96: @agrare Currently set to "Indicates the connection state of the VM, e.g. connected, disconnected, etc.". Is that alright?

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: @agrare @gmcculloug Currently set to "Indicates whether or not the VM is being scanned for SmartState analysis. Deprecated.". If that's ok, please mark this resolved.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: @agrare @Fryguy Should I mark this deprecated, or update it in some way? Otherwise, please resolve.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: @agrare @gmcculloug @Fryguy Currently set to "Typically, the ID of the user that provisioned the VM (if provisioned through the app). Can also be set via inventory, the UI or Automate."

If you're ok with that please resolve.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: @Fryguy I'm guessing it would depend on the provider. I don't remember where I got that example from now, maybe @mfeifer.

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: @agrare For now it's just set to "The location of the VM, or its underlying storage." Is that alright?

db/migrate/20181025230931_add_comments_to_vms_table.rb
```diff
@@ -0,0 +1,105 @@
...
+    change_column_comment :vms, :linked_clone, "Indicates whether the VM is a linked clone, i.e its disk is a referencing "\
+                                               "disk pointing to snapshot of another VM."
+
+    change_column_comment :vms, :location, "The location of the VM, or its underlying storage."
+    change_column_comment :vms, :memory_hot_add_enabled, "Indicates whether or not memory can be hot added to the running VM."
+
+    change_column_comment :vms, :memory_hot_add_increment, "Memory in MB that can be added to a running VM must be in "\
+                                                           "increments of this value."
+
+    change_column_comment :vms, :memory_hot_add_limit, "The maximum amount of memory, in MB, than can be added to a running VM."
```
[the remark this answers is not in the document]
djberg96: Fixed, thanks.

db/migrate/20181025230931_add_comments_to_vms_table.rb
agrare: I don't see any of the cloud parsers setting this but lets hold off on marking it deprecated

db/migrate/20181025230931_add_comments_to_vms_table.rb
agrare: I don't see this being set by any parser either, @roliveri do you know if this is set by smartstate?

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: I'm good with this one, but I don't have the "Resolve Conversation" button on this repo.  So please accept this in it's place: :white_check_mark:

db/migrate/20181025230931_add_comments_to_vms_table.rb
gmcculloug: A little wordy, but covers all the possibilities.  ✅

db/migrate/20181025230931_add_comments_to_vms_table.rb
djberg96: Confirmed by @gmcculloug to be deprecated.

djberg96:
@agrare Switched "change" to "up" as per your suggestion.

@Fryguy @agrare @bdunne I think it's ready to go.

PR #302, opened 2018-11-08
Create migrations to add resource_type row

juliancheal:
Adds new row
Migrates existing data to the new row

This PR needs to be merged in order see issue for ordering https://github.com/ManageIQ/manageiq/issues/18222

### DIFF
--- a/db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
+++ b/db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
@@ -1,0 +1,20 @@
+class MigrateMiqPolicyTowhatToResourceType < ActiveRecord::Migration[5.0]
+  class MiqPolicy < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Moving MiqPolicy towhat to resource type") do
+      MiqPolicy.find_each do |policy|
+        policy.update(:resource_type => policy.towhat)
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving MiqPolicy resource type to towhat") do
+      MiqPolicy.find_each do |policy|
+        policy.update(:towhat => policy.resource_type)
+      end
+    end
+  end
+end
--- a/db/migrate/20181108195945_migrate_conditions_tow_hat_to_resource_type.rb
+++ b/db/migrate/20181108195945_migrate_conditions_tow_hat_to_resource_type.rb
@@ -1,0 +1,20 @@
+class MigrateConditionsTowHatToResourceType < ActiveRecord::Migration[5.0]
+  class Condition < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Moving Condition towhat to resource type") do
+      Condition.find_each do |condition|
+        condition.update(:resource_type => condition.towhat)
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving Condition resource type to towhat") do
+      Condition.find_each do |condition|
+        condition.update(:towhat => condition.resource_type)
+      end
+    end
+  end
+end
--- a/spec/migrations/20181108195546_migrate_miq_policy_towhat_to_resource_type_spec.rb
+++ b/spec/migrations/20181108195546_migrate_miq_policy_towhat_to_resource_type_spec.rb
@@ -1,0 +1,23 @@
+require_migration
+
+describe MigrateMiqPolicyTowhatToResourceType do
+  let(:miq_policy_stub) { migration_stub(:MiqPolicy) }
+
+  migration_context :up do
+    it "converts the old data" do
+      miq_policy = miq_policy_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(miq_policy.reload.resource_type).to eq("ContainerImage")
+    end
+
+    it "ignores already converted data" do
+      miq_policy = miq_policy_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(miq_policy.reload.resource_type).to eq("ContainerImage")
+    end
+  end
+end
--- a/spec/migrations/20181108195945_migrate_conditions_tow_hat_to_resource_type_spec.rb
+++ b/spec/migrations/20181108195945_migrate_conditions_tow_hat_to_resource_type_spec.rb
@@ -1,0 +1,23 @@
+require_migration
+
+describe MigrateConditionsTowHatToResourceType do
+  let(:condition_stub) { migration_stub(:Condition) }
+
+  migration_context :up do
+    it "converts the old data" do
+      condition = condition_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(condition.reload.resource_type).to eq("ContainerImage")
+    end
+
+    it "ignores already converted data" do
+      condition = condition_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(condition.reload.resource_type).to eq("ContainerImage")
+    end
+  end
+end
--- a/db/migrate/20181108194142_add_resource_type_to_miq_policy_and_condition.rb
+++ b/db/migrate/20181108194142_add_resource_type_to_miq_policy_and_condition.rb
@@ -1,0 +1,6 @@
+class AddResourceTypeToMiqPolicyAndCondition < ActiveRecord::Migration[5.0]
+  def change
+    add_column :miq_policies, :resource_type, :string
+    add_column :conditions,   :resource_type, :string
+  end
+end


jrafanie:
@juliancheal how hard is it to get all the callers to use the new column?  If we merge this, they'll be looking for towhat column, which no longer exists.  It feels like we might need to:

* add the new column, `resource_type`
* override `towhat` to `read_attribute(:towhat) || read_attribute(:resource_type)`
* migrate the data from `towhat` to `resource_type`
* change callers to use `resource_type`
* Drop the override for `towhat` in step 2

juliancheal:
@jrafanie 
> override towhat to read_attribute(:towhat) || read_attribute(:resource_type)

Is a great idea. Apart from tests for this PR and adding that above line, I think in the core and ui repos i've removed all the callers (that i know of) to `towhat`.

This sounds like a good way to proceed.

db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
```diff
@@ -0,0 +1,24 @@
+class MigrateMiqPolicyTowhatToResourceType < ActiveRecord::Migration[5.0]
+
+  class MiqPolicy < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Moving MiqPolicy towhat to resource type") do
+      MiqPolicy.all.each do |policy|
```
jrafanie: I dont think there's too many, but maybe we want find_each

db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
```diff
@@ -0,0 +1,24 @@
...
+        policy.resource_type = policy.towhat
+        policy.save
```
jrafanie: `policy.update(:resource_type => policy.towhat)`

db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving MiqPolicy resource type to towhat") do
+      MiqPolicy.all.each do |policy|
+        policy.towhat = policy.resource_type
+        policy.save
```
jrafanie: same as above

db/migrate/20181108195945_migrate_conditions_tow_hat_to_resource_type.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  class Condition < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Moving Condition towhat to resource type") do
+      Condition.all.each do |condition|
+        condition.resource_type = condition.towhat
+        condition.save
```
jrafanie: same as above

db/migrate/20181108195945_migrate_conditions_tow_hat_to_resource_type.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+  end
+
+  def down
+    say_with_time("Moving Condition resource type to towhat") do
+      Condition.all.each do |condition|
+        condition.towhat = condition.resource_type
+        condition.save
```
jrafanie: ditto

spec/migrations/20181108195546_migrate_miq_policy_towhat_to_resource_type_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+describe MigrateMiqPolicyTowhatToResourceType do
+  let(:miq_policy_stub) { migration_stub(:MiqPolicy) }
+
+  migration_context :up do
+    it "converts the old data" do
+      miq_policy = miq_policy_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(miq_policy.reload).to eq(:resource_type => "ContainerImage")
```
jrafanie: `expect(miq_policy.reload.resource_type).to eq("ContainerImage")`

spec/migrations/20181108195546_migrate_miq_policy_towhat_to_resource_type_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    it "ignores already converted data" do
+      miq_policy = miq_policy_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(miq_policy.reload).to eq(:resource_type => "ContainerImage")
```
jrafanie: same as above

spec/migrations/20181108195945_migrate_conditions_tow_hat_to_resource_type_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      expect(condition.reload).to eq(:resource_type => "ContainerImage")
+    end
+
+    it "ignores already converted data" do
+      condition = condition_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(condition.reload).to eq(:resource_type => "ContainerImage")
```
jrafanie: same

spec/migrations/20181108195945_migrate_conditions_tow_hat_to_resource_type_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+describe MigrateMiqPolicyTowHatToResourceType do
+  let(:condition_stub) { migration_stub(:Condition) }
+
+  migration_context :up do
+    it "converts the old data" do
+      condition = condition_stub.create!(:towhat => "ContainerImage")
+
+      migrate
+
+      expect(condition.reload).to eq(:resource_type => "ContainerImage")
```
jrafanie: same

db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
```diff
@@ -0,0 +1,21 @@
+class MigrateMiqPolicyTowhatToResourceType < ActiveRecord::Migration[5.0]
+  class MiqPolicy < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
```
lpichler: it is not needed for `MiqPolicy` as there is no `MiqPolicy#type` column see: https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-310643185

db/migrate/20181108195945_migrate_conditions_tow_hat_to_resource_type.rb
```diff
@@ -0,0 +1,21 @@
+class MigrateConditionsTowHatToResourceType < ActiveRecord::Migration[5.0]
+  class Condition < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
```
lpichler: same here , it is not needed https://github.com/ManageIQ/manageiq-schema/issues/4#issuecomment-310643185

db/migrate/20181108195546_migrate_miq_policy_towhat_to_resource_type.rb
lpichler: or eventually one sql query `MiqPolicy.update_all('resource_type = resource_type')`

lpichler:
LGTM and I had also some minor comments 🇨🇿 😄

juliancheal:
@jrafanie seeing as you're off next week, shall we put a pin in this until you return?

jrafanie:
LGTM, I can't merge here but it looks good.  @juliancheal you can work with @bdunne or @carbonin, I believe if they're good with the approach, it will be fine to do while I'm away.

bdunne:
Any reason not to `rename_column`?  It would be much faster and avoid the triple migration (create, migrate data, delete).  We could `deprecate_attribute` on the model side as well.

juliancheal:
> Any reason not to `rename_column`? It would be much faster and avoid the triple migration (create, migrate data, delete). We could `deprecate_attribute` on the model side as well.

@bdunne I'm sure there is a reason @jrafanie?

jrafanie:
> Any reason not to `rename_column`? It would be much faster and avoid the triple migration (create, migrate data, delete). We could `deprecate_attribute` on the model side as well.

The only reason is that we can't rename the column and change the caller code in core and ui-classic all in one PR.  The core and ui-classic PRs depend on the `resource_type` existing so they can't go green until the column is created.

jrafanie:
The idea was to:
* create the `resource_type` column (get it merged)
* create the core, ui-classic and schema (data migration) PRs, get them all green, then merge together
* then drop the column

jrafanie:
The intent is to create the column, leave it unused, get the 3 co-dependent PRs to use/migrate to the new column merged together, then drop the column.

I'm not sure how to do rename column with two other repos needing to use the new or old column based on the schema version.  How do you suggest we do that? The core and ui-classic PRs can't become green until the column is renamed so you can't time them to use the correct column until they're green and merged.

juliancheal:
After speaking with @bdunne and @jrafanie we have a plan!

PR #303, opened 2018-11-08
[WIP] Switch to manageiq-password instead of gems-pending

agrare:
Use the extracted manageiq-password gem instead of gems-pending

### DIFF
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -16,12 +16,13 @@ Gem::Specification.new do |s|
   s.files = Dir["{db,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
 
   s.add_dependency "activerecord-id_regions", "~> 0.2.2"
+  s.add_dependency "manageiq-password", "~> 0.1.0"
   s.add_dependency "more_core_extensions", "~> 3.5"
   s.add_dependency "pg", "~> 0.18.2"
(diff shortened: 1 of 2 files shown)


martinpovolny:
@Fryguy, @chessbyte : seems this one is on the critical path to fix master and you 2 are the only ones with the permissions.

This cannot be fixed from this side of the ocean.

Or can @NickLaMuro also merge this one?

agrare:
@martinpovolny I think its better to rollback the change to smartstate before applying any more of these

manageiq-schema.gemspec
```diff
@@ -16,12 +16,13 @@ Gem::Specification.new do |s|
   s.files = Dir["{db,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
 
   s.add_dependency "activerecord-id_regions", "~> 0.2.2"
+  s.add_dependency "manageiq-password", "~> 0.1.0"
   s.add_dependency "more_core_extensions", "~> 3.5"
   s.add_dependency "pg", "~> 0.18.2"
   s.add_dependency "pg-pglogical", "~> 2.1.1"
   s.add_dependency "rails", "~> 5.0.2"
 
-  s.add_dependency "manageiq-gems-pending"  # This is just for MiqPassword for now
```
bdunne: This should also be removed from the `Gemfile`

bdunne:
I can merge this when ready

PR #305, opened 2018-11-19
Default VM connection_state is now 'connected'

miha-plesko:
With this commit we migrate all occurances of

```
vm.connection_state == nil
```

to now contain 'connected' instead nil.

@miq-bot assign @agrare 

/cc @kbrock

### DIFF
--- a/db/migrate/20181119154009_set_vm_connection_state.rb
+++ b/db/migrate/20181119154009_set_vm_connection_state.rb
@@ -1,0 +1,14 @@
+class SetVmConnectionState < ActiveRecord::Migration[4.2]
+  class VmOrTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+    self.table_name = 'vms'
+  end
+
+  def up
+    say_with_time('Filling nil connection_state to "connected"') do
+      VmOrTemplate.in_my_region.where(:connection_state => nil).update_all(:connection_state => 'connected')
+    end
+  end
+end
--- a/spec/migrations/20181119154009_set_vm_connection_state_spec.rb
+++ b/spec/migrations/20181119154009_set_vm_connection_state_spec.rb
@@ -1,0 +1,26 @@
+require_migration
+
+describe SetVmConnectionState do
+  let(:vm_stub)          { migration_stub(:VmOrTemplate) }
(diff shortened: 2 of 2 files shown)


miha-plesko:
Related to https://github.com/ManageIQ/manageiq/pull/18216

@miq-bot add_label data

db/migrate/20181119154009_set_vm_connection_state.rb
```diff
@@ -0,0 +1,11 @@
+class SetVmConnectionState < ActiveRecord::Migration[4.2]
+  class Vm < ActiveRecord::Base
```
agrare: Should be VmOrTemplate

db/migrate/20181119154009_set_vm_connection_state.rb
```diff
@@ -0,0 +1,11 @@
...
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time('Filling nil connection_state to "connected"') do
+      Vm.where(:connection_state => nil).update_all(:connection_state => 'connected')
```
agrare: should add `VmOrTemplate.in_my_region.where`

miha-plesko:
Nice find @agrare , didn't think of VmOrTemplate, updated.

spec/migrations/20181119154009_set_vm_connection_state_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        let(:connection_state) { 'disconnected' }
+        it 'disconnected stays disconnected' do
+          migrate
+          vm.reload
+          expect(vm.connection_state).to eq(connection_state)
+        end
+      end
+    end
+
+    describe 'in other region' do
```
agrare: @miha-plesko we don't need to test that `in_my_region` works so you can just get rid of this and simplify the tests a bit

miha-plesko:
Simplified the specs, good to go @agrare

agrare:
Thanks @miha-plesko

kbrock:
Also, thanks for using `update_all` and ignoring rubocop.

Good stuff

agrare:
> Also, thanks for using update_all and ignoring rubocop.

:+1: agreed

> (0% snark)

:laughing:

PR #309, opened 2018-11-30
classification#parent_id = nil when no parents

kbrock:
We used to have parent_id=0 when there was no parent.
Rails prefers to use a nil

This is in response to @Fryguy comment in ManageIQ/manageiq#17210

### DIFF
--- a/db/migrate/20181130203334_classification_parent_null.rb
+++ b/db/migrate/20181130203334_classification_parent_null.rb
@@ -1,0 +1,14 @@
+class ClassificationParentNull < ActiveRecord::Migration[5.0]
+  class Classification < ActiveRecord::Base
+  end
+
+  def up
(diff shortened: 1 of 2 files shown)


cben:
Kudos for not giving up on simplifying classifications! :bowing_man:

LGTM.  Needs matching change in core of course.

kbrock:
@Fryguy do you want the code to handle the parent_id = 0 case, or do we assume the people have run the migrations? (I'm thinking we should handle parent_id = 0)

e.g.:

```ruby
class Classification
  scope :is_category, -> { where(:parent_id => [0, nil]) }
end
```

Currently it only supports `:parent_id => 0`. The alternative would be to change this to `:parent_id => nil`.

Fryguy:
Once the migration is merged we can safely assume that the user has run through it.  The can't use a release of the code without migrating first.

PR #319, opened 2019-01-08
Rename WebsocketWorker to RemoteConsoleWorker

skateman:
Renaming both the role and the worker, dropping old workers from the DB.

Parent issue: https://github.com/ManageIQ/manageiq/issues/18300
Core PR: https://github.com/ManageIQ/manageiq/pull/18337

@miq-bot add_label hammer/no

### DIFF
--- a/db/migrate/20190108135546_rename_websocket_to_remote_console.rb
+++ b/db/migrate/20190108135546_rename_websocket_to_remote_console.rb
@@ -1,0 +1,61 @@
+class RenameWebsocketToRemoteConsole < ActiveRecord::Migration[5.0]
+  class SettingsChange < ActiveRecord::Base; end
+
+  class ServerRole < ActiveRecord::Base; end
+
+  class MiqWorker < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    rename_column :miq_servers, :has_active_websocket, :has_active_remote_console
+
+    say_with_time("Renaming the websocket server role to remote_console") do
+      ServerRole.where(:name => 'websocket').update(:name => 'remote_console', :description => "Remote Consoles")
+    end
+
+    say_with_time("Renaming the /log/level_websocket key to /log/level_remote_console in SettingsChange") do
+      SettingsChange.where(:key => '/log/level_websocket').update(:key => '/log/level_remote_console')
+    end
+
+    say_with_time("Renaming all websocket_worker related SettingsChange keys to remote_console_worker") do
+      SettingsChange.where("key LIKE '/workers/worker_base/websocket_worker/%'").select('id, key').each do |row|
+        row.update(:key => row.key.sub('websocket_worker', 'remote_console_worker'))
+      end
+    end
+
+    say_with_time("Renaming all websocket server role SettingsChange records to remote_console") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%websocket%'").select('id', 'value').each do |row|
+        row.update(:value => row.value.sub('websocket', 'remote_console'))
+      end
+    end
+
+    say_with_time("Removing all MiqWorker records related to WebsocketWorker instances") do
+      MiqWorker.where(:type => 'MiqWebsocketWorker').delete_all
+    end
+  end
+
+  def down
+    rename_column :miq_servers, :has_active_remote_console, :has_active_websocket
+
+    say_with_time("Renaming the remote_console server role to websocket") do
+      ServerRole.where(:name => 'remote_console').update(:name => 'websocket', :description => "Websocket")
+    end
+
+    say_with_time("Renaming the /log/level_remote_console key to /log/level_websocket in SettingsChange") do
+      SettingsChange.where(:key => '/log/level_remote_console').update(:key => '/log/level_websocket')
+    end
+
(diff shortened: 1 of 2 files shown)


jrafanie:
> I think there might be other migrations needed...

👍 

See: https://github.com/ManageIQ/manageiq-schema/pull/320

It's fine to just remove the worker rows on upgrade to the version where you renamed the class.

spec/migrations/20190108135546_rename_websocket_to_remote_console_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    it "converts the log levels" do
+      setting_changed = settings_change_stub.create!(:key => "/log/level_websocket", :value => "foo")
+      setting_ignored = settings_change_stub.create!(:key => "/log/level_api", :value => "bar")
+
+      migrate
+
+      expect(setting_changed.reload.key).to eq("/log/level_remote_console")
+      expect(setting_ignored.reload.key).to eq("/log/level_api")
```
jrafanie: I feel like these settings changes should be testing the key and value to verify they're still the same.  Someone might have set level_websocket to debug and would want to ensure the renamed key is still debug.

skateman: Is it enough if I check `setting_changed.value` and `setting_ignored.value`? Or do I have to do some extra querying as well and if yes how?

jrafanie: Yes, I think if you add a check for the `.value`, that's more than enough

db/migrate/20190108135546_rename_websocket_to_remote_console.rb
```diff
@@ -0,0 +1,63 @@
...
+      ServerRole.where(:name => 'websocket').update(:name => 'remote_console', :description => "Remote Consoles")
+    end
+
+    say_with_time("Renaming the /log/level_websocket key to /log/level_remote_console in SettingsChange") do
+      SettingsChange.where(:key => '/log/level_websocket').update(:key => '/log/level_remote_console')
+    end
+
+    say_with_time("Renaming all websocket_worker related SettingsChange keys to remote_console_worker") do
+      SettingsChange.where("key LIKE '/workers/worker_base/websocket_worker/%'").select('id, key').each do |row|
+        SettingsChange.where(:id => row.id).update(:key => row.key.sub('websocket_worker', 'remote_console_worker'))
```
jrafanie: Can't we do this without this second `where`?  We already have the row.  Can we remove the `.select` on line 24 and perform the update on the `row`?

db/migrate/20190108135546_rename_websocket_to_remote_console.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+    end
+
+    say_with_time("Renaming all websocket server role SettingsChange records to remote_console") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%websocket%'").select('id', 'value').each do |row|
+        SettingsChange.where(:id => row.id).update(:value => row.value.sub('websocket', 'remote_console'))
```
jrafanie: Same as above

db/migrate/20190108135546_rename_websocket_to_remote_console.rb
```diff
@@ -0,0 +1,63 @@
...
+      ServerRole.where(:name => 'remote_console').update(:name => 'websocket', :description => "Websocket")
+    end
+
+    say_with_time("Renaming the /log/level_remote_console key to /log/level_websocket in SettingsChange") do
+      SettingsChange.where(:key => '/log/level_remote_console').update(:key => '/log/level_websocket')
+    end
+
+    say_with_time("Renaming all remote_console_worker related SettingsChange keys to websocket_worker") do
+      SettingsChange.where("key LIKE '/workers/worker_base/remote_console_worker/%'").select('id, key').each do |row|
+        SettingsChange.where(:id => row.id).update(:key => row.key.sub('remote_console_worker', 'websocket_worker'))
```
jrafanie: same

db/migrate/20190108135546_rename_websocket_to_remote_console.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+    end
+
+    say_with_time("Renaming all remote_console server role SettingsChange records to websocket") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%remote_console%'").select('id', 'value').each do |row|
+        SettingsChange.where(:id => row.id).update(:value => row.value.sub('remote_console', 'websocket'))
```
jrafanie: ditto

jrafanie:
This looks like a great approach!  I had a few questions above.

db/migrate/20190108135546_rename_websocket_to_remote_console.rb
skateman: Oh yes :laughing: I was thinking too much in SQL :sweat_smile:

skateman:
Okay, who is brave enough to merge this with its core counterpart as well? @Fryguy @agrare @bdunne @carbonin ?

jrafanie:
🎉 🏃

PR #324, opened 2019-01-17
Add multiattach field to CloudVolume

mansam:
Adds a "multiattach" boolean column to the CloudVolume model
https://bugzilla.redhat.com/show_bug.cgi?id=1660896

### DIFF
--- a/db/migrate/20190110201414_add_multi_attachment_to_cloud_volume.rb
+++ b/db/migrate/20190110201414_add_multi_attachment_to_cloud_volume.rb
@@ -1,0 +1,5 @@
+class AddMultiAttachmentToCloudVolume < ActiveRecord::Migration[5.0]
+  def change
+    add_column :cloud_volumes, :multi_attachment, :boolean
+  end
+end


db/migrate/20190110201414_add_multiattach_to_cloud_volume.rb
```diff
@@ -0,0 +1,5 @@
+class AddMultiattachToCloudVolume < ActiveRecord::Migration[5.0]
+  def change
+    add_column :cloud_volumes, :multiattach, :boolean
```
agrare: Hm `multiattach` looks overly shortened when it doesn't need to be, how about `multi_attach` or `multiple_attachment`?

aufi: Correct, but it matches to OpenStack keywords - https://docs.openstack.org/cinder/latest/admin/blockstorage-volume-multiattach.html

agrare: True but this is provider common code so matching any specific provider isn't as important as it matching the rest of the platform consistently.

mansam: @agrare multi_attach or multiple_attachment would suit me fine, do you have a preference?

agrare: @mansam I think `multi_attachment` would be good, I'm bad at naming :)

PR #326, opened 2019-01-22
Split up Storages to be unique per EMS and ems_ref

agrare:
Currently storages are not owned by any single EMS and can actually be shared by multiple providers (of different types even).

This is nice for identifying the underlying physical storage that might be shared by a few providers but it presents a few challenges.
A lot of properties on a storage record are actually hypervisor-level properties and can be overwritten by each refresh.  For example ems_ref, name, master, etc... can be different on two different providers that the storage is added to.

The solution is to create a new storage record for each [ems_id, ems_ref] pair so that storages become just like any other virtual record.

The multiple-ems_ref-per-storage issue was worked-around by dumping the ems_ref into the host_storages table (https://bugzilla.redhat.com/show_bug.cgi?id=1280402).  This worked for the most critical property but it isn't feasible to copy every datastore-type property into host_storages.

We can use the ems_ref from host_storages to get a list of all of the new storage records that we would need.  Once these storage records are created we can update host_storages to point to the new storage.

There are two BZs that this is related to,
https://bugzilla.redhat.com/show_bug.cgi?id=1394909 - This one needs to store if a datastore is in maintenance mode or not, since a datastore can be added to multiple providers it is possible for it to be in maintenance mode on one VC and not another causing race conditions with provisioning depending on which EMS was refreshed last.
https://bugzilla.redhat.com/show_bug.cgi?id=1668020 - This one needs to store the status of the datastore (e.g. is it disconnected), again this can be okay on one VC and not okay on another leading to race conditions.

Core PR: https://github.com/ManageIQ/manageiq/pull/19617
Cross Repo Tests: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/33

### DIFF
--- a/db/migrate/20190729170013_split_storages_per_ems.rb
+++ b/db/migrate/20190729170013_split_storages_per_ems.rb
@@ -1,0 +1,62 @@
+class SplitStoragesPerEms < ActiveRecord::Migration[5.0]
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class HostStorage < ActiveRecord::Base
+    belongs_to :host,    :class_name => "::SplitStoragesPerEms::Host"
+    belongs_to :storage, :class_name => "::SplitStoragesPerEms::Storage"
+  end
+
+  class Storage < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    Storage.in_my_region.find_each do |storage|
+      # First we need to group HostStorage records by their ems_id and ems_ref since these will
+      # form the new Storage records. Once the Storage records are created we can link them
+      # to the hosts through these HostStorage records.
+      host_storages_by_ems_id_and_ems_ref = HostStorage.includes(:host).where(:storage_id => storage.id).group_by do |hs|
+        ems_id            = hs.host&.ems_id
+        datastore_ems_ref = hs.ems_ref || storage.ems_ref
+
+        next if ems_id.nil?
+
+        [ems_id, datastore_ems_ref]
+      end
+
+      storages_by_ems_id_and_ems_ref = {}
+      all_ems_id_and_ems_refs        = host_storages_by_ems_id_and_ems_ref.keys
+
+      # Re-use the existing storage record
+      storage.ems_id, storage.ems_ref = all_ems_id_and_ems_refs.shift
+      storage.save!
+      storages_by_ems_id_and_ems_ref[[storage.ems_id, storage.ems_ref]] = storage
+
+      # Create new storages for the rest of the ems_ids and ems_refs
+      all_ems_id_and_ems_refs.each do |ems_id, ems_ref|
+        next if ems_id.nil?
+
+        new_storage = storage.attributes.except("id").merge("ems_id" => ems_id, "ems_ref" => ems_ref)
+        storages_by_ems_id_and_ems_ref[[ems_id, ems_ref]] = Storage.create!(new_storage)
+      end
+
+      # Link up the host_storage records to the new storages
+      host_storages_by_ems_id_and_ems_ref.each_key do |ems_id, ems_ref|
+        # Connect all of the HostStorage records to the new Storage record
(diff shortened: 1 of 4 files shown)


himdel:
I'm getting...

```
== 20190729170013 SplitStoragesPerEms: migrating ==============================
rake aborted!

undefined method `map' for nil:NilClass
/home/himdel/manageiq/lib/extensions/ar_migration.rb:88:in `migrate'
/home/himdel/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/home/himdel/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'

Caused by:
NoMethodError: undefined method `map' for nil:NilClass
/home/himdel/manageiq/lib/extensions/ar_migration.rb:88:in `migrate'
/home/himdel/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/home/himdel/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
```

agrare:
@himdel can you send me your database?

That is looping over the hash keys so it'd have to be `hash[key] = nil`

db/migrate/20190729170013_split_storages_per_ems.rb
```diff
@@ -0,0 +1,62 @@
...
+  def up
+    Storage.in_my_region.find_each do |storage|
+      # First we need to group HostStorage records by their ems_id and ems_ref since these will
+      # form the new Storage records. Once the Storage records are created we can link them
+      # to the hosts through these HostStorage records.
+      host_storages_by_ems_id_and_ems_ref = HostStorage.includes(:host).where(:storage_id => storage.id).group_by do |hs|
+        ems_id            = hs.host&.ems_id
+        datastore_ems_ref = hs.ems_ref || storage.ems_ref
+
+        next if ems_id.nil?
```
himdel: *possibly* this `next` should happen in a `map` and not a `group_by`?

agrare: Yeah agreed

himdel:
This is my ` HostStorage.includes(:host).where(:storage_id => storage.id)`:

```
#<ActiveRecord::Relation [
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 1, read_only: false, id: 1, ems_ref: "datastore-874", accessible: nil>,
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 2, read_only: false, id: 3, ems_ref: "datastore-874", accessible: nil>,
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 3, read_only: false, id: 5, ems_ref: "datastore-874", accessible: nil>,
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 22, read_only: false, id: 192, ems_ref: "datastore-11", accessible: nil>,
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 23, read_only: false, id: 194, ems_ref: "datastore-11", accessible: nil>,
  #<SplitStoragesPerEms::HostStorage storage_id: 1, host_id: 24, read_only: false, id: 196, ems_ref: "datastore-11", accessible: nil>]>
]
```

and `Host.find([1,2,3,22,23,24]).pluck(:ems_id)` yields `[nil, nil, nil, 55, 55, 55]`.

agrare:
:+1: I have a test for an archived host, let me see why that's not catching this...

spec/migrations/20190729170013_split_storages_per_ems_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+        it "links the right host to the right datastore" do
+          migrate
+
+          expect(host_1.reload.storages.first.ems_id).to eq(host_1.ems_id)
+          expect(host_2.reload.storages.first.ems_id).to eq(host_2.ems_id)
+        end
+
+        context "with an archived host" do
+          let(:archived_host) do
```
himdel: It's missing `!`, the host never gets created..

And then, the syntax error, as `ems_ref` is undefined :)

agrare: Yes, see https://github.com/ManageIQ/manageiq-schema/pull/444

PR #327, opened 2019-01-22
Use views for the metrics and metric_rollups tables

carbonin:
This will allow us to use an "INSTEAD OF" trigger to insert the
rows into the sub-tables rather than having to insert and remove
the rows from the base table. This reduces overhead as well as
eliminates table bloat from deleted rows on the base tables.

This approach also preserves the inheritance structure which
continues to allow us to take advantage of the query optomization
from the check constraints on the sub-tables. (This would not
have been possible if we removed the "_base" tables entirely).

This also requires a small change to the models to specify the
primary key. https://github.com/ManageIQ/manageiq/pull/18384

### DIFF
--- a/db/migrate/20190122213042_use_views_for_metrics.rb
+++ b/db/migrate/20190122213042_use_views_for_metrics.rb
@@ -1,0 +1,162 @@
+class UseViewsForMetrics < ActiveRecord::Migration[5.0]
+  include MigrationHelper
+
+  def up
+    # Drop the triggers
+    drop_trigger "metrics", "metrics_inheritance_before"
+    drop_trigger "metrics", "metrics_inheritance_after"
+    drop_trigger "metric_rollups", "metric_rollups_inheritance_before"
+    drop_trigger "metric_rollups", "metric_rollups_inheritance_after"
+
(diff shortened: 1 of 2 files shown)


carbonin:
Almost forgot about this :+1: @Fryguy @gtanzillo anyone want to pull the trigger on these :laughing: :running_man:

juliancheal:
@carbonin MiQ-bot looks unhappy :trollface:

carbonin:
@Fryguy any more thoughts on this one?

Fryguy:
To get rid of the rubocop issues, we should update the global one on manageiq guides to have these DSL methods listed.

juliancheal:
Seem to be getting

```
 ActiveRecord::StatementInvalid:
       PG::CheckViolation: ERROR:  new row for relation "metric_rollups_03" violates check constraint "metric_rollups_03_inheritance_check"
```

In travis tests, wondering if it's related.

juliancheal:
> To get rid of the rubocop issues, we should update the global one on manageiq guides to have these DSL methods listed.

You sub-commenting me there @Fryguy :)

PR #332, opened 2019-01-29
[WIP] Add provision state to service

d-m-u:
Tina, Loic, and Dennis say we need another state field in the service object to indicate the provision state.

It's WIP for reasons? Because we haven't seemed to have had enough discussions around this to know what's happening with it? Something?


d-m-u:
Haha no but in lieu of a bz link or any kind of actual documentation that this conversation happened I thought it prudent to apportion blame early (and often.)

d-m-u:
Closing because I think this got changed to be a more generalized concept than just a provisioning state and the schema name should reflect that.

PR #336, opened 2019-02-14
Add comments to the miq_tasks table

djberg96:
In my continuing effort to understand things better by documenting them, this PR adds comments for the columns of the `miq_tasks` table.

WIP for now until I can clarify and/or confirm some things:

~~- identifier - what's this?~~ (solved)
~~- results - what's this?~~ (old column, no longer used)
~~- zone - why isn't this `zone_id` with a `belongs_to` on the `zones` table?~~ (short answer: not needed and/or historical reasons)
~~- userid - confirm that 'system' is always used for internal tasks.~~ (current comment seems to be valid)


djberg96:
@jrafanie @agrare @Fryguy @bdunne Please take a look.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "The timestamp the task was added to the app inventory."
+    change_column_comment :miq_tasks, :id, "The internal database ID."
+    change_column_comment :miq_tasks, :identifier, "TODO"
```
agrare: This is used (exclusively I think) by metrics rollups for EmsClusters [ref](https://github.com/ManageIQ/manageiq/blob/master/app/models/metric/capture.rb#L170).  Identifier is `"#{parent.class}:#{parent.id}"` [ref](https://github.com/ManageIQ/manageiq/blob/master/app/models/metric/capture.rb#L152)

djberg96: Thanks, updated.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "The timestamp the task was added to the app inventory."
```
yrudman: I do not think we are updating this column explicitly, do we need to comment on default behavior ?

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :id, "The internal database ID."
+    change_column_comment :miq_tasks, :identifier, "The identifier used by metrics rollups for EMS clusters."
```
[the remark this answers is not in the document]
agrare: I don't think it is intended to be only for metric rollups, just happens to be only used there.

More accurately might be described as "The resource this task is related to, format `"#{klass}:#{id}"`.  Honestly this would make more sense as a polymorphic resource than a string column that jams the class+id together.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: It doesn't hurt. On a side note, I've noticed minor differences between tables, e.g. `created_at` vs `created_on`. Does rails treat this as synonyms?

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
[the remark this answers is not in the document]
djberg96: @bdunne @agrare Ok, I've updated it.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "The timestamp the task was added to the app inventory."
+    change_column_comment :miq_tasks, :id, "The internal database ID."
+    change_column_comment :miq_tasks, :identifier, "The resource this task is related to in '#{klass}:#{id}' format."
+    change_column_comment :miq_tasks, :message, "Message that describes the reason for the current status."
```
yrudman: this message shown in UI on task management screen

yrudman: also, message could be updated independently from status

djberg96: Many fields are shown within the UI, so I'm not sure I want to add that note explicitly here. As for the message contents, how you would you more generically describe it then?

yrudman: I think this column `message` should have only one responsibility (compare to other many field) - to be shown in log or in UI  (however some may use it for different purposes)

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
bdunne: Yes, Rails recently switched from `created_on` to `created_at`.  Maybe we should add a prefix like `Builtin` or more specifically `Rails`?  Then it could say `"Builtin: the timestamp that the record was created"`

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :miq_server_id, "ID of the ManageIQ server that the task belongs to."
```
yrudman: it is id of server where task was executed

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :name, "The name of the task. Typically doubles as a description."
```
yrudman: there is no `description` column in `miq_task` table, I am not sure where name of task doubles as a description

djberg96: In practice what I'm seeing is stuff like "Generate Widget: 'EVM: Recently Discovered Hosts'". That's really more of a description than a name, at least as I typically think of it.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, updated.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Well, we haven't done that in the past, and it wouldn't make sense to other applications that might be referencing this table (if that ever happens). I'd like to just leave it if it's all the same.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Alright how does, "Message associated with the task for logging or UI purposes" sound?

yrudman: 👍

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "The timestamp the task was added to the app inventory."
+    change_column_comment :miq_tasks, :id, "The internal database ID."
+    change_column_comment :miq_tasks, :identifier, "The resource this task is related to in '#{klass}:#{id}' format."
+    change_column_comment :miq_tasks, :message, "Message that describes the reason for the current status."
+    change_column_comment :miq_tasks, :miq_server_id, "The ID of the ManageIQ server where the task was executed."
+    change_column_comment :miq_tasks, :name, "The name of the task. Typically doubles as a description."
+    change_column_comment :miq_tasks, :pct_complete, "The current completion percentage of the task."
+    change_column_comment :miq_tasks, :results, "TODO"
```
yrudman: it is not in use anymore

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, updated.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: I can mark it deprecated. Any idea what it was originally used for?

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :started_on, "The timestamp the task was actually started."
+    change_column_comment :miq_tasks, :state, "The current state of the task - Initialized, Queued, Active, or Finished."
```
yrudman: There are more possible states and other states could be added later on (vs status with has more-less frozen possible values). 
screen shot 2019-02-14 at 12 01 47 pm

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
bdunne: We don't have much of a precedent yet (one other table documented).
I agree with @yrudman that it is different from other columns in the table in that our app doesn't "own" (and shouldn't change) this column, so our documentation should probably reflect that in some way.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Hm, I was looking at miq_task.rb. Those must be added in subclasses somewhere.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :status, "The current status of the task - Ok, Warn, Error, Timeout, Expired, or Unknown."
+    change_column_comment :miq_tasks, :updated_on, "The timestamp the task was last updated within the appliance."
+    change_column_comment :miq_tasks, :userid, "The userid that created the task, or 'system' if created by the appliance."
+    change_column_comment :miq_tasks, :zone, "The zone of the request that created the task."
```
yrudman: it is `zone` in which task should be executed, if `nil` than any zone can execute this task

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Well, there's also the vms table from a short while back. Ok, I'll start tagging those columns with "builtin".

djberg96: Ok, tagged.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, updated so that those are just provided as examples.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Updated, thanks.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "[builtin] The timestamp the task was added to the app inventory."
+    change_column_comment :miq_tasks, :id, "The internal database ID."
```
yrudman: `id` column also should be marked as `[buitin]` the internal id of record

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "[builtin] The timestamp the task was added to the app inventory."
```
yrudman: would be more  clear if it say "...record created"

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
yrudman: Ohh, you meant it usually used to put descriptive name

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Done!

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Done!

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: At least some of our tables have separate `name` and `description` fields, but here it seems to be used interchangeably.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: I've marked it deprecated. Let me know if you want to add more to it.

yrudman: it was used to keep result task execution

djberg96: Ok, made an update, thanks.

djberg96:
I think the last question I had was the `userid` field and if my comment is correct.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
+class AddCommentsToMiqTasksTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :miq_tasks, "ManageIQ background tasks"
+
+    change_column_comment :miq_tasks, :context_data, "Serialized metadata that may contain more detailed information about the task."
+    change_column_comment :miq_tasks, :created_on, "[builtin] The timestamp the record was added to the app inventory."
```
yrudman: "record was added to the app inventory" sounds confusing to me, it implies that it is different than just timestamp when record created.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :id, "[builtin] The internal database ID."
```
yrudman: it should be "... record id", may be specify that this column used as the primary key

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
```diff
@@ -0,0 +1,21 @@
...
+    change_column_comment :miq_tasks, :identifier, "The resource this task is related to in 'class:id' format."
+    change_column_comment :miq_tasks, :message, "Message associated with the task for logging or UI purposes."
+    change_column_comment :miq_tasks, :miq_server_id, "The ID of the ManageIQ server where the task was executed."
+    change_column_comment :miq_tasks, :name, "The name of the task. Typically doubles as a description."
+    change_column_comment :miq_tasks, :pct_complete, "The current completion percentage of the task."
+    change_column_comment :miq_tasks, :results, "Task execution result. Deprecated."
+    change_column_comment :miq_tasks, :started_on, "The timestamp the task was actually started."
+    change_column_comment :miq_tasks, :state, "The current state of the task - Initialized, Queued, Active, Finished, etc."
+    change_column_comment :miq_tasks, :status, "The current status of the task - Ok, Warn, Error, Timeout, Expired, or Unknown."
+    change_column_comment :miq_tasks, :updated_on, "[builtin] The timestamp the record was last updated within the appliance."
```
yrudman: I would remove "within the appliance"

yrudman:
`userid` description sounds good

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, done.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, done.

db/migrate/20190214125845_add_comments_to_miq_tasks_table.rb
djberg96: Ok, updated.

yrudman:
👍 comments looks good to me

PR #337, opened 2019-02-18
Create a new migration to fix special characters in the rabbitmq password

carbonin:
Previously the arguments to the find_by call were incorrect so
the migration would never find a record. The issue is that the
userid was "tower" not "ansible".

In the new version we only search by the type and authtype attributes
which should never change and are clearly defined in the application code

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1678337

### DIFF
--- a/db/migrate/20181012160010_remove_special_characters_from_ansible_rabbitmq_password.rb
+++ b/db/migrate/20181012160010_remove_special_characters_from_ansible_rabbitmq_password.rb
@@ -1,25 +1,3 @@
-require 'securerandom'
-
 class RemoveSpecialCharactersFromAnsibleRabbitmqPassword < ActiveRecord::Migration[5.0]
-  # used only in specs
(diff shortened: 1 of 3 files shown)


carbonin:
There is a workaround in the bug for people who have already migrated past the broken migration. When they run those steps it will make the new migration a no-op as the password will not contain special characters.

simaishi:
Hammer backport details:

```
$ git log -1
commit 6a02bbf2746253330090106579507ddf124bc42b
Date:   Mon Feb 18 16:10:36 2019 -0500

    Merge pull request #337 from carbonin/remove_special_characters_from_ansible_rabbitmq_two
    
    Create a new migration to fix special characters in the rabbitmq password
    
    (cherry picked from commit 39121487608e8336c199c135389ba5f7ebd27284)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1678489
```

PR #338, opened 2019-02-19
[WIP] Add comments to the authentications table

djberg96:
In my continuing effort to understand things better by documenting them, this PR adds comments for the columns of the `authentications` table.

WIP for now as there are many columns I know nothing about, and some of the current ones I'm guessing on.

### DIFF
--- a/db/migrate/20190219180039_add_comments_to_authentications_table.rb
+++ b/db/migrate/20190219180039_add_comments_to_authentications_table.rb
@@ -1,0 +1,58 @@
+class AddCommentsToAuthenticationsTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :authentications, "Authentication information for various resources"
+
+    change_column_comment :authentications, :id, "[builtin] The internal record ID. This is the primary key."
+    change_column_comment :authentications, :name, "Symbolic name for the authentication"
+    change_column_comment :authentications, :authtype, "The authentication type, e.g. ssh_keypair, default, etc."
+    change_column_comment :authentications, :userid, "The userid used in a userid + password authentication approach."
+    change_column_comment :authentications, :password, "The password for the given userid, if any. Encrypted."
+    change_column_comment :authentications, :resource_id, "The ID of the associated resource."
+    change_column_comment :authentications, :resource_type, "The STI type of the associated resource."
+    change_column_comment :authentications, :created_on, "[builtin] The timestamp the record was created."
+    change_column_comment :authentications, :updated_on, "[builtin] The timestamp the record was last updated."
+    change_column_comment :authentications, :last_valid_on, "The timestamp this authentication was last used successfully."
+    change_column_comment :authentications, :last_invalid_on "The timestamp this authentication last failed."
+    change_column_comment :authentications, :credentials_changed_on, "The timestamp that credentials were last changed."
+    change_column_comment :authentications, :status, "The current status of the authentication, e.g. valid, invalid, incomplete, etc."
+    change_column_comment :authentications, :status_details, "Detailed information regarding the current status."
+    change_column_comment :authentications, :type, "The STI type of the authentication."
+    change_column_comment :authentications, :auth_key, "The private ssh key associated with the resource. Encrypted."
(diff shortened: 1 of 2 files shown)


db/migrate/20190219180039_add_comments_to_authentications_table.rb
```diff
@@ -0,0 +1,58 @@
+class AddCommentsToAuthenticationsTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :authentications, "Authentication information for various resources."
+
+    change_column_comment :authentications, :id, "[builtin] The internal record ID. This is the primary key."
+    change_column_comment :authentications, :name, "Symbolic name for the authentication"
+    change_column_comment :authentications, :authtype, "The authentication type or role, e.g. ssh_keypair, default, amqp, etc."
```
agrare: > The authentication type or role

Not a comment on the comment ;) But this is my biggest source of confusion.  This is as accurate as I think we can be, it is either the type of authentication (e.g. sometimes it is ssh_keypair or bearer token) and sometimes it is the role (e.g. hawkular, prometheus, metrics).

I would have expected the the type would have been encoded in the STI type and authtype would just be the role.

djberg96:
@jvlcek I don't suppose you could speak to any of these, could you? The ldap columns perhaps?

db/migrate/20190219180039_add_comments_to_authentications_table.rb
```diff
@@ -0,0 +1,58 @@
...
+    change_column_comment :authentications, :service_account
+    change_column_comment :authentications, :challenge
+    change_column_comment :authentications, :login
+    change_column_comment :authentications, :public_key, "An SSH public key."
+    change_column_comment :authentications, :htpassd_users
+    change_column_comment :authentications, :ldap_id
+    change_column_comment :authentications, :ldap_email
+    change_column_comment :authentications, :ldap_name
+    change_column_comment :authentications, :ldap_preferred_user_name
+    change_column_comment :authentications, :ldap_bind_dn
```
jvlcek: bind dn - Is the LDAP identity by which LDAP operations are performed.

db/migrate/20190219180039_add_comments_to_authentications_table.rb
```diff
@@ -0,0 +1,58 @@
...
+    change_column_comment :authentications, :ldap_insecure
```
jvlcek: I am not sure but I assume this is likely a boolean indicating LDAP or LDAPS

db/migrate/20190219180039_add_comments_to_authentications_table.rb
```diff
@@ -0,0 +1,58 @@
...
+    change_column_comment :authentications, :ldap_url
```
jvlcek: Most likely the URL to the LDAP server.

db/migrate/20190219180039_add_comments_to_authentications_table.rb
```diff
@@ -0,0 +1,58 @@
+class AddCommentsToAuthenticationsTable < ActiveRecord::Migration[5.0]
+  def up
+    change_table_comment :authentications, "Authentication information for various resources."
+
+    change_column_comment :authentications, :id, "[builtin] The internal record ID. This is the primary key."
+    change_column_comment :authentications, :name, "Symbolic name for the authentication"
+    change_column_comment :authentications, :authtype, "The authentication type or role, e.g. ssh_keypair, default, amqp, etc."
+    change_column_comment :authentications, :userid, "The userid used in a userid + password authentication approach."
```
jvlcek: The userid is the unique string identifier used in the MiQ database to identify a user record. The format of the userid varies depending on authentication mode but for all external authentication modes it is UPN format.

db/migrate/20190219180039_add_comments_to_authentications_table.rb
jvlcek: @djberg96 I tried to shed some light on some of these but I am not sure on some. If you would like to schedule a meeting I'd be glad to  research them together.
Joe

PR #340, opened 2019-02-25
Add the date of when inventory was collected

agrare:
We have a column tracking when the EMS was last successfully saved
(#last_refresh_date).  This tells us when save_inventory last finished,
but a lot of code wants to know when the inventory which was saved was
collected.

An example of this is "I just added a disk to a VM at time T1, I want to
know when this change is represented in VMDB"

You can wait until last_refresh_date >= T1, but it is possible that
save_inventory took a long time and the inventory was actually
collected at T0 and just saved at T2.

If we had a column that stored when the inventory was collected then no
matter how long the refresh sat in the queue we would know for a fact
that the EMS includes our changes.

Currently this is done by getting a task_id back with the refresh queue
item, but e.g. VMware streaming refresh doesn't use the normal queue for
refreshes so this method won't work.

### DIFF
--- a/db/migrate/20190225142729_add_last_inventory_date_to_ems.rb
+++ b/db/migrate/20190225142729_add_last_inventory_date_to_ems.rb
@@ -1,0 +1,5 @@
+class AddLastInventoryDateToEms < ActiveRecord::Migration[5.0]
+  def change
+    add_column :ext_management_systems, :last_inventory_date, :datetime
+  end
+end


agrare:
This fixes the [`ReconfigVM_Task_Complete` event handler](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/EVM.class/reconfigvm_task_complete.yaml) which uses the [`Refresh` state machine](https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class) to wait for the refresh task to complete when using vmware streaming refresh.

db/migrate/20190225142729_add_last_collected_date_to_ems.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastCollectedDateToEms < ActiveRecord::Migration[5.0]
+  def change
+    add_column :ext_management_systems, :last_collected_date, :datetime
```
Fryguy: Discussion note, _date was used here for consistency with other date columns...would have like this to be called _at otherwise.

Fryguy:
Naming wise last_collected still feels like "after" the fact.  Perhaps `last_refresh_start_date`?  Then at least that pair are tied together as bookends?

agrare:
Yeah I like tying this to the last_refresh_date attribute, only issue I have with `last_refresh_start_date` is that looking at that I would assume that is when the refresh was started, not when the `collect_inventory_for_targets` stage was completed.

High level I'd like this to represent the timestamp for how old the inventory data is that we saved rather than when the save finished.

jrafanie:
Isn't naming fun?

I think the process of collecting or refreshing the inventory "clouds" the meaning of the current suggestions... what a user really wants to know is what inventory we have:

inventory_captured_at
inventory_captured_date
date_of_inventory
time_of_inventory

Throw in _last if you want, such as:
date_of_last_inventory

agrare:
I do like `inventory_captured_at/date`, maybe `last_refresh_inventory_date` to keep it closely related to `last_refresh_date` ?

jrafanie:
do we care about the word capture or refresh?  

`last_inventory_date` tells me the inventory is from that date.  I don't care about the mechanics of capture/refresh, only that the inventory is relevant to that specific date, right?

jrafanie:
naming is hard

agrare:
Okay `last_inventory_date` sounds pretty good, updated.

jrafanie:
note, the old filename and class reference the original idea.  I'm ok with it... but in case it bothered you...

`class AddLastCollectedDateToEms`

agrare:
Good catch updated

agrare:
@Fryguy what do you think of the name?

jrafanie:
He'll have an opinion after I merge it 😉 😉 😉 ... will give a few more hours and then pull the trigger

PR #341, opened 2019-02-27
Migration for checking guid uniqueness

d-m-u:
All the things that have the uuid mixin should validate that the guids generated are unique, I know it's a ridiculously stupid nit, and I know that we are so unlikely to ever find a case in which this migration is necessary (outside of ```dup``` which will be dealt with in the related PR) that I'm better off buying a lottery ticket for the three mill run today or whatever, but I wanted the satisfaction of having dealt with this so... 

TLDR blame Brandon. 

also, ```guidniqueness```?

Related:
https://github.com/ManageIQ/manageiq/pull/18492

There are two commits in this PR because providers currently have no validations on guids, nor are they required to have guids, and so since in the main repo PR we're adding the mixin to providers, which has a guid presence validation, this PR also checks to make sure that all providers do in fact have guids.

### DIFF
--- a/db/migrate/20190226184255_check_guid_uniqueness.rb
+++ b/db/migrate/20190226184255_check_guid_uniqueness.rb
@@ -1,0 +1,54 @@
+class CheckGuidUniqueness < ActiveRecord::Migration[5.0]
+  # several models already have unique constraints on index so we don't have to deal with them in this migration:
+  # vm, miq_set, scan_item, miq_worker, miq_server, miq_event_definition, miq_action, lifecycle event, job, host, ems
+
+  class AutomateWorkspace < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true, :presence => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class MiqAeWorkspace < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class MiqPolicy < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class MiqRegion < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class MiqWidget < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class Service < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class ServiceTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    [AutomateWorkspace, MiqAeWorkspace, MiqPolicy, MiqRegion, MiqWidget, Service, ServiceTemplate].each do |thing|
+      say_with_time("Checking #{thing} for guid uniqueness") do
+        thing.in_my_region.where(:guid => thing.in_my_region.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid)).order(:id).drop(1).each { |obj| obj.update_attributes!(:guid => SecureRandom.uuid) }
+      end
+    end
+  end
+end
--- a/db/migrate/20190304192641_add_provider_guids.rb
+++ b/db/migrate/20190304192641_add_provider_guids.rb
@@ -1,0 +1,11 @@
+class AddProviderGuids < ActiveRecord::Migration[5.0]
+  class Provider < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    Provider.in_my_region.where(:guid => nil).each { |provider| provider.update_attributes!(:guid => SecureRandom.uuid) }
+  end
+end
--- a/spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
+++ b/spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
@@ -1,0 +1,34 @@
+require_migration
+
+describe CheckGuidUniqueness do
+  let(:automate_workspace) { migration_stub(:AutomateWorkspace) }
+  let(:miq_ae_workspace) { migration_stub(:MiqAeWorkspace) }
+  let(:miq_policy) { migration_stub(:MiqPolicy) }
+  let(:miq_region) { migration_stub(:MiqRegion) }
+  let(:miq_widget) { migration_stub(:MiqWidget) }
+  let(:service) { migration_stub(:Service) }
+  let(:service_template) { migration_stub(:ServiceTemplate) }
+
+  migration_context :up do
+    %i[automate_workspace miq_ae_workspace miq_policy miq_region miq_widget service service_template].each do |obj|
+      it "#{obj} dup resets guid" do
+        guid = SecureRandom.uuid
+        klass = send(obj)
+        obj1 = klass.create!(:guid => guid)
+        obj2 = klass.new(:guid => guid)
+        obj2.save(:validate => false)
+        obj3 = klass.create!(:guid => SecureRandom.uuid)
+        guid2 = obj3.guid
+
+        expect(obj1.reload.guid).to eq(guid)
+        expect(obj2.reload.guid).to eq(guid)
+
+        migrate
+
+        expect(obj1.reload.guid).to eq(guid)
+        expect(obj2.reload.guid).not_to eq(guid)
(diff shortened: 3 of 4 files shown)


d-m-u:
@miq-bot assign @bdunne

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+       miq_region,
+       miq_server,
+       miq_widget,
+       miq_worker,
+       scan_item,
+       service,
+       service_template,
+       miq_set,
+       vm].each do |obj|
+        p obj.to_s
```
bdunne: Please remove debugging statements

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        obj.create!.dup.save(:validate => false)
```
bdunne: What do you think of something more like this?

```ruby
guid = SecureRandom.uuid

obj_1 = obj.create!(:guid => guid)
obj_2 = obj.dup.save(:validate => false)

expect(obj_1.reload.guid).to eq(guid)
expect(obj_2.reload.guid).to eq(guid)

migrate

expect(obj_1.reload.guid).to eq(guid)
expect(obj_2.reload.guid).not_to eq(guid)
```

db/migrate/20190226184255_check_guid_uniqueness.rb
```diff
@@ -0,0 +1,106 @@
...
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    [AutomateWorkspace, ExtManagementSystem, Host, Job, LifecycleEvent, MiqAction, MiqAeWorkspace, MiqEventDefinition,
+     MiqPolicy, MiqRegion, MiqServer, MiqWidget, MiqWorker, ScanItem, Service, ServiceTemplate,
+     MiqSet, ScanItem, Vm].each do |thing|
+      say_with_time("Checking #{thing} for guid uniqueness") do
+        thing.where(:guid => thing.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid))[0...-1].each { |obj| obj.guid = SecureRandom.uuid; obj.save! }
```
bdunne: I don't think you want the `pluck` in here, right?

bdunne: Also, the `[0...-1]` will leave out the last one.  I think we would actually want to leave the first record alone since it would be the original. (we may want to `order(:id)` to guarantee that)

bdunne: `.each { |obj| obj.guid = SecureRandom.uuid; obj.save! }` could be simplified to `.each { |obj| obj.update_attributes!(:guid => SecureRandom.uuid) }`

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        expect(obj.where(:guid => obj.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid)).count).to eq(2)
+        migrate
+        expect(obj.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid).count).to eq(0)
+      end
+    end
+
+    it "automate workspace dup resets guid" do
+      automate_workspace.create!(:guid => SecureRandom.uuid).dup.save(:validate => false)
```
bdunne: I think this can be covered by the test above, right?

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
d-m-u: I left this in because it's nice to see what object it's testing, is there a way of doing this that is better?

bdunne: Yes, the `[].each` can be moved up one level of nesting.

```ruby
migration_context :up do
  [things...].each do |test_subject|
    it "#{thing.name} guids are unique" do
      ...the test logic...
    end
  end
end
```

bdunne:
@carbonin Can you take a look too? :eyes:

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:service) { migration_stub(:Service) }
+  let(:service_template) { migration_stub(:ServiceTemplate) }
+
+  migration_context :up do
+    %i(automate_workspace miq_ae_workspace miq_policy miq_region miq_widget service service_template).each do |obj|
+      it "#{obj} dup resets guid" do
+        guid = SecureRandom.uuid
+        klass = send(obj)
+        obj1 = klass.create!(:guid => guid)
+        obj2 = klass.new(:guid => guid)
```
carbonin: Can you also create a third object with a unique guid to ensure it isn't updated?

db/migrate/20190226184255_check_guid_uniqueness.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  class ServiceTemplate < ActiveRecord::Base
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    [AutomateWorkspace, MiqAeWorkspace, MiqPolicy, MiqRegion, MiqWidget, Service, ServiceTemplate].each do |thing|
+      say_with_time("Checking #{thing} for guid uniqueness") do
+        thing.where(:guid => thing.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid)).order(:id).drop(1).each { |obj| obj.update_attributes!(:guid => SecureRandom.uuid) }
```
bdunne: Also, this should probably scope to `in_my_region` to avoid updating records from other regions

db/migrate/20190226184255_check_guid_uniqueness.rb
```diff
@@ -0,0 +1,54 @@
...
+  class ServiceTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    [AutomateWorkspace, MiqAeWorkspace, MiqPolicy, MiqRegion, MiqWidget, Service, ServiceTemplate].each do |thing|
+      say_with_time("Checking #{thing} for guid uniqueness") do
+        thing.in_my_region.where(:guid => thing.select(:guid).group(:guid).having("count(*) > 1").pluck(:guid)).order(:id).drop(1).each { |obj| obj.update_attributes!(:guid => SecureRandom.uuid) }
```
bdunne: The `select` for finding the duplicates also needs the `in_my_region` scope, right?

spec/migrations/20190226184255_check_guid_uniqueness_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        guid2 = obj3.guid
+
+        expect(obj1.reload.guid).to eq(guid)
+        expect(obj2.reload.guid).to eq(guid)
+
+        migrate
+
+        expect(obj1.reload.guid).to eq(guid)
+        expect(obj2.reload.guid).not_to eq(guid)
+        expect(obj3.guid).to eq(guid2)
```
carbonin: I think you need a reload here for this to be valid, right?

db/migrate/20190304192641_add_provider_guids.rb
```diff
@@ -0,0 +1,11 @@
+class AddProviderGuids < ActiveRecord::Migration[5.0]
+  class Provider < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    validates :guid, :uniqueness => true
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    Provider.where(:guid => nil).each { |provider| provider.update_attributes!(:guid => SecureRandom.uuid) }
```
carbonin: This looks like it's solving a different problem than uniqueness ... Can this be a separate PR or are they somehow related?

carbonin: Also you added the mixin for regions, but are not using `.in_my_region` here ...

d-m-u:
@jrafanie can I please get eyes on this?

djberg96:
Just poking my nose in where it doesn't belong...

Postgres has supported a uuid data type since 8.3. Is that something we could leverage here instead of relying on AR?

jrafanie:
> Postgres has supported a uuid data type since 8.3. Is that something we could leverage here instead of relying on AR?

@djberg96 database agnosticism originally... now?  I'm not sure.  I guess it depends if changing it to uuid buys us anything considering we already chose a different column type.  Maybe?

djberg96:
@jrafanie It would allow for greater flexibility in terms of input formatting, and you could also setup the schema to autogenerate a default value.

https://www.postgresql.org/docs/10/datatype-uuid.html and https://www.postgresql.org/docs/10/uuid-ossp.html

d-m-u:
@miq-bot assign @Fryguy

bdunne:
@d-m-u Was this resolved by another PR?

spec/migrations/20190304192641_add_provider_guids_spec.rb
```diff
@@ -0,0 +1,18 @@
+require_migration
+
+describe AddProviderGuids do
+  let(:provider) { migration_stub(:Provider) }
+
+  migration_context :up do
+    it "sets provider guids" do
+      obj1 = provider.create!
+      obj1.save(:validate => false)
```
bdunne: Why `:validate => false` here?  Should this have been `dup`'ed?

bdunne:
@carbonin Any other concerns?

PR #343, opened 2019-03-01
Remove RSS table

juliancheal:
Follow on from https://github.com/ManageIQ/manageiq/pull/18478

### DIFF
--- a/db/migrate/20190301174502_remove_rss_feeds.rb
+++ b/db/migrate/20190301174502_remove_rss_feeds.rb
@@ -1,0 +1,19 @@
+class RemoveRssFeeds < ActiveRecord::Migration[5.0]
+  def up
+    drop_table :rss_feeds
+  end
+
+  def down
+    create_table "rss_feeds" do |t|
+      t.string   "name"
+      t.text     "title"
+      t.text     "link"
+      t.text     "description"
+      t.datetime "created_on"
+      t.datetime "updated_on"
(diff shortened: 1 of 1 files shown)


db/migrate/20190301174502_remove_rss_feeds.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveRssFeeds < ActiveRecord::Migration[5.0]
+  def change
+    drop_table :rss_feeds
```
jrafanie: The cop is complaining because this isn't reversible?  If we don't want to put the table in the down migration, maybe we should write an `up` without a `down`.... but why not just recreate it as it is now in the `down`?

Fryguy: The reverse is required.

juliancheal:
> * ❗️ - [Line 7](https://github.com/juliancheal/manageiq-schema/blob/4b4e72aceab1adb53b583dc6e794599974302b4b/db/migrate/20190301174502_remove_rss_feeds.rb#L7), Col 5 - [Rails/CreateTableWithTimestamps](http://rubydoc.info/gems/rubocop/0.52.1/RuboCop/Cop/Rails/CreateTableWithTimestamps) - Add timestamps when creating a new table.

Presumably we can ignore this cop?

juliancheal:
@miq-bot assign @Fryguy

PR #346, opened 2019-03-07
Add provider_services_supported column to Availability Zone

mansam:
This PR adds a `services` string array column to the Availability Zone table. Openstack Availability Zones are visible to particular Openstack services, and this column allows those services to be marked during inventory collection and filtered on in queries. This is a part of the implementation of https://bugzilla.redhat.com/show_bug.cgi?id=1549128.

See also https://github.com/ManageIQ/manageiq-providers-openstack/pull/447


carbonin:
Is there a better name for this? We have a `services` table ... is there any chance people could get confused between the two?

agrare:
Yeah, maybe something like `provider_services` or `provider_services_supported/enabled` ?

mansam:
@agrare thanks for the review, works for me!

PR #347, opened 2019-03-07
Add name validation to service templates

d-m-u:
We added a method to copy service_templates 21 days ago here https://github.com/ManageIQ/manageiq/commit/955f49b080ec742878d9287c218ff364036118d1#diff-9257fff45a7d1258305c9a6b7d15e322R48 and in the process discovered that it's valid for templates to not have names and I think templates should have names.

### DIFF
--- a/db/migrate/20190228210310_service_templates_should_have_names.rb
+++ b/db/migrate/20190228210310_service_templates_should_have_names.rb
@@ -1,0 +1,12 @@
+class ServiceTemplatesShouldHaveNames < ActiveRecord::Migration[5.0]
+  class ServiceTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    say_with_time("Checking service templates for names") do
+      ServiceTemplate.in_my_region.where(:name => [nil, ""]).each { |t| t.update_attributes!(:name => "ServiceTemplate" + SecureRandom.uuid) }
+    end
+  end
+end
--- a/spec/migrations/20190228210310_service_templates_should_have_names_spec.rb
+++ b/spec/migrations/20190228210310_service_templates_should_have_names_spec.rb
@@ -1,0 +1,20 @@
+require_migration
+
+describe ServiceTemplatesShouldHaveNames do
+  let(:service_template) { migration_stub(:ServiceTemplate) }
+
+  migration_context :up do
+    it "sets template_names" do
+      obj1 = service_template.create!
+      obj2 = service_template.create!(:name => "")
+
+      expect(obj1.reload.name).to eq(nil)
+      expect(obj2.reload.name).to eq("")
+
+      migrate
+
+      expect(obj1.reload.name).not_to eq(nil)
+      expect(obj2.reload.name).not_to eq("")
+    end
+  end
+end


d-m-u:
@miq-bot assign @bdunne 
🎉 
it's the gift that keeps on giving.

db/migrate/20190228210310_service_templates_should_have_names.rb
```diff
@@ -0,0 +1,13 @@
+class ServiceTemplatesShouldHaveNames < ActiveRecord::Migration[5.0]
+  class ServiceTemplate < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    ServiceTemplate.where(:name => [nil, ""]).each do |template|
```
bdunne: This should probably be scoped to `in_my_region`.

db/migrate/20190228210310_service_templates_should_have_names.rb
```diff
@@ -0,0 +1,13 @@
...
+      say_with_time("Checking #{template} for name presence") do
+        template.update_attributes!(:name => "ServiceTemplate")
```
bdunne: If the name doesn't need to be unique and they'll all get the save value, then prefer `.update_all` rather than `update_attributes` on each instance.

db/migrate/20190228210310_service_templates_should_have_names.rb
```diff
@@ -0,0 +1,13 @@
+class ServiceTemplatesShouldHaveNames < ActiveRecord::Migration[5.0]
+  class ServiceTemplate < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    ServiceTemplate.where(:name => [nil, ""]).each do |template|
+      say_with_time("Checking #{template} for name presence") do
```
bdunne: The `say_with_time` should wrap the entire operation, not each `update`.

spec/migrations/20190228210310_service_templates_should_have_names_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_migration
+
+describe ServiceTemplatesShouldHaveNames do
+  let(:service_template) { migration_stub(:ServiceTemplate) }
+
+  migration_context :up do
+    it "sets template_names" do
+      obj1 = service_template.create!
+
+      expect(obj1.reload.name).to eq(nil)
```
bdunne: Probably a good idea to have another record that has `name => ""`

db/migrate/20190228210310_service_templates_should_have_names.rb
```diff
@@ -0,0 +1,12 @@
+class ServiceTemplatesShouldHaveNames < ActiveRecord::Migration[5.0]
+  class ServiceTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    say_with_time("Checking service templates have names") do
+      ServiceTemplate.in_my_region.where(:name => [nil, ""]).update_all(:name => "ServiceTemplate" + SecureRandom.uuid)
```
bdunne: Using `update_all`, every record that is found will get the same name.

d-m-u:
@jrafanie can I please get eyes on this?

PR #350, opened 2019-03-18
[WIP] Add price and currency cols to services

d-m-u:
As part of an RFE that I ended up with, we need backend support to be able to save a price of catalog when adding a Catalog Item, aka a service. Since we have no way of storing that information on services at the moment, we need a price and a currency column.

The schema portion of https://bugzilla.redhat.com/show_bug.cgi?id=1602072

### DIFF
--- a/db/migrate/20190315154516_add_price_to_services.rb
+++ b/db/migrate/20190315154516_add_price_to_services.rb
@@ -1,0 +1,6 @@
+class AddPriceToServices < ActiveRecord::Migration[5.0]
+  def change
+    add_column :services, :price, :bigint
+    add_column :services, :currency, :string
+  end
+end


d-m-u:
@carbonin sorry but could you please 👀 for me?

d-m-u:

lpichler:
fyi @d-m-u we have model `ChargebackRateDetailCurrency`and related table is seeded with 164 currencies with other information  like symbol, code,..:
```
 pp ChargebackRateDetailCurrency.first
  ChargebackRateDetailCurrency Load (0.3ms)  SELECT  "chargeback_rate_detail_currencies".* FROM "chargeback_rate_detail_currencies" ORDER BY "chargeback_rate_detail_currencies"."id" ASC LIMIT $1  [["LIMIT", 1]]
  ChargebackRateDetailCurrency Inst Including Associations (0.1ms - 1rows)
#<ChargebackRateDetailCurrency:0x00007fb5633035c0
 id: 1,
 code: "USD",
 name: "United States Dollar",
 full_name: "United States Dollar",
 symbol: "$",
 unicode_hex: "36",
 created_at: Thu, 07 Feb 2019 12:54:12 UTC +00:00,
 updated_at: Mon, 18 Mar 2019 10:18:58 UTC +00:00>
```

so I am asking if you want to use it and tie  `Services#currency` with `ChargebackRateDetailCurrency`
through some relation.  But I don't know use cases.

tinaafitz:
HI @lpichler, We haven't had to deal with prices/currency up until now. The use case is that we need to add a price and currency field during Service item creation to be displayed to the user at Catalog ordering time.  Based on that, how should we proceed?

carbonin:
I think @lpichler 's approach is probably best. That way we won't have to duplicate currency options. @tinaafitz is this just for display? Or are you going to integrate with chargeback at a more significant level?

tinaafitz:
Thanks @carbonin. My understanding is that this is just for display purposes.

carbonin:
Okay, then it's probably best to just use a reference to the existing, seeded, table rather than copying that effort. It will likely make things easier on the UI side as well.

PR #357, opened 2019-03-27
Dialog field - set load_values_on_init to true where show_refresh_button was not enabled

himdel:
Replaces code that was previously hardcoded in the model..

```ruby
    def load_values_on_init?
      return true unless show_refresh_button
      load_values_on_init
    end
```

Needed to make it posssible to set both these fields to false,
without breaking all the old fields where the value of `load_values_on_init` previously didn't matter.

https://bugzilla.redhat.com/show_bug.cgi?id=1684575
https://bugzilla.redhat.com/show_bug.cgi?id=1684567

(Depended on by:
* (core) https://github.com/ManageIQ/manageiq/pull/18600
* (ui-components) https://github.com/ManageIQ/ui-components/pull/376
)

### DIFF
--- a/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
+++ b/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
@@ -1,0 +1,21 @@
+require_migration
+
+describe DialogFieldLoadValuesOnInit do
+  let(:dialog_field_stub) { migration_stub(:DialogField) }
+
+  migration_context :up do
+    before do
+      [true, false, nil].each do |show|
+        [true, false, nil].each do |load|
+          dialog_field_stub.create!(:show_refresh_button => show, :load_values_on_init => load)
+        end
+      end
+
+      migrate
+    end
+
+    it "updates load_values_on_init" do
+      expect(dialog_field_stub.pluck(:load_values_on_init)).to eq([true, false, nil, true, true, true, true, true, true])
+    end
+  end
+end
--- a/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
+++ b/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
@@ -5,9 +5,9 @@ describe DialogFieldLoadValuesOnInit do
 
   migration_context :up do
     before do
-      [true, false, nil].each do |show|
-        [true, false, nil].each do |load|
-          dialog_field_stub.create!(:show_refresh_button => show, :load_values_on_init => load)
+      [true, false, nil].each do |show_button|
+        [true, false, nil].each do |load_values|
+          dialog_field_stub.create!(:show_refresh_button => show_button, :load_values_on_init => load_values)
         end
       end
 
@@ -15,7 +15,11 @@ describe DialogFieldLoadValuesOnInit do
     end
 
     it "updates load_values_on_init" do
-      expect(dialog_field_stub.pluck(:load_values_on_init)).to eq([true, false, nil, true, true, true, true, true, true])
+      expect(dialog_field_stub.sort_by(:id).pluck(:load_values_on_init)).to eq([
+        true, false, nil, # show_button was true
+        true, true, true,
+        true, true, true,
+      ])
     end
   end
 end
--- a/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
+++ b/spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
@@ -5,21 +5,35 @@ describe DialogFieldLoadValuesOnInit do
 
   migration_context :up do
     before do
-      [true, false, nil].each do |show_button|
-        [true, false, nil].each do |load_values|
-          dialog_field_stub.create!(:show_refresh_button => show_button, :load_values_on_init => load_values)
-        end
+      [true, false, nil].each do |load_values|
+        dialog_field_stub.create!(:show_refresh_button => show_button, :load_values_on_init => load_values)
       end
 
       migrate
     end
 
-    it "updates load_values_on_init" do
-      expect(dialog_field_stub.sort_by(:id).pluck(:load_values_on_init)).to eq([
-        true, false, nil, # show_button was true
-        true, true, true,
-        true, true, true,
-      ])
+    context "when show_refresh_button is true" do
+      let(:show_button) { true }
+
+      it "doesnt update load_values_on_init" do
+        expect(dialog_field_stub.order(:id).pluck(:load_values_on_init)).to eq([true, false, nil])
+      end
+    end
+
+    context "when show_refresh_button is false" do
+      let(:show_button) { false }
+
+      it "updates load_values_on_init to true" do
+        expect(dialog_field_stub.order(:id).pluck(:load_values_on_init)).to eq([true, true, true])
+      end
+    end
+
+    context "when show_refresh_button is nil" do
+      let(:show_button) { nil }
+
+      it "updates load_values_on_init to true" do
+        expect(dialog_field_stub.order(:id).pluck(:load_values_on_init)).to eq([true, true, true])
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


himdel:
Cc @gmcculloug @eclarizio
(this is the last one of the three :))

himdel:
Agreed @eclarizio , will add that to the ui-components PR (assuming that's what sets the initial value), thanks :)

eclarizio:
Actually, should it be added here? For people that use the API to create fields? (Is that even a thing people do?)

@gmcculloug Maybe you have some insight on this.

himdel:
Well, not in schema, but possibly `default_value_for` in the model?

eclarizio:
Yeah, that's probably fine

JPrause:
@himdel if this will be able to be backported, can you add the hammer/yes label.

himdel:
@eclarizio I'm still working on the other dialog_field_text_box specs, but.. for sorted item,
when show_refresh_button is false, load_values_on_init also false,
and the default value is not in the set of values provided,
it uses the default value anyway.

Which, on the one hand makes sense, because there is no list of values to compare against,
but OTOH is a random difference from the load_values_on_init=true case, where the random defaullt value does not get used.

Do you think that's acceptable behaviour?

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_migration
+
+describe DialogFieldLoadValuesOnInit do
+  let(:dialog_field_stub) { migration_stub(:DialogField) }
+
+  migration_context :up do
+    before do
+      [true, false, nil].each do |show|
+        [true, false, nil].each do |load|
```
carbonin: Using keywords as variable names makes me nervous ... maybe `load_value` here?

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+          dialog_field_stub.create!(:show_refresh_button => show, :load_values_on_init => load)
+        end
+      end
+
+      migrate
+    end
+
+    it "updates load_values_on_init" do
+      expect(dialog_field_stub.pluck(:load_values_on_init)).to eq([true, false, nil, true, true, true, true, true, true])
```
carbonin: Can you add an explicit sort here so we don't get bit by locale or OS default sorting rules?

carbonin:
@JPrause We typically don't backport schema PRs. Is there a compelling reason for this one to be backported?

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
himdel: I was hoping we can reliably get them in order they were created.

Is there any nice way of doing that?
(But I can live with sort too...)

himdel:
>> @himdel if this will be able to be backported, can you add the hammer/yes label.
> @JPrause We typically don't backport schema PRs. Is there a compelling reason for this one to be backported?

Explicitly leaving the backport/no backport decision on @gmcculloug & @loicavenel.

It looks like there is a need for fields that will only refresh because other fields change in hammer, and this is the only (sane) way of achieving that.

himdel:
> What is the actual relationship between these columns? Should they really be just one column with some logic around it?

@carbonin it's described in this PR's description ;)

Before, `load_on_init` was completely ignored (and assumed to be true) when `show_refresh_button` was not set to true.

Now, `load_on_init` and `show_refresh_button` will be completely independent.

But we still need to make sure that all those "assumed to be true" `load_on_init` values get explicitly set to true for this to not break old fields.

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
himdel: Fixed :+1:

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
himdel: Added a `sort_by(:id)` since we probably *can* rely on the ids being ascending, right?

And split into 3 lines to better show the relation.

carbonin:
Okay, that makes sense @himdel 

As far as the backport decision, it actually seems to me that this *cannot* be backported because we wouldn't be able to rely on this logic for people who have already upgraded to hammer (and by extension will not run this migration).

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      migrate
+    end
+
+    it "updates load_values_on_init" do
+      expect(dialog_field_stub.sort_by(:id).pluck(:load_values_on_init)).to eq([
+        true, false, nil, # show_button was true
+        true, true, true,
+        true, true, true,
```
carbonin: I think I would just prefer individual specs for each case. For example:
```ruby
it "doesnt update load_values_on_init when show_refresh_button is true" do
...
end

it "updates load_values_on_init when show_refresh_button is not true" do
...
end
```

Then we don't have to play this ordering game at all.

himdel:
> As far as the backport decision, it actually seems to me that this cannot be backported because we wouldn't be able to rely on this logic for people who have already upgraded to hammer (and by extension will not run this migration).

Agreed that the whole feature can absolutely not work without the migration backported too, if that's what you mean, yes.

(But if we did, it would, right?)

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
himdel: Yeah, maybe better with less magic, fixing :)

carbonin:
> Agreed that the whole feature can absolutely not work without the migration backported too, if that's what you mean, yes.

> (But if we did, it would, right?)

No, because people who have already migrated to hammer will not run this migration. Users are only expected to run migrations on major releases. This is why we don't backport migrations.

himdel:
Ah, well, in that case..

@miq-bot add_label hammer/no

That pretty much settles it I think, unless we want to wrap it in a product feature switch or something like that.

Thanks! :)

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
```diff
@@ -0,0 +1,41 @@
+require_migration
+
+describe DialogFieldLoadValuesOnInit do
+  let(:dialog_field_stub) { migration_stub(:DialogField) }
+
+  migration_context :up do
+    let (:show_button) { nil }
```
carbonin: I don't think this is needed.

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    before do
+      [true, false, nil].each do |load_values|
+        dialog_field_stub.create!(:show_refresh_button => show_button, :load_values_on_init => load_values)
+      end
+
+      migrate
+    end
+
+    context "when show_refresh_button is true" do
+      let (:show_button) { true }
```
carbonin: Pretty sure there shouldn't be a space between the `let` and `(` here. And that's why you're getting the rubocop issues.

carbonin: Same with the other `let(:show_button)` calls.

himdel:
Thanks @carbonin , those should be resolved now..

spec/migrations/20190327132620_dialog_field_load_values_on_init_spec.rb
himdel: It was, `undefined local variable or method `show_button' for #<RSpec::ExampleGroups::DialogFieldLoadValuesOnInit::Up:0x0000000009226718>` otherwise.

PR #364, opened 2019-04-11
Allow the application to specify the rails version.

jrafanie:
But rails 6 is out of the question!


jrafanie:
Boo, id_regions pulls in the rails version to test against.  I guess I need to fix it all at once.  😭

PR #367, opened 2019-05-13
Add price to service templates.

lfu:
Add columns for price and currency to service templates. So a service author can specify the price when creating a service template.

Part of https://github.com/ManageIQ/manageiq/pull/18754

https://bugzilla.redhat.com/show_bug.cgi?id=1602072

@miq-bot add_label enhancement, hammer/no, changelog/yes

### DIFF
--- a/db/migrate/20190513155251_add_price_to_service_templates.rb
+++ b/db/migrate/20190513155251_add_price_to_service_templates.rb
@@ -1,0 +1,6 @@
+class AddPriceToServiceTemplates < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :service_templates, :currency, :type => :bigint
+    add_column :service_templates, :price, :float
+  end
+end


carbonin:
Also, #350 was adding this to services while this PR is adding it to service_templates. Which is the correct place?

carbonin:
Oh I see it's an `add_reference` now.

carbonin:
I guess the question of which is the correct table is still out there though. @lfu why `service_templates` when @d-m-u was going to add it to `services`?

lpichler:
> I guess the question of which is the correct table is still out there though. @lfu why `service_templates` when @d-m-u was going to add it to `services`?

Not sure why but usually when anybody mention in BZ "Service" he most likely means Catalog Items or Catalog Bundle. (Service Templates)

there is mentioned path  Service -> Catalog in UI, - so I think that Catalog Items or Catalog Bundle. (Service Templates) are correct here. Can you confirm it @lfu ?

db/migrate/20190513155251_add_price_to_service_templates.rb
```diff
@@ -0,0 +1,6 @@
+class AddPriceToServiceTemplates < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :service_templates, :currency, :type => :bigint
```
lpichler: I think we should have `chargeback_rate_detail_currency` here - to see that it is related to `chargeback_rate_detail_currencies` table.

`add_reference :service_templates, :chargeback_rate_detail_currency, :type => :bigint`

you can specify foreign key in the model.

@carbonin what do you think ?

lfu:
A service author would create a service template (that can be a catalog item or catalog bundle) that users can order to get services. The relationship of service template and service is kind of like template and VM.

db/migrate/20190513155251_add_price_to_service_templates.rb
carbonin: I think the idea is to make the method `#currency` and the column `currency_id` so the relationship in the model specifies the class https://github.com/ManageIQ/manageiq/pull/18754/files#diff-9257fff45a7d1258305c9a6b7d15e322R66

carbonin: It's unfortunate that `chargeback_rate_detail_currency` isn't named just `currencies` because it doesn't look like the table has anything chargeback specific in it.

lfu: @carbonin Exactly what I am going to say.

lpichler: 👍 👍

Fryguy: What is the effort to rename the table :)

PR #369, opened 2019-05-13
Change plugin common methods to class methods

Fryguy:
Change plugin common methods to class methods


carbonin:
@Fryguy does this need to be backported?

Fryguy:
@carbonin Nope.

PR #370, opened 2019-05-14
[WIP] Removing deleted Custom Buttons

romanblanco:
After https://github.com/ManageIQ/manageiq/pull/18368 the purpose of `set_data[:button_order]`
has been changed, and an error is raised after going through an array,
where the button does not exist anymore

TODO:

before merging, this needs to 
  - update two forms
     1. (Automation -> Automate -> Generic Object) -> Add a new Button
     2. (Automation -> Automate -> Customization -> Buttons) -> Add a new Button
  - add a hook to remove id from CustomButtonSet every time CustomButton is removed

cc/ @lpichler @himdel

### DIFF
--- a/db/migrate/20190509142148_update_custom_button_sets.rb
+++ b/db/migrate/20190509142148_update_custom_button_sets.rb
@@ -1,0 +1,21 @@
+class UpdateCustomButtonSets < ActiveRecord::Migration[5.0]
+  class MiqSet < ActiveRecord::Base
+    serialize :set_data
+  end
+
+  class CustomButton < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Removing deleted buttons from Custom Button Sets") do
+      MiqSet.select(:id, :set_data).where(:set_type => 'CustomButtonSet').each do |cbs|
+        next if cbs.set_data[:button_order].blank?
+        existing_buttons = CustomButton
+          .where(:id => cbs.set_data[:button_order])
+          .order("position(id::text in '#{cbs.set_data[:button_order].join(',')}')")
+          .ids
+        cbs.set_data[:button_order] = existing_buttons
+        cbs.save!
+      end
+    end
+  end
+end
--- a/spec/migrations/20190509142148_update_custom_button_sets_spec.rb
+++ b/spec/migrations/20190509142148_update_custom_button_sets_spec.rb
@@ -1,0 +1,25 @@
+require_migration
+
+describe UpdateCustomButtonSets do
+  let(:custom_button_set_stub) { migration_stub :MiqSet }
+  let(:custom_button_stub) { migration_stub :CustomButton }
+
+  migration_context :up do
+    it 'Removes non-existing buttons while keeping the buttons sorted' do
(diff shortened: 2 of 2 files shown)


db/migrate/20190509142148_update_custom_button_sets.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  class CustomButton < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Removing deleted buttons from Custom Button Sets") do
+      MiqSet.select(:id, :set_data).where(:set_type => 'CustomButtonSet').each do |cbs|
+        next if cbs.set_data[:button_order].blank?
+        existing_buttons = CustomButton
+          .where(:id => cbs.set_data[:button_order])
+          .order("position(id::text in '#{cbs.set_data[:button_order].join(',')}')")
```
himdel: Ugh, that looks like an SQL injection.
We can *believe* `button_order` only contains ints, but unless we check it, this is dangerous.

db/migrate/20190509142148_update_custom_button_sets.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateCustomButtonSets < ActiveRecord::Migration[5.2]
```
carbonin: This needs to be `[5.0]` as manageiq doesn't support 5.2 yet.

We're going to move this repo back to 5.0 to avoid this until then.

romanblanco: @carbonin corrected. Thank you :+1:

db/migrate/20190509142148_update_custom_button_sets.rb
romanblanco: @lpichler as you've corrected me not to use Ruby sort, is there a good reason not to use the Ruby sort?

lpichler: I am not against ruby solution I knew that we are using implementation in SQL

https://github.com/ManageIQ/manageiq/pull/18060/files#diff-72d768ae2e73b9e77080d4a7df59cf56R4

so you can copy `with_array_order` here or use ruby.

db/migrate/20190509142148_update_custom_button_sets.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateCustomButtonSets < ActiveRecord::Migration[5.0]
+  class MiqSet < ActiveRecord::Base
```
lpichler: let's use `CustomButtonSet` to be consistent with `core`

himdel:
It may be too late for a migration as this change has made it to hammer.

To fix https://bugzilla.redhat.com/show_bug.cgi?id=1770300, I think we'll need to change the validation to skip missing buttons instead of throwing.

So, I think we can close this one..

h-kataria:
@himdel https://github.com/ManageIQ/manageiq/pull/19573 is merged, should we close this?

himdel:
Indeed, there will be one more core PR to make sure service UI and the automate buttons tree is right too, but we can close this one.

@miq-bot close_issue

PR #371, opened 2019-05-16
Introduce :firmware_registry and :firmware_binary tables

miha-plesko:
We introduce four new tables that correspond to three new Rails models needed for managing infrastructure firmware:

```
FirmwareRegistry
  .name             # unique name
  .last_refresh_on  # when relationships were last invoked
  .type             # needed for STI

FirmwareBinary
  .name        # freeform name
  .description # freeform description
  .version     # firmware version
  .type        # needed for STI

FirmwareTarget
      .manufacturer # physical component's manufacturer
      .model        # physical component's model name
```
The Binary and the Target are connected into many-to-many relation.

Needed for https://github.com/ManageIQ/manageiq/pull/18774

@miq-bot add_label schema,enhancement
@miq-bot assign @agrare

/cc @tadeboro @matejar

### DIFF
--- a/db/migrate/20190514115219_create_firmware_registries.rb
+++ b/db/migrate/20190514115219_create_firmware_registries.rb
@@ -1,0 +1,11 @@
+class CreateFirmwareRegistries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_registries do |t|
+      t.string   :name
+      t.datetime :last_refresh_on
+      t.text     :last_refresh_error
+      t.string   :type
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20190514124322_create_firmware_binaries.rb
+++ b/db/migrate/20190514124322_create_firmware_binaries.rb
@@ -1,0 +1,13 @@
+class CreateFirmwareBinaries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries do |t|
+      t.string     :name
+      t.string     :external_ref
+      t.text       :description
+      t.string     :version
+      t.string     :type
+      t.references :firmware_registry, :type => :bigint, :index => true
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20190517093412_create_firmware_targets.rb
+++ b/db/migrate/20190517093412_create_firmware_targets.rb
@@ -1,0 +1,9 @@
+class CreateFirmwareTargets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_targets do |t|
+      t.string :manufacturer
+      t.string :model
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20190517093516_create_firmware_binary_firmware_targets.rb
+++ b/db/migrate/20190517093516_create_firmware_binary_firmware_targets.rb
@@ -1,0 +1,10 @@
+class CreateFirmwareBinaryFirmwareTargets < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries_firmware_targets do |t|
+      t.references :firmware_binary, :type => :bigint, :index => false
+      t.references :firmware_target, :type => :bigint, :index => false
+
+      t.index %i[firmware_binary_id firmware_target_id], :unique => true, :name => :index_firmware_binaries_firmware_targets
+    end
+  end
+end


agrare:
Hey @miha-plesko couple of questions, are these registries owned by the physical infra manager?  How are they refreshed?  Are they only refreshed on demand or is it part of the ems_refresh for the provider?

tadeboro:
These two models are not tied to any specific provider. And while it is true that currently, only Redfish provider will use them when performing firmware upgrades, these two can be used by any other part of the ManageIQ.

The way we envisioned this to work is quite similar to how bare metal provisioning works. First, someone has to add a `PxeServer` to ManageIQ (under `Compute` -> `Infrastructure` -> `PXE`) and refresh it. This refresh fetches the list of all available OS images from the PXE server and stores it into a series of `PxeImage`s. Note that ManageIQ only stores metadata about the OS image in the database, not the actual image content. Server admin can now use these images when she is provisioning the servers.

The firmware update process should follow roughly the same path. A system admin will create new `FirmwareRegistry` in `here` -> `lives` -> `Firmware` and refresh it. Refresh process will download a list of all available firmware binaries and store the data about each binary into `FirmwareBinary` instance. Again, ManageIQ only stores metadata about the binaries. And once ManageIQ knows about the binaries, end users can use them to update their hardware.

miha-plesko:
Hi @agrare yes so it's like @tadeboro described, it's not part of the Redfish provider. We're still working on UI but basically we're adding a new left menu entry Compute > Infrastructure > Firmware:

image

image

User will be offered to add a new registry there and manually click "Refresh Relationships" on it to fetch the binaries. The approach is similar to either PXE server thing (as @tadeboro mentioned) or to embedded ansible repository thing.

### Note
Beware @agrare there is currently no standardized interface for the remote part of the thing (where're the MIQ is connecting to to do "Refresh Relations"). We're implementing our own super simple REST service that will hold metadata for binaries. The actual binaries will reside on separate ftp/samba/http/whatever and the REST service will only return a ftp/samba/... link for each binary plus the metadata.

This is why we allow for subclassing the FirmwareRegistry (in the core PR), so that it's super easy to add support for arbitrary implementation of the remote depot in the future. But as I said, for now we only implement FirmwareRegistrySimple which connects to this our REST service.

Looking forward to be hearing your opinion. Do you perhaps know of a standardized way of storing tagged firmware binaries that we could adopt?

miha-plesko:
The core PR is now covered with tests that are green when using this this schema enhancement. So I think we're good to go.

db/migrate/20190514115219_create_firmware_registries.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFirmwareRegistries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_registries do |t|
+      t.string   :name
+      t.datetime :last_refresh_on
+      t.string   :type
+      t.bigint   :endpoint_id
```
agrare: Endpoint has a polymorphic `belongs_to :resource` so you don't need the `endpoint_id` here, you just need `has_one :endpoint` in your model

db/migrate/20190514115219_create_firmware_registries.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.bigint   :authentication_id
```
agrare: Same thing here

db/migrate/20190514115219_create_firmware_registries.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFirmwareRegistries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_registries do |t|
+      t.string   :name
+      t.datetime :last_refresh_on
```
agrare: You might want to also add a `:last_refresh_error` to track failures

db/migrate/20190514124322_create_firmware_binaries.rb
```diff
@@ -0,0 +1,13 @@
+class CreateFirmwareBinaries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries do |t|
+      t.string   :name
+      t.string   :ems_ref
+      t.text     :description
+      t.string   :version
+      t.string   :type
+      t.bigint   :firmware_registry_id
```
agrare: You can use `t.references :firmware_registry` instead of calling out the exact id column.  You might want to also add `:index => true`

db/migrate/20190517093516_create_firmware_binary_firmware_constraints.rb
```diff
@@ -0,0 +1,13 @@
+class CreateFirmwareBinaryFirmwareConstraints < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries_firmware_constraints do |t|
+      t.bigint   :firmware_binary_id
```
agrare: Same here, you can use `t.references`

db/migrate/20190517093516_create_firmware_binary_firmware_constraints.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.bigint   :firmware_constraint_id
```
agrare: Same, you probably don't want to `:index => true` on this but add a standalone `t.index %i[firmware_binary_id firmware_constraint_id], :unique => true` so there aren't extra indexes.

db/migrate/20190517093516_create_firmware_binary_firmware_constraints.rb
```diff
@@ -0,0 +1,13 @@
...
+    end
+
+    add_index :firmware_binaries_firmware_constraints,
```
agrare: You should be able to do a `t.index` in the `create_table` block

agrare:
@carbonin @bdunne can you take a look also?

db/migrate/20190514124322_create_firmware_binaries.rb
```diff
@@ -0,0 +1,13 @@
+class CreateFirmwareBinaries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries do |t|
+      t.string   :name
+      t.string   :ems_ref
```
carbonin: This is confusing, I thought you said these were not associated with an EMS? I understand that this may need some kind of native identifier, but if the system isn't an EMS then this should probably have a different name.

db/migrate/20190517093412_create_firmware_constraints.rb
```diff
@@ -0,0 +1,9 @@
+class CreateFirmwareConstraints < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_constraints do |t|
```
carbonin: I'm not sure I understand the name of this table. How does this represent a constraint? It just looks like some shared details.

db/migrate/20190517093516_create_firmware_binary_firmware_constraints.rb
miha-plesko: Thanks. Turns out `:index => true` is default so I had to manually disable it.

db/migrate/20190517093412_create_firmware_constraints.rb
agrare: @miha-plesko would it make sense to embed this into the firmware record then?  Why split it out into its own model

db/migrate/20190514124322_create_firmware_binaries.rb
```diff
@@ -0,0 +1,13 @@
+class CreateFirmwareBinaries < ActiveRecord::Migration[5.0]
+  def change
+    create_table :firmware_binaries do |t|
+      t.string     :name
+      t.string     :external_id
```
agrare: Looks like the specs are failing because it is expecting any `_id` to be a bigint.  How about `:external_ref` here instead?

db/migrate/20190517093412_create_firmware_constraints.rb
carbonin: Agree, this feels like attributes of firmware that should be indexed.

miha-plesko: Correct me if I'm wrong, but you're suggesting to add those two fields on FirmwareBinary?

```
FirmwareBinary.compatible_manufacturer
FirmwareBinary.compatible_model
```

But usually the same firmware binary can be applied on more than one kind of physical server. How would we handle this then? Should we just duplicate the firmware binary itself?

carbonin: Ah I see. Maybe just change the name to something more descriptive ... like `firmware_targets`? "Constraints" makes me feel like we're implementing a version constraint system of some kind.

What do you think @agrare ?

carbonin: This also feels like an attribute of the physical server, but I guess that relationship isn't being modeled and instead you'll match up the manufacturer and model when you need to do the update, right?

miha-plesko: Nice, if @agrare also agrees I'll rename like you suggest. Was in search for better name myself but couldn't think of one

agrare: Ah okay I thought you were modelling this the other way around, that one target platform could use multiple firmwares not one firmware could be used by multiple target platforms.

`:firmware_targets` makes sense then :+1:

carbonin: FWIW @agrare I think it's a many-to-many relationship (judging by the join table :wink: )

miha-plesko: @agrare it's both ways - that's why we're using many-to-many.

- a firmware can be applied to multiple platforms (like `Dell ZX01`, `Dell ZX8` etc)
- a platform can have many firmwares applied (like `nic-firmware-upodate-5.2.3.exe`, `bios-update-3.1.exe` etc)

@carbonin yeah both `manufacturer` and `model` are attribute on physical server. In theory we could extend inventory parser to link appropriate `firmware_target` as well, but that would bring redundancy. Unless we remove those two attributes from server, but how would you then update a server... It gets complicated

carbonin: Yeah, I was thinking about that too. Since neither is actually dependent on each other you probably need to collect them separately. (e.g. you can have firmware that targets a server that isn't in inventory yet).

You could probably have a separate many-to-many relationship between physical servers and firmware targets, but I don't think that really gains you anything over joining using the manufacturer and model ...

carbonin:
@miha-plesko Moving this out of the diff conversation ...

Are the manufacturer and model collected from the firmware registry and the physical server always expected to match exactly? Is it safe to assume that we can match on strings with confidence?

miha-plesko:
@carbonin good question, it bothers me too: unfortunately that's just a freeform string as far as I know; no pre-defined values. I'm planning to at least downcase the string when matching to be case-invariant, but in general it's administrator's job to make sure it'll match.

In other words: firmware depot owner will have to ensure his firmware is properly tagged. I think that is a valid assumption. I also don't see a better way of doing it?

PR #377, opened 2019-05-31
Add guid to miq_databases

Fryguy:
It turns out the only real "product-wide" guid we have is in
miq_regions, but the problem with miq_regions is that it's replicated,
so it's slightly complicated to find my_region.guid.  Instead, let's
have a miq_databases.guid.

@agrare @jrafanie Please review.

Note that some of the motivation is for identification via the Red Hat Cloud Services stuff.

### DIFF
--- a/db/migrate/20190531161732_add_guid_to_miq_databases.rb
+++ b/db/migrate/20190531161732_add_guid_to_miq_databases.rb
@@ -1,0 +1,16 @@
+class AddGuidToMiqDatabases < ActiveRecord::Migration[5.2]
+  class MiqDatabase < ActiveRecord::Base
+  end
+
+  def up
+    add_column :miq_databases, :guid, :string, :limit => 36
+
+    say_with_time("Create default guids for miq_databases") do
+      MiqDatabase.all.each { |d| d.update!(:guid => SecureRandom.uuid) }
+    end
+  end
+
+  def down
+    remove_column :miq_databases, :guid
+  end
+end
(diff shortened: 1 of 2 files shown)


spec/migrations/20190531161732_add_guid_to_miq_databases_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_migration
+
+describe AddGuidToMiqDatabases do
+  let(:miq_database) { migration_stub(:MiqDatabase) }
+
+  migration_context :up do
+    it "sets miq_database guids" do
+      obj1 = miq_database.create!
+
+      expect(obj1.reload.guid).to eq(nil)
```
agrare: @Fryguy before the migration `#guid` doesn't exist

Fryguy: Yeah I removed that...seems I didn't push it 😩

db/migrate/20190531161732_add_guid_to_miq_databases.rb
```diff
@@ -0,0 +1,16 @@
+class AddGuidToMiqDatabases < ActiveRecord::Migration[5.2]
```
agrare: Not sure why this didn't fail specs but when ran this locally:
```
ArgumentError: Unknown migration version "5.2"; expected one of "4.2", "5.0"
/var/lib/gems/2.5.0/gems/activerecord-5.0.7.2/lib/active_record/migration/compatibility.rb:9:in `find'
/var/lib/gems/2.5.0/gems/activerecord-5.0.7.2/lib/active_record/migration.rb:528:in `[]'
```

agrare: https://github.com/ManageIQ/manageiq-schema/pull/379

PR #378, opened 2019-05-31
Add missing say_with_time

Fryguy:
This data migration missed the `say_with_time`, so this adds it in.

@jrafanie Please review

cc @d-m-u

### DIFF
--- a/db/migrate/20190304192641_add_provider_guids.rb
+++ b/db/migrate/20190304192641_add_provider_guids.rb
@@ -6,6 +6,10 @@ class AddProviderGuids < ActiveRecord::Migration[5.0]
   end
 
   def up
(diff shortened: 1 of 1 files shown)


db/migrate/20190304192641_add_provider_guids.rb
```diff
@@ -6,6 +6,10 @@ class Provider < ActiveRecord::Base
   end
 
   def up
-    Provider.in_my_region.where(:guid => nil).each { |provider| provider.update_attributes!(:guid => SecureRandom.uuid) }
+    say_with_time("Update provider guid values") do
+      Provider.in_my_region.where(:guid => nil).each do |provider|
+        provider.update_attributes!(:guid => SecureRandom.uuid)
```
jrafanie: FYI, I think update_attributes is deprecated in rails 5.2.  That's going to be fun.

PR #365, opened 2019-04-11
Support rails 5.1  (Now Supports rails 5.2 too :tada:)

jrafanie:
Adds support for rails 5.1 and 5.2.

PR #379, opened 2019-05-31
Fix ActiveRecord Migration version

agrare:
ActiveRecord version 5.2 is unsupported by manageiq:
```
ArgumentError: Unknown migration version "5.2"; expected one of "4.2", "5.0"
/var/lib/gems/2.5.0/gems/activerecord-5.0.7.2/lib/active_record/migration/compatibility.rb:9:in `find'
/var/lib/gems/2.5.0/gems/activerecord-5.0.7.2/lib/active_record/migration.rb:528:in `[]'
...
```

Introduced by https://github.com/ManageIQ/manageiq-schema/pull/377

### DIFF
--- a/db/migrate/20190531161732_add_guid_to_miq_databases.rb
+++ b/db/migrate/20190531161732_add_guid_to_miq_databases.rb
@@ -1,4 +1,4 @@
-class AddGuidToMiqDatabases < ActiveRecord::Migration[5.2]
+class AddGuidToMiqDatabases < ActiveRecord::Migration[5.0]
   class MiqDatabase < ActiveRecord::Base
   end
 


carbonin:
oops ... :grimacing:

agrare:
Yeah trying to figure out how it passed specs

carbonin:
Because schema is using rails 5.2 for specs it seems :scream: https://travis-ci.org/ManageIQ/manageiq-schema/jobs/539851236#L293

carbonin:
https://github.com/ManageIQ/manageiq-schema/pull/365

carbonin:
@jrafanie how close are we to rails 5.2 in core. Should we stick to testing with the version miq uses to avoid things like this?

agrare:
Yeah if not imminent we could fix breakages introduced by 5.2 but keep locked to the version MIQ core uses

agrare:
And yeah confirmed that this fails specs with rails ~> 5.0.7.2

jrafanie:
> @jrafanie how close are we to rails 5.2 in core. Should we stick to testing with the version miq uses to avoid things like this?

Good question.  We know virtual attributes has an incompatibility that Keenan's very close to fixing.  I believe it's then just a matter of playing with 5.1 and fixing anything that wasn't covered by tests (actioncontroller parameters no longer a hash, I'm looking at you), then move to 5.2.  I don't think there's anything in 5.2 that's nearly as bad as the ones in 5.1 so as soon as I get some 👍 from the UI team, I'd like to move to 5.1 on master and then 5.2 soon after if that doesn't blow up.

jrafanie:
> Yeah if not imminent we could fix breakages introduced by 5.2 but keep locked to the version MIQ core uses

I purposely unlocked the schema repo from manageiq to ensure it worked with 5.1 and 5.2, even though it only tests with the latest 5.2.  We should really test with all supported versions but we didn't do that for now.  Maybe that's really what we should do.

I didn't think people would have manageiq bundled with rails 5.1 or 5.2 until we had that on core so didn't think it would be a problem with generating a migration for new versions.

jrafanie:
Sorry, I didn't understand before.  I see now that if you bundle update with schema, it will use 5.2 and that's what the generated migration will create, because it's untied from manageiq.  I'm chatting with others to see if we can get to 5.1 very soon, and if so, I'll lock schema to 5.1 at the most and then we'll get to 5.2 when 5.1 looks good.

PR #380, opened 2019-06-11
Open URL through automate for various objects

martinpovolny:
Adding a table for temporary storing URL returned from automate.

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1550002

UI PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/5688
Guides article: https://github.com/martinpovolny/guides/blob/automate_url_opener/automate_url_open.md (from PR: https://github.com/ManageIQ/guides/pull/355)
Core PR: https://github.com/ManageIQ/manageiq/pull/18849
Automate PR: https://github.com/ManageIQ/manageiq-automation_engine/pull/328

### DIFF
--- a/db/migrate/20190604090631_create_external_urls.rb
+++ b/db/migrate/20190604090631_create_external_urls.rb
@@ -1,0 +1,11 @@
+class CreateExternalUrls < ActiveRecord::Migration[5.0]
+  def change
+    create_table :external_urls, :id => :bigserial do |t|
+      t.references :resource, :polymorphic => true, :type => :bigint
+      t.string  :url
+      t.belongs_to :user, :type => :bigint
+    end
+
+    add_index "external_urls", [:resource_id, :resource_type], :name => "external_urls_on_resource_id_and_resource_type"
+  end
+end


martinpovolny:
ping @Ladas

db/migrate/20190604090631_create_saved_urls.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSavedUrls < ActiveRecord::Migration[5.0]
+  def change
+    create_table :saved_urls do |t|
+      t.bigint  "refers_to_id", :type => :bigint
+      t.string  "refers_to_type"
```
Ladas: `t.references :refers_to, polymorphic: true, index: true`

should create the `refers_to_id` and ..._type` and the index

db/migrate/20190604090631_create_saved_urls.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSavedUrls < ActiveRecord::Migration[5.0]
+  def change
+    create_table :saved_urls do |t|
```
Ladas: should be `create_table :saved_urls, :id => :bigserial` I believe

Ladas:
cc @lpichler what's the rake command to generate the schema? We should still run it, right? (The CI was failing without it in the past)

martinpovolny:
@Ladas : PR updated according to your guidance. Please, check again.

db/migrate/20190604090631_create_external_urls.rb
```diff
@@ -0,0 +1,11 @@
+class CreateExternalUrls < ActiveRecord::Migration[5.0]
+  def change
+    create_table :external_urls, :id => :bigserial do |t|
+      t.references :refers_to, :polymorphic => true, :index => true, :type => :bigint
```
Ladas: if the `index => true` here does't work, lets remove it (it should create `external_urls` for you, but it's not implemented in all migration methods)

db/migrate/20190604090631_create_external_urls.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string  "url"
```
lpichler: nitpick: but we are using symbols:

`t.string  "url"`

db/migrate/20190604090631_create_external_urls.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.belongs_to :user, :type => :bigint
+    end
+
+    add_index "external_urls", ["refers_to_id", "refers_to_type"], :name => "saved_urls_on_refers_to_id_and_refers_to_type"
```
lpichler: same here

lpichler:
> cc @lpichler what's the rake command to generate the schema? We should still run it, right? (The CI was failing without it in the past)

looks like it was removed: http://talk.manageiq.org/t/new-schema-specs-for-new-replication/1404/4

db/migrate/20190604090631_create_external_urls.rb
lpichler: `refers_to` maybe just `resource` ?

martinpovolny:
@Ladas : fixed.

@lpichler : renamed and also renamed in the core PR here: https://github.com/ManageIQ/manageiq/pull/18849

lpichler:
@miq-bot assign @carbonin

db/migrate/20190604090631_create_external_urls.rb
carbonin: Agree with @lpichler here, can we use all symbols for the column names?

carbonin:
@martinpovolny I think the BZ link should be https://bugzilla.redhat.com/show_bug.cgi?id=1643331, right?

The one in the initial PR comment is closed.
Also could you add it to the commit message so the bot will pick it up and comment on the BZ?

carbonin:
Also, what objects are going to be referenced by this url and if it's temporary, why do we need a table?

agrare:
@martinpovolny If the goal is to show the vm summary page why a new table and not an attribute on the vm like we do with topology?

Also, the URL for vsphere is going to be different depending on version, if you have the html5 UI or just the old flash UI, etc... so I'm curious who would be generating this URL.

E.g. for vSphere 6.7 it looks like this `https://dev-vc67/ui/#?extensionId=vsphere.core.inventory.serverObjectViewsExtension&objectId=urn:vmomi:VirtualMachine:vm-337:a6b3087a-3d6b-4852-9732-17094d702269&navigator=vsphere.core.viTree.vmsAndTemplatesView`

martinpovolny:
@carbonin : the link is right.

> @martinpovolny If the goal is to show the vm summary page why a new table and not an attribute on the vm like we do with topology?

No that is not the goal. The goal is in the BZ. If the goal is not clear from the BZ, then, please, read the (partial) documentation in this PR: https://github.com/ManageIQ/guides/pull/355

This PR has nothing to do with the VM consoles although it can be used to replace a VM console with a custom one.

martinpovolny:
> Also, what objects are going to be referenced by this url and if it's temporary, why do we need a table?

How else do we pass data from a worker running on one appliance to a different appliance?

agrare:
The bz you have linked to in the description is closed as a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1643331

The goal of the BZ is pretty clear:
> Description of request:
> 
> Ability to create CloudForms custom buttons which can open an HTTPS URL to a provider's web client.

Also important that this also isn't the VM's console, but the VM's summary page.

db/migrate/20190604090631_create_external_urls.rb
martinpovolny: I missed that. Fixed now.

martinpovolny:
@NickLaMuro, @agrare: this is about extending an existing mechanism (mentioned for example here:
http://talk.manageiq.org/t/lauch-ssh-error/4343/15 ) to work with several new types of entities.

I have implemented that mechanism several years ago, and at that time I did put the URL in the `SystemConsole` table. That limits the usefulnes of that to just VMs.

Here I am trying to do the same in a more universal way.

I do not know how else to explain it to you. Above I have linked a PR to a guides PR where the mechanism is explained: https://github.com/martinpovolny/guides/blob/automate_url_opener/automate_url_open.md.

Do you need any further explanation? Shall we schedule a video chat?

NickLaMuro:
@martinpovolny Wrong "Nick" I think...

cc @carbonin

martinpovolny:
> The bz you have linked to in the description is closed as a duplicate of > https://bugzilla.redhat.com/show_bug.cgi?id=1643331

I have linked the one that contains a better explanation of the issue. On purpose.

You are linking the one that only mentiones one usecase and very shortly. Please, read the BZ that I linked.

agrare:
Okay sounds like this bz https://bugzilla.redhat.com/show_bug.cgi?id=1643331 should have actually been marked as a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1550002 then?  We shouldn't be pointing PRs to closed BZs so lets either reopen the original one or add enough description to the current BZ.

martinpovolny:
I am glad we moved forward ;-)

Are there any changes needed here on top of what @Ladas and @lpichler requested (and I already did)?

PR #382, opened 2019-06-13
Add ansible columns to authentications

carbonin:
This will allow us to store ansible credentials in a more sane way


carbonin:
@bdunne before doing this I was thinking that we should review the whole authentications table and try to pare it down a bit. At least since we're dropping a bunch of providers which probably account for a good amount of the specific columns in there.

But for this case I don't think anything else fits unless we have just a ton of encrypted columns then some serialized metadata column to tell us what is where, but that seems way worse than having more descriptive column naming even if we end up with a ton of columns.

I think there is minimal downside to having more columns and it's still the best way for new developers to understand what's going on (even if it is objectively confusing).

Fryguy:
>  At least since we're dropping a bunch of providers which probably account for a good amount of the specific columns in there.

Not in upstream.

NickLaMuro:
@miq-bot add_label embedded ansible

PR #383, opened 2019-06-18
Don't sync the schema_migrations_ran table when adding it

carbonin:
The purpose of the table is to track when migrations are run on
remote regions. If we sync the whole table outside of the scope
of the transaction based replication then we're violating the
assumption that the migration version number only appears in the
global region when the data from that migration was replicated
up to the global region.

This was causing a timing issue where the sync could take place
just before the global region runs its migrations. Then the global
would blaze past migrations without successfully replicating up
the data needed for a particular schema version.

This could lead to an irrecoverable situation in replication where
the subscription would have to be removed and re-added.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1721596

### DIFF
--- a/db/migrate/20171031010000_add_schema_migrations_ran.rb
+++ b/db/migrate/20171031010000_add_schema_migrations_ran.rb
@@ -10,7 +10,9 @@ class AddSchemaMigrationsRan < ActiveRecord::Migration[5.0]
 
     pglogical = ActiveRecord::Base.connection.pglogical
     if pglogical.enabled? && pglogical.replication_sets.include?('miq')
-      pglogical.replication_set_add_table('miq', "schema_migrations_ran", true)
+      # don't do an initial data sync for this table
+      # we want the data do come in as it's replicated rather than in bulk
+      pglogical.replication_set_add_table('miq', "schema_migrations_ran", false)
     end
   end
 


carbonin:
I was able to reproduce the conflicts we saw in the `schema_migrations_ran` table, but not the actual failure of replication. I suspect that I would need some more data in the tables in question to hit the timing necessary to make it happen.

I was able to verify our suspicion that the entire contents of the remote region's `schema_migrations_ran` table were replicated to the global after just migrating to the version that adds the table on the global. That is really all we need to know to prove, in theory, that we could get into the situation described in the BZ.

After applying this patch there were no more conflicts in the PG logs and the migration also completed without issue.

carbonin:
This really does need to be backported, but I have no idea what the implications of that would be ...

bdunne:
@carbonin Do we need to test again with this change that the only records in the table after running this migration on the global are the old records?

carbonin:
@bdunne I think that's not exactly how it would work.

It sounds like you think that `rake db:migrate VERSION=20171031010000` on the global after a full migration on the remote would lead to the global `schema_migrations_ran` table only containing remote region records with `version` < the latest version migrated to on the global.

What actually happens is that all of the changes which *can* be applied to the current schema from the remote to the global are replicated. Put another way, replication runs until it blows up.

What we should expect is that the global `schema_migrations_ran` table will contain only records for versions which are before the first migration which generates data which references schema that is not yet present in the global database.

So the short answer is :point_up: is what I reproduced, but with the `visible` column which was added to `zones` in 20180618083035. Left to it's own this would resolve itself and not actually cause the irrecoverable situation we see with product_name and switches because the column is not subsequently removed.

db/migrate/20171031010000_add_schema_migrations_ran.rb
```diff
@@ -10,7 +10,7 @@ def up
 
     pglogical = ActiveRecord::Base.connection.pglogical
     if pglogical.enabled? && pglogical.replication_sets.include?('miq')
-      pglogical.replication_set_add_table('miq', "schema_migrations_ran", true)
+      pglogical.replication_set_add_table('miq', "schema_migrations_ran")
```
Fryguy: As discussed, I'd prefer if we explicitly passed `false` here, and maybe add a comment as well to declare our intention.

simaishi:
Hammer backport details:

```
$ git log -1
commit ef8bc8889f913f577342acffffcc0b4916de7f06
Date:   Wed Jun 19 10:12:35 2019 -0400

    Merge pull request #383 from carbonin/dont_sync_the_schema_migrations_ran
    
    Don't sync the schema_migrations_ran table when adding it
    
    (cherry picked from commit 88dfe84442cf4ff6c5fb82f9e706478ffd54aa59)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1722540
```

PR #385, opened 2019-06-21
Migrate awx credentials to authentications

carbonin:
### The process for this migration goes something like this:
1. Pull credential information out of the `awx` database using PG
    - NOTE: we pull the database connection info from the current AR raw connection and replace the dbname value.
2. Decrypt the encrypted fields ~~using the `standalone_decrypt.py` script.~~
3. Save the resulting attributes to the existing authentication record

### Here are most of the edge cases I thought of and how the migration handles each:
  - Bail with a migration "success" if we can't connect to the `awx` database
  - Ignore anything we pull from the awx database that we don't know where to put (not in options and also not in the field map)
  - Only allow fields in the options column that we have listed as options fields. (i.e. if there used to be some other field pulled in by inventory sync, we will throw that away)
  - Set options to `nil` if it ends up empty
  - Fail hard if we fail to decrypt something from the `awx` database

## ~~About that python script ...~~

~~To build that I pretty much just copied a bunch of `awx` source code. Here's a link: https://github.com/ansible/awx/blob/e269634afc9386807ceaf56a61f7f0fad6671bf0/awx/main/utils/encryption.py~~

~~That code depends on some django utilities so we will need to ensure that the `django` python module is installed to run this properly. (https://github.com/ManageIQ/manageiq-appliance-build/pull/324)~~

### If this is failing for you ...

I'm sorry :sweat_smile:
This is the only situation I can imagine you're in:
  - You have some embedded ansible credentials in your miq database
  - You have the awx database locally and accessible with the credentials built as described above
  - ~~If you have `python3` and `django` locally:~~
    - The awx secret key in your vmdb database probably has been regenerated and doesn't go with the awx database that we're connecting to.
    - You should `DROP DATABASE awx` and rerun the migration
  - ~~else~~
    - ~~install `python3` and `django` and rerun the migration~~
    - ~~or `DROP DATABASE awx` and rerun the migration~~

-----------------------------------------

EDIT:

Removed all the bits about decrypting in python since @NickLaMuro built a decrypt class in ruby :tada:


spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+    credentials.each do |cred|
+      data = cred["awx_attributes"].to_json
+      stub_awx_cred_for_id(connection, cred["id"], [{"inputs" => data}])
+    end
+  end
+
+  def expect_credential_decrypts
+    stubs = YAML.load_file(data_dir.join("awesome_stubs.yaml"))
+    stubs.each do |h|
+      result = AwesomeSpawn::CommandResult.new("python3 standalone_decrypt.py", h[:output], "", 0)
```
Fryguy: I thought awesome_spawn had helper methods for this.

Fryguy: See https://github.com/ManageIQ/awesome_spawn/blob/master/lib/awesome_spawn/spec_helper.rb - does that solve it?

Fryguy: Example usage from the bot: https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/minigit_service_spec.rb#L23-L24

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+    JSON.parse(cred_info["inputs"])
+  end
+
+  def awx_connection
+    @awx_connection ||= PG::Connection.new(ApplicationRecord.connection.raw_connection.conninfo_hash.merge(:dbname => "awx").delete_blanks)
+  end
+
+  def update_authentication(auth, awx_info)
+    auth.options = auth.options.slice(*OPTIONS_FIELDS.map(&:to_sym))
+    auth.options = nil if auth.options.empty?
```
Fryguy: These 2 lines can simplify to 

```ruby
auth.options = auth.options.slice(*OPTIONS_FIELDS.map(&:to_sym)).presence
```

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+    return value unless value.include?("$encrypted$")
+    env = {
+      "FIELD_NAME"  => field,
+      "VALUE"       => value,
+      "SECRET_KEY"  => secret_key,
+      "PRIMARY_KEY" => pk
+    }
+    result = AwesomeSpawn.run("python3", :params => [script_path], :env => env)
+
+    raise "Failed to decrypt #{field} for credential #{pk}" if result.failure?
```
Fryguy: Should we just use `run!` since we don't actually expect this expectation?

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
```diff
@@ -0,0 +1,58 @@
+import base64
```
Fryguy: I think we should somehow reference the source of this code, in case we have to follow it back later.

Fryguy:
file name typo - 
`spec/migrations/data/20190617191109_move_awx_credentials_to_authentications_spec/initial_vmdb_authenticaions.yaml`
should be
`spec/migrations/data/20190617191109_move_awx_credentials_to_authentications_spec/initial_vmdb_authentications.yaml`

Fryguy:
> 💣 💥 🔥 🚒 - Bad Migration Timestamp

That feels like a bug in the bot...maybe because it's a python file?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: Given that we have passwords in plaintext in the ENV hash I wanted tighter control of the error message.

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
carbonin: I did reference it in the commit message https://github.com/ManageIQ/manageiq-schema/pull/385/commits/89787bad5841b54d571221aa39c4eac15cd531bf but I can definitely put something in the source as well.

Do you have a format in mind?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+    client
+    domain
+    host
+    project
+    subscription
+    tenant
+  ].freeze
+
+  def up
+    embedded_ansible_authentications.each do |auth|
```
bdunne: Should this be using `find_each` to work in batches?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::ScmCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::VaultCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::AmazonCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::AzureCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::GoogleCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::OpenstackCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::RhvCredential
+      ManageIQ::Providers::EmbeddedAnsible::AutomationManager::VmwareCredential
+    ]
+    Authentication.where(:type => types)
```
bdunne: This should probably be scoped to the current region to avoid the global region trying to migrate credentials from a remote region, right?

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
carbonin: So a few things there...

https://github.com/ManageIQ/awesome_spawn/blob/master/lib/awesome_spawn/spec_helper.rb#L14-L20 just looks broken.

And the bot just vendors the file https://github.com/ManageIQ/miq_bot/blob/master/spec/support/awesome_spawn_helper.rb

So I can go ahead and fix that file up, but I'm not sure if we want to hold this for a new version of awesome spawn.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+      next unless auth.manager_ref
+
+      say_with_time("Migrating credential #{auth.name} from awx to vmdb") do
+        awx_info = awx_credential_info(auth.manager_ref)
+        update_authentication(auth, awx_info)
+      end
+    end
+  rescue PG::ConnectionBad
+    say("awx database is unreachable, credentials cannot be migrated")
+    return
```
bdunne: Is this `return` needed?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: I'm expecting a maximum of something like 10 or 20 of these, but sure.

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
carbonin: Also even with this (I did the same thing as the bot and copied the file into `spec/support`) I would need to build out the path to the python script to get the stub right rather than using `an_array_including` plus add env support to the existing awesome spawn spec helper.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    return value unless value.include?("$encrypted$")
+
+    env = {
+      "FIELD_NAME"  => field,
+      "VALUE"       => value,
+      "SECRET_KEY"  => secret_key,
+      "PRIMARY_KEY" => pk
+    }
+    result = AwesomeSpawn.run("python3", :params => [script_path], :env => env)
```
NickLaMuro: First off, I expect this comment be rejected with a "too much work response" but...

Curious if is possible to write the script in a way where we aren't doing a process spawn for every record individual record in the awx auth table.  Might make it more insecure, so not really sure if that is worth it, just curious if that might things a bit more efficient.

Do we want to possibly check with some folks over on the tower team to see if they have a reference database/value available of the max number of auth records they have seen.  I don't expect we would even have someone even close to matching that value using `EmbeddedAnsible` with MIQ, but it would give us an upper bound to consider for this migration.

NickLaMuro: Also, is there any `AwesomeSpawn` logging that we need to be disabling or filtering here?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,143 @@
...
+  def awx_connection
+    @awx_connection ||= PG::Connection.new(ApplicationRecord.connection.raw_connection.conninfo_hash.merge(:dbname => "awx").delete_blanks)
+  end
+
+  def update_authentication(auth, awx_info)
+    auth.options = auth.options.slice(*OPTIONS_FIELDS.map(&:to_sym)).presence
+
+    awx_info.each do |k, v|
+      if OPTIONS_FIELDS.include?(k)
+        auth.options ||= {}
```
NickLaMuro: Confused by this line a bit I guess.  This should only need to run once, so is there is there a reason it isn't under line  87?

Also, seems like we should just be putting this guard in with the original `presence` line if it can possibly "turn `nil`"

```diff
- auth.options = auth.options.slice(*OPTIONS_FIELDS.map(&:to_sym)).presence
+ auth.options = auth.options.slice(*OPTIONS_FIELDS.map(&:to_sym)).presence || {}
```

Because there is no way you can do `auth.options.slice` if `auth.options == nil`, so my guess is that `nil` isn't a possible value for `auth.options` naturally, right?

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
NickLaMuro: A few questions to piggy back off of this:

Maybe github or some other direct links to source above each method you are ripping off here? (I am under the assumption that is what you did based on your commit message, without looking at the source yet)

Can we put the assumed `django` and `ansible` versions in here as reference?  While I expect these libs to be pretty static, it would be helpful in the scenario where a customer may have upgraded `ansible` on the appliance, and now we have different versions to contend with.

spec/migrations/data/20190617191109_move_awx_credentials_to_authentications_spec/awesome_stubs.yaml
```diff
@@ -0,0 +1,237 @@
...
+    SECRET_KEY: ecad5764714a254a619d74ccc1c4387b
+    PRIMARY_KEY: '3'
+  :output: |
+    -----BEGIN RSA PRIVATE KEY-----
```
NickLaMuro: Incase anyone is curious, I provided this data, and just used a SSH key that I generated for a `vagrant` environment I was using for `EmbeddedAnsible` testing.  Generated at the top of my vagrant file by doing the following:

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Setup keys for inter-VM communitcation to the :share VM
_, priv_key, pub_key = [nil, nil, nil]
unless File.exist? "share.id_rsa" and File.exist? "share.id_rsa.pub"
  require "vagrant/util/keypair"

  _, priv_key, pub_key = Vagrant::Util::Keypair.create

  File.write "share.id_rsa",     priv_key
  File.write "share.id_rsa.pub", pub_key
end
```

Private keys used here are from the `shared.id_rsa` file, which I will be deleting shortly.

... also, this message will self destruct in 5 seconds. :boom:

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+  def stub_awx_credentials(connection)
+    credentials = YAML.load_file(data_dir.join("credential_attributes.yaml"))
+    credentials.each do |cred|
+      data = cred["awx_attributes"].to_json
+      stub_awx_cred_for_id(connection, cred["id"], [{"inputs" => data}])
+    end
+  end
+
+  def expect_credential_decrypts
```
NickLaMuro: I guess I more have a concern about this from a higher level than @Fryguy 's concern below:

Is there a reason we are even bothering testing that `AwesomeSpawn` is receiving the proper arguments?  Isn't that just an implementation detail, and as long as stuff doesn't blow up, there is no point in testing it.  Specifically if we are potentially going to switch over to a ruby solution (if time permits).

Personally, I would rather see that the `python` script is being exercised, and that we receive the expected input and output from the migration.

NickLaMuro: Actually, I might be able to answer this myself:  My guess is we don't have `django`/`ansible` on CI, so the `python` script is probably never exercised on travis.

The above comment probably can be ignored, since I really don't have a good alternative for you on that one... 😞

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: Looks like the default batch size is 1000, for `find_each` so I really don't think it's necessary.

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
carbonin: The version of ansible isn't really relevant as this is and ansible-tower only detail and we control the version of tower.

I'm installing django `2.2.2` in https://github.com/ManageIQ/manageiq-appliance-build/pull/324, but can reference that here as well.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,142 @@
...
+    domain
+    host
+    project
+    subscription
+    tenant
+  ].freeze
+
+  def up
+    embedded_ansible_authentications.each do |auth|
+      next unless auth.manager_ref
```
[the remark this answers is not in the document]
carbonin: > Are we going to be cleaning any of these up?

I was going to change the manager refs in a separate migration.

> Should we maybe at least output when an auth record of this type is being skipped?

Probably, I can add some output for these cases (even though I'm fairly sure it's impossible or, rather incredibly unlikely).

> Along the same line, do we want to report about credentials that exist in the awx database, but don't in MIQ?

I don't think so as these are not something that the customer can actually make use of currently so I think mentioning them might just add confusion.

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
NickLaMuro: @carbonin for what it is worth, it seems that `awx` uses `django-1.11.1` on the reference appliance I am using.

> The version of ansible isn't really relevant


Yup, derp on my part.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: Actually I have a test case to change `options` to `nil` if we don't have any [[ref]](https://github.com/ManageIQ/manageiq-schema/pull/385/files#diff-b1b98e5c9ff8bdf8d7f905d42329aaceR91). And the column is nullable so I should take the nil case into account here.

db/migrate/data/20190617191109_move_awx_credentials_to_authentications/standalone_decrypt.py
carbonin: I'm seeing version `1.11.16` on my appliance, but yeah, good point. Let me change that over there.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: > just curious if that might things a bit more efficient.

I'm sure it would, but like I said with the find in batches comment, I'm not expecting a ton of these.

> Do we want to possibly check with some folks over on the tower team to see if they have a reference database/value available of the max number of auth records they have seen.

Asked

> is there any AwesomeSpawn logging that we need to be disabling or filtering here?

Maybe? It looks like it will log the stdout if the stderr is empty and the script fails, so I guess that could have something sensitive in it.

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
carbonin: Okay, opened a PR to fix the helper https://github.com/ManageIQ/awesome_spawn/pull/46

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
carbonin: They're saying that they see "from dozens to hundreds right now". So I'm not sure it's worth it given that in a very non-scientific test I'm seeing about 10 decrypts per second locally using this script.

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
carbonin: So I can write up some specs that will skip if we don't have the dependencies we need.

If most people will run embedded ansible with runner locally than it's possible that the specs will run in a fair number of cases.

I can also look into installing what we need on CI ...

carbonin:
I'm actually going to put this as WIP and add the manager ref munging to this PR.

I don't want there to even be a chance that we can run the other migration that resets the manager ref before this one so I'd rather do it all in the same transaction.

carbonin:
Added a spec that verifies the script behavior with the same data set we use to stub the AwesomeSpawn calls.

If the command `python3 -c "from django.utils.encoding import smart_str, smart_bytes"` fails on the system the spec will skip. That should ensure that we don't fail CI or mess up anyone who wants to run all the specs locally.

Let me know just how bad of an idea this is @NickLaMuro @Fryguy :sweat_smile:

NickLaMuro:
@carbonin I saw the:

```
[WIP] Move this into the migration context because there's no way this passed on CI ... I'm convinced it's not running
```

commit fly by in my email, but it seems like it isn't here now.  Are you now "convinced" it is working as expected?

NickLaMuro:
@carbonin So... yeah, I got about 90% into my response to:

> Let me know just how bad of an idea this is @NickLaMuro @Fryguy 😅

(I will paste what I almost wrote below for reference)

And then [this happened](https://github.com/carbonin/manageiq-schema/compare/migrate_awx_credentials_to_authentications...NickLaMuro:migrate_awx_credentials_to_authentications_in_ruby)...

Now, first off, this isn't some kind of "told you it was possible in ruby!" bragging (though, I would be lying if I am not happy I got it working).  Instead, I was considering while writing my response:

- suggesting that we vendor just the libs we need for this instead of installing on an appliance in the https://github.com/ManageIQ/manageiq-appliance-build/pull/324 PR
- dove into what was involved for that and understanding the code there...
- realized that you are using `python3` for this, but `python2.7` is what is vendored on our appliance for `awx`...
- realized most of the included code is adapters for python2 to python3 and other safety stuff like that
- ????
- and then it is now 1:40am and I am typing this after getting something working in Ruby...

Anyway, we can discuss "later this morning", but thought I would share what I got working and we can determine if it is good enough to move forward with, and determine how we should integrate it.

* * *

* * *

* * *

And this is what I was about to respond with...

> First off, I like this idea.  👍 
> 
> Second:  Now that these specs have a guard so they aren't being running against hosts where the `django.utils` aren't available, is there a reason that we should be running this as a "standalone_decrpyt.py"?
> 
> Curious if it just makes sense to now un-stub things everywhere, and just skip the entire spec file if we can't run with the decrypt script.  Seems to me that is the main thing we want to be testing here.
> 
> Along with that, is this something that we should add an override for when it comes to the build? (do we even run tests off TRAVIS for the build?)  Reason I say that is yes, for the daily build, I think this is a perfect solution.  But for the build I would want to make sure we are running this spec prior to a release with the utils in place, otherwise blow up and fail.
> 
> Also also:  does it maybe just make sense to vendor `django.utils` (and necessary supporting libs) for the purposes of this?  Then we can simplify this to maybe just check is `python3` exists...

carbonin:
That's awesome @NickLaMuro Thanks!

So to be sure about the behavior I created a test that encrypts and decrypts random assortments of values
https://gist.github.com/carbonin/0a189a4bdf45e180e71f3f8e4bc54c9e

I'm running it until I leave the office today and if it lasts then I figure we can be pretty sure this will work :sweat_smile:

carbonin:
Also, @Fryguy suggested using the tests from MiqPassword as a starting point for some specs for this PR. If you clean up the commit message and make a PR to my branch I'll merge it and get on that.

NickLaMuro:
> If you clean up the commit message and make a PR to my branch I'll merge it and get on that.

Sure, though I don't mind if you just want to rebase this in without attributing me.  That is fine as well.

I mostly want to:

- Add documentation for where I found things, pinned to specific versions in use (where possible...)
- Move this to an external file, outside of the migration (in case we want to use it again)
- Maybe a few cleanup things

You already did the heavy lifting of brute force testing this, so I feel a bit better about this already.

carbonin:
> Move this to an external file, outside of the migration (in case we want to use it again)

I don't actually think we want to do this. Because migrations are a snapshot in time I don't really want this to be able to change out of sync with the migration itself.

carbonin:
I can do the cleanup. But I'd like to keep the commit around so if you just remove the `[WIP] [FIXUP]` bits from the message I'll cherry-pick then carry on from there.

NickLaMuro:
@Fryguy could you add the `embedded ansible` label to this repo as well for tracking purposes?

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+            :value => "şŞ",
+            :encrypted => "$encrypted$UTF8$AESCBC$Z0FBQUFBQmRFN1htOUdENV9Eendob0s3NVRSRjN1Yjd4T25JZmthbEYtVWtSbEdLV0pOcjhuTlluRjBOeUFXelV2OUhtWFFCVzE2MzV6ZE1DZmVJRFE1YWJSVDlOUndfcVE9PQ=="
+          }
+        ]
+      end
+
+      it "decrypts special cases successfully" do
+        cases.each do |c|
+          decrypted = described_class::AnsibleDecrypt.decrypt("password", c[:encrypted], "1")
+          expect(decrypted.force_encoding('utf-8')).to eq(c[:value])
```
carbonin: Wondering if the fact that I needed to add this `force_encoding` call here means that we also need to add it in the actual migration?

NickLaMuro: I think you beat me to this (sorry, long lunch after the above), but yeah, that was pretty much the role of the `django.utils.encoding` for the most part.  So seems like this is a reasonable thing to have to do.

NickLaMuro: So with this, this might be something we are seeing just because we aren't accessing it with `ActiveRecord`, but unit testing the method specifically.

The other thing I wanted to check was what the validations where on the `awx` end, and see if it is something there is a character specification there.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,327 @@
...
+      @cipher = OpenSSL::Cipher::AES256.new(:CBC)
+      @hmac   = OpenSSL::HMAC.new signing_key, OpenSSL::Digest::SHA256.new
+    end
+
+    # This is from the main bulk of the Fernet#decrypt method in the
+    # cryptography python lib:
+    #
+    #   https://github.com/pyca/cryptography/blob/ed5a7717/src/cryptography/fernet.py#L105-L122
+    #
+    # The verify bit will be described below, but the bluk of what is in the
```
Fryguy: Typo `bluk`

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,327 @@
...
+    # cryptography lib can simply be replaced with the cipher decrypt
+    # functionality in Ruby's OpenSSL implementation.
+    #
+    # Since the fernet lib is also in charge of a decent amount of FFI
+    # interplay that the OpenSSL equivalent in Ruby takes care of for us in
+    # it's C implementation directly, so if this looks a bit simpler, that is
```
Fryguy: Typo `it's` -> `its`

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,327 @@
...
+    # python and differs from the notation in ruby, there looks to be "off by
+    # one" errors all over the place, but that is intentional.  For example,
+    # for the `cipher.iv` assignment, in python it is:
+    #
+    #   iv = data[9:25]
+    #
+    # Which translates to "start at the 9th char, and bring back everything to
+    # the 25 char", where in ruby it is:
+    #
+    #   iv = data[9,16]
```
Fryguy: Ruby can do `data[9..25]`, so I'm curious why you chose to use the `,` notation instead?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -0,0 +1,327 @@
...
+      cipher.key = encryption_key
+      ciphertext = data[25..-33]
+
+      decrypted  = cipher.update ciphertext
+      decrypted << cipher.final
+    end
+
+    private
+
+    # This method is a smaller portional of the Fernet#decrypt method from the
```
Fryguy: Typo `portional` -> `portion`?

Fryguy:
@NickLaMuro This conversion to Ruby is AWESOME.  Great work!

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: It was 12am this morning and what made sense at the time?

Slightly more accurate/serious answer:  mostly to avoid a `range` object allocation and instead use just integers... #overOptimizingMmmkay

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: I don't know... I kinda like `bluk` better... has a fun ring to it!

carbonin:
Okay, fixed up a few commits and fixed the typos. I think this is finally ready to go.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: For what it is worth, the reason why I did switch between is the `[start:end]` syntax for this in python ("slicing" I think it is called) is a little more flexible than the ruby `[start, length]` syntax.

In python, you are able to omit either the `start` or the `end` values, and it assume the beginning or the end of the string.  In Ruby, that isn't available in this form, and you only can simulate the functionality with the `[start..end]` syntax.  That said, `start` and `end` are still required for a range value, you just are allowed to use negative numbers.

The `ciphertext = data[25..-33]` is a better example of this, since the python equivalent was `ciphertext = data[25:-32]`, so a bit of a minor "off by one" difference in implementation even when using a "more similar" interface.

Fryguy: Thanks for the explanation.

> start and end are still required for a range value

FYI, soon...

2.6 - https://ruby-doc.org/core-2.6/Range.html#class-Range-label-Endless+Ranges
2.7 - https://ruby-doc.org/core-2.7.0.preview1/Range.html#class-Range-label-Beginless-2FEndless+Ranges

PR #390, opened 2019-07-09
Remove serialization of non-existing classes/objects

jrafanie:
Rails 5.0.0 - 5.0.6 serialized internal classes that don't exist in rails
5.1+.

For example, ActiveModel::Type::Text was moved to ActiveRecord so
anything that serialized the raw_attributes, which contain lots of
bloated information, and this class was in the raw_attributes, we would
be unable to deserialize it using rails 5.1+.

Additionally, if the notificiation has a subject in the options column
that no longer exists, therefore we have to also remove these.

See also:
https://github.com/rails/rails/issues/25145

### DIFF
--- a/db/migrate/20190708192323_fix_unserializable_notification_options.rb
+++ b/db/migrate/20190708192323_fix_unserializable_notification_options.rb
@@ -1,0 +1,39 @@
+class FixUnserializableNotificationOptions < ActiveRecord::Migration[5.1]
+  class Notification < ActiveRecord::Base
+    belongs_to :notification_type, :class_name => "FixUnserializableNotificationOptions::NotificationType"
+  end
+
+  class NotificationType < ActiveRecord::Base
+  end
+
+  # only used for testing
+  class CloudVolume < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    invalid_options_ids = []
+    removed_records_ids = []
+    Notification.where.not(:options => nil).each do |n|
+      begin
+        # Options could have no longer available activerecord/activemodel classes from rails 5.0.0 -> 5.0.6, which will raise an ArgumentError on deserialization
+        # Fixing these to use 5.0.7+ smaller serialization format without all the private attributes is too hard, especailly since you can only read their YAML strings
+        # and make changes as a string, not as a Hash.
+        opts = YAML.load(n.options)
+      rescue ArgumentError
+        invalid_options_ids << n.id
+      else
+        begin
+          opts_subject = opts[:subject]
+          opts_subject.reload if opts_subject.respond_to?(:reload)
+        rescue ActiveRecord::RecordNotFound
+          # Options are readable. Options[:subject] is a AR::Base object but has been deleted.
+          removed_records_ids << n.id
+        end
+      end
+    end
+
+    Notification.where(:id => invalid_options_ids).update_all(:options => {:subject => "Unreadable"}.to_yaml)
+    Notification.where(:id => removed_records_ids).update_all(:options => {:subject => "Removed"}.to_yaml)
+  end
+end
--- a/spec/migrations/20190708192323_fix_unserializable_notification_options_spec.rb
+++ b/spec/migrations/20190708192323_fix_unserializable_notification_options_spec.rb
@@ -1,0 +1,66 @@
+require_migration
+
+describe FixUnserializableNotificationOptions do
+  let(:notification_stub)      { migration_stub(:Notification) }
+  let(:notification_type_stub) { migration_stub(:NotificationType) }
+
+  migration_context :up do
+    it "fixes options[:subject] classes that don't exist" do
+      notification1 = notification_stub.create!(:notification_type => notification_type_stub.create!)
+      notification1.update(
+        :options =>
+          <<~OPTIONS
+            ---
+            :subject: !ruby/object:NonExistingCloudVolume
+              concise_attributes:
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: id
+                value_before_type_cast: 3
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: type
+                value_before_type_cast: NonExistingCloudVolume
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: name
+                value_before_type_cast: MyCloudVolume
+              new_record: false
+              active_record_yaml_version: 2
+          OPTIONS
+      )
+
+      expect { YAML.load(notification1.reload.options) }.to raise_error(ArgumentError)
+
+      migrate
+
+      expect(YAML.load(notification1.reload.options)[:subject]).to eq("Unreadable")
+    end
+
+    it "Sets options[:subject] where the record no longer exists" do
+      notification1 = notification_stub.create!(:notification_type => notification_type_stub.create!)
+      notification1.update(
+        :options =>
+          <<~OPTIONS
+            ---
+            :subject: !ruby/object:FixUnserializableNotificationOptions::CloudVolume
+              concise_attributes:
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: id
+                value_before_type_cast: 3
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: type
+                value_before_type_cast: ManageIQ::Providers::Openstack::CloudManager::CloudVolume
+              - !ruby/object:ActiveRecord::Attribute::FromDatabase
+                name: name
+                value_before_type_cast: MyCloudVolume
+              new_record: false
+              active_record_yaml_version: 2
+          OPTIONS
+      )
+
+      expect { YAML.load(notification1.reload.options)[:subject].reload }.to raise_error(ActiveRecord::RecordNotFound)
+
+      migrate
+
+      expect(YAML.load(notification1.reload.options)[:subject]).to eq("Removed")
+    end
+  end
+end


db/migrate/20190708192323_fix_unserializable_notification_options.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def up
+    Notification.where.not(:options => nil).includes(:subject).all.each do |n|
+      begin
+        # Options could have no longer available activerecord/activemodel classes from rails 5.0.0 -> 5.0.6, which will raise an ArgumentError on deserialization
+        # Fixing these to use 5.0.7+ smaller serialization format without all the private attributes is too hard, especailly since you can only read their YAML strings
+        # and make changes as a string, not as a Hash.
+        opts = YAML.load(n.options)
+      rescue ArgumentError
+        new_value = n.subject.try(:name) || n.subject.try(:description)
```
jrafanie: Because subject is polymorphic, I believe I have to enumerate all classes that could be the subject of the notification as migration stubs to avoid autoloading the class from the filesystem.  With provider classes, this could be a large number of classes.  Do we care about that here?

jrafanie: I can avoid the call to `.subject` entirely and just blank out the options here and 👇 on line 37...

jrafanie: tldr;  I tried to retain anything we have in the notification to help identify the subject of the notification... this could cause a bloated migration with all possible polymorphic subject base classes or I can just give up and clear the options if it's unreadable.

db/migrate/20190708192323_fix_unserializable_notification_options.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        # We have no option but to blow away whatever was in options and set the subject information to what's in the associated subject object, even if it's been removed.
+        # puts "Notification: #{n.id} options are not loadable, setting options[:subject] from subject association: #{n.subject_type}: #{n.subject_id} with value: #{new_value}"
+
+        # Must use update_all to avoid callbacks and checks on this instance.
+        n.class.where(:id => n.id).update_all(:options => {:subject => new_value}.to_yaml)
```
bdunne: Would `n.update_column` work?  Does it make the database interaction better or worse?

db/migrate/20190708192323_fix_unserializable_notification_options.rb
```diff
@@ -0,0 +1,44 @@
+class FixUnserializableNotificationOptions < ActiveRecord::Migration[5.1]
+  class Notification < ActiveRecord::Base
+    belongs_to :subject, :polymorphic => true
```
Fryguy: This is really dangerous because it can load arbitrary classes

jrafanie: Yeah, did you see my comment above?  If we abandon trying to recover the data and blow away options, we don't need this.  I was trying to capture anything I could from the existing notification.

db/migrate/20190708192323_fix_unserializable_notification_options.rb
jrafanie: No, I tried many things on the instance but the dirty tracking blows up because it wants to evaluate if any attributes are changed.  Unfortunately, dirty tracking still occurs if you avoid callbacks.

db/migrate/20190708192323_fix_unserializable_notification_options.rb
Fryguy: I read your comment, but it sounded like enumerating all classes was still a possibliity.

carbonin:
Is this solving two distinct issues or is there some relation between the serialization issue and the subject not existing?

jrafanie:
> Is this solving two distinct issues or is there some relation between the serialization issue and the subject not existing?

Yes, there are two distinct issues.  In the database I had...

* A few rows with serialized classes in the `options` column that fail to deserialize because they no longer exists in rails 5.1
* A few rows with both a `subject` association and a `options[:subject]` containing a serialized object (Vm, CloudVolume, etc.) that doesn't exist anymore (the record was removed).  I need to open a PR to fix the notification model / notification UI code to handle this record not existing, and figure out why we're serializing objects in `options[:subject]` instead of just a name to be used in the notification.... but fixing this in the database seemed appropriate... what do you think?

db/migrate/20190708192323_fix_unserializable_notification_options.rb
jrafanie: If we decide it's worth it to try to get this information corrected, I could remove the `belongs_to :subject` association and instead determine the `migration_stub` model from the `subject_type`, then find it using the stub using the `subject_id`... that should work as long as I enumerate all the base models as stubs.  Is this worth it?

jrafanie:
Ok, @bdunne @carbonin and I discussed this:
* 2 rows from several hundreds were affected in an internal testing database
* we normally purge notifications older than 1 week  (starting with hammer-1 / 5.10) [here](https://github.com/ManageIQ/manageiq/commit/382008678278e2a49a34c117c1c05058870c8218) 
* rails 5.0.0-5.0.6 releases were only in gaprindashvili 7 and older (< 5.9.9.0)

It seems best to just clear these bad options columns instead of trying to instantiate and possibly fix them.   So, now we just mark them as "Unreadable" or Removed".

db/migrate/20190708192323_fix_unserializable_notification_options.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  # only used for testing
+  class CloudVolume < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    invalid_options_ids = []
+    removed_records_ids = []
+    Notification.where.not(:options => nil).all.each do |n|
```
jrafanie: note, I changed this back to `.all.each` instead of `.find_each` because the latter fails with the lab database with:

```
PG::SyntaxError: ERROR:  zero-length delimited identifier at or near """"
LINE 1: ..."."options" IS NOT NULL) ORDER BY "notifications"."" ASC LIM...
                                                             ^
: SELECT  "notifications".* FROM "notifications" WHERE ("notifications"."options" IS NOT NULL) ORDER BY "notifications"."" ASC LIMIT $1
```

🤷‍♂

jrafanie:
Ok, @carbonin @bdunne ready for your 🍅

db/migrate/20190708192323_fix_unserializable_notification_options.rb
carbonin: I think you can do just `.each`, no need for the `all` when you already have a relation.

jrafanie: yeah, you're right... I was 😢 when `not(...).find_each` didn't work

jrafanie: ok, fixed

db/migrate/20190708192323_fix_unserializable_notification_options.rb
```diff
@@ -0,0 +1,39 @@
...
+          opts_subject.reload if opts_subject.respond_to?(:reload)
+        rescue ActiveRecord::RecordNotFound
+          # Options are readable. Options[:subject] is a AR::Base object but has been deleted.
+          removed_records_ids << n.id
+        end
+      end
+    end
+
+    Notification.where(:id => invalid_options_ids).update_all(:options => {:subject =>  "Unreadable record"}.to_yaml)
+    Notification.where(:id => removed_records_ids).update_all(:options => {:subject =>  "Removed record"}.to_yaml)
```
carbonin: Maybe just "Removed"/"Unreadable" here, not sure "record" adds anything for the user.

jrafanie: done

carbonin:
@jrafanie were you able to track down where we were creating a notification with an AR object in the `options`? It doesn't really matter for this PR, but I'm curious.

jrafanie:
@carbonin first one is from here (serialized classes that no longer exist):  https://github.com/ManageIQ/manageiq-providers-openstack/blob/ac0b2d4283eb1e0ee98a03949416574674f16d6b/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb#L13

second one is here (options[:subject] was removed):
https://github.com/ManageIQ/manageiq-providers-openstack/blob/ac0b2d4283eb1e0ee98a03949416574674f16d6b/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb#L70

It's still a problem.  

It looks like it was added [here](https://github.com/ManageIQ/manageiq-providers-openstack/pull/85) and has never been changed.

Note, the first notification (vm_destroy_success) has a subject (vm).  the second one (cloud_volume_delete_success), doesn't.  Note, I don't see them setting the subject association in most places.

Does anyone work on openstack to fix this?  I'm not sure if they're using `options[:subject]` at all.  It would be great to just use the association and not store the `options[:subject => self`.

```
02:19:11 ~/Code/manageiq (master) (2.5.5) - grep -r -A 3 with_notification /Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_tenant.rb:    with_notification(:cloud_tenant_delete, :options => {:subject => self}) do
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_tenant.rb-      ext_management_system.with_provider_connection(connection_options) do |service|
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_tenant.rb-        service.delete_tenant(ems_ref)
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_tenant.rb-      end
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb:    with_notification(:cloud_volume_attach,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                        :subject =>       self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                        :instance_name => server_ems_ref,
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb:    with_notification(:cloud_volume_detach,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                        :subject =>       self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb-                        :instance_name => server_ems_ref,
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb:    with_notification(:cloud_volume_create,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                        :volume_name => options[:name],
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb:    with_notification(:cloud_volume_update,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb:    with_notification(:cloud_volume_delete,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb:    with_notification(:cloud_volume_snapshot_create,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                        :snapshot_name => options[:name],
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                        :volume_name   => cloud_volume.name,
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb:    with_notification(:cloud_volume_snapshot_delete,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                        :subject       => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume_snapshot.rb-                        :volume_name   => cloud_volume.name,
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb:    with_notification(:vm_destroy,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb:  def with_notification(type, options: {}, &block)
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb:    self.class.with_notification(type, :options => options, &block)
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-  end
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-  module ClassMethods
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb:    def with_notification(type, options: {})
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-      # extract success and error options from options
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-      # :success and :error keys respectively
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/helper_methods.rb-      # with all other keys common for both cases
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb:    with_notification(:cloud_network_delete,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_subnet.rb:    with_notification(:cloud_subnet_delete,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_subnet.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_subnet.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/cloud_subnet.rb-                      }) do
--
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/network_router.rb:    with_notification(:network_router_delete,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/network_router.rb-                      :options => {
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/network_router.rb-                        :subject => self,
/Users/joerafaniello/.gem/ruby/2.5.5/bundler/gems/manageiq-providers-openstack-ac0b2d4283eb/app/models/manageiq/providers/openstack/network_manager/network_router.rb-                      }) do
```

carbonin:
@agrare do you know who we can ping about :point_up: ?

jrafanie:
@agrare to summarize:  We have a `subject` association and a serialized `options[:subject]` and using the latter caused a data integrity problem... so we'd like to use the `subject` association

agrare:
ping @aufi ^^

PR #392, opened 2019-07-12
Existing services should have a lifecycle state of provisioned

d-m-u:
Any existing service that isn't retired but is provisioned should have a lifecycle state of provisioned. 

I really wish this had been done as part of the earlier work for https://bugzilla.redhat.com/show_bug.cgi?id=1595511. 

@tinaafitz @lfu @gmcculloug 

This blocks https://github.com/ManageIQ/manageiq/pull/18943

@jrafanie please review?

### DIFF
--- a/db/migrate/20190712135032_set_provisioned_state_to_services.rb
+++ b/db/migrate/20190712135032_set_provisioned_state_to_services.rb
@@ -1,0 +1,12 @@
+class SetProvisionedStateToServices < ActiveRecord::Migration[5.1]
+  class Service < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    say_with_time("Update service lifecycle state values") do
+      Service.in_my_region.where(:lifecycle_state => nil, :retired => nil).update_all(:lifecycle_state => 'provisioned')
+    end
+  end
+end
--- a/spec/migrations/20190712135032_set_provisioned_state_to_services_spec.rb
+++ b/spec/migrations/20190712135032_set_provisioned_state_to_services_spec.rb
@@ -1,0 +1,17 @@
+require_migration
+
+describe SetProvisionedStateToServices do
+  let(:service) { migration_stub(:Service) }
+
+  migration_context :up do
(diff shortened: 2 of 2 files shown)


db/migrate/20190712135032_set_provisioned_state_to_services.rb
```diff
@@ -0,0 +1,14 @@
+class SetProvisionedStateToServices < ActiveRecord::Migration[5.1]
+  class Service < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    say_with_time("Update service lifecycle state values") do
+      Service.in_my_region.where(:lifecycle_state => nil, :retired => nil).each do |service|
```
lpichler: I think it can be  1 query 

```
Service.in_my_region.where(:lifecycle_state => nil, :retired => nil).update_all(:lifecycle_state => 'provisioned')
```

db/migrate/20190712135032_set_provisioned_state_to_services.rb
```diff
@@ -0,0 +1,12 @@
+class SetProvisionedStateToServices < ActiveRecord::Migration[5.1]
+  class Service < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    say_with_time("Update service lifecycle state values") do
+      Service.in_my_region.where(:lifecycle_state => nil, :retired => nil).update(:lifecycle_state => 'provisioned')
```
lfu: Lifecycle_state can be either 'provisioned' or 'error_in_provisioning' depending on the request status.
Is there any issue if we blindly set all lifecycle_state to 'provisioned'?

d-m-u: Tina answered this in a chat message but it was documented [here](https://github.com/ManageIQ/manageiq/pull/18943#discussion_r303019962).

carbonin: I think `update_all` would be better here as it will not instantiate all of the records by sending a single query.

PR #394, opened 2019-07-23
[WIP] Move ConfigurationScriptSource#authentication to GitRepository

Fryguy:
This is an attempt to deal with the "double ownership problem" with ConfigurationScriptSource#authentication and GitRepository.  The idea is that in the model we can delegate authentication stuff to the GitRepository.  Thus, this migration first ensures that a git_repository exists, then moves the authentication over to it.

In GitRepository model, while it might have it's own authentication via AuthenticationMixin, we also add a separate `belongs_to :authentication` which is a referenced authentication.  This way we can support both directly owned authentications and not-owned authentications.

cc @gtanzillo @carbonin @agrare @NickLaMuro

### DIFF
--- a/db/migrate/20190723023214_add_authentication_id_to_git_repository.rb
+++ b/db/migrate/20190723023214_add_authentication_id_to_git_repository.rb
@@ -1,0 +1,5 @@
+class AddAuthenticationIdToGitRepository < ActiveRecord::Migration[5.1]
+  def change
+    add_column :git_repositories, :authentication_id, :bigint
+  end
+end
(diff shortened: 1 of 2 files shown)


Fryguy:
Missing specs, but wanted to just get it out there.

carbonin:
Won't we lose the SCM credentials for existing ConfigurationScriptSources with this change? Should this also create GitRepository instances in the case that we don't have one currently? I don't see the authentication getting moved over in the create step in https://github.com/ManageIQ/manageiq/pull/19027.

PR #395, opened 2019-07-23
Add Authentication to GitRepository

Fryguy:
Replaces #394 

@carbonin Please review

cc @gtanzillo  @NickLaMuro


carbonin:
@Fryguy just to make this clear. Long-term we will want to change the existing authentication/git_repositories relationships to all be this way rather than using the `AuthenticationMixin`, but this is the way to deal with relating `ConfigurationScriptSource`s `GitRepository` records to authentications for embedded ansible.

PR #398, opened 2019-07-25
Reset column information to not have data migration stub cache conflicts 

d-m-u:
Schema changes, if the stub for that table is already loaded, all subsequent data migrations for that table may fail if they're using the cached schema. So we need to reset the cache if it gets populated so we're not using the old column information. 

This is for backport to solve the bz only, there will be an incoming LJ PR to add reset_col_information more generally than just for OrchestrationStacks. Since the specs protect against this happening here: https://github.com/ManageIQ/manageiq-schema/blob/5e66328b27ec65372daddcec5a9daeef76803721/spec/support/migration_helper.rb#L67 we should do a more general fix for this going forward. 

thanks @jrafanie 

Fixes  https://bugzilla.redhat.com/show_bug.cgi?id=1732808

### DIFF
--- a/db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
+++ b/db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
@@ -1,7 +1,12 @@
 class AddEvmOwnerToOrchestrationStacks < ActiveRecord::Migration[5.0]
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
   def change
     add_reference :orchestration_stacks, :evm_owner, :type => :bigint, :index => true
     add_reference :orchestration_stacks, :miq_group, :type => :bigint, :index => true
     add_reference :orchestration_stacks, :tenant,    :type => :bigint, :index => true
+    OrchestrationStack.reset_column_information
   end
 end
--- a/spec/migrations/20181010134649_add_evm_owner_to_orchestration_stacks_spec.rb
+++ b/spec/migrations/20181010134649_add_evm_owner_to_orchestration_stacks_spec.rb
@@ -1,0 +1,18 @@
+require_migration
+
+describe AddEvmOwnerToOrchestrationStacks do
+  let(:orchestration_stack) { migration_stub(:OrchestrationStack) }
+
+  migration_context :up do
+    it "reset column information to avoid RangeError" do
+      expect(self).to receive(:clear_caches).at_least(:once)
+      stack = orchestration_stack.create!
+
+      migrate
+
+      stack.reload
+      stack.update!(:evm_owner_id => 34_000_000_000_001)
+      expect(stack.evm_owner_id).to eq(34_000_000_000_001)
+    end
+  end
+end


d-m-u:
```bash
git grep -E "reset_column_information|add_reference"
db/migrate/20151021174044_add_tenant_default_group.rb:    Tenant.reset_column_information
db/migrate/20151104120951_transfer_rate_value_to_tiers.rb:    ChargebackRateDetail.reset_column_information
db/migrate/20160120151642_migrate_url_from_provider_to_endpoints.rb:    Endpoint.reset_column_information
db/migrate/20160226092911_separate_openstack_network_manager_from_openstack_cloud_manager.rb:    ExtManagementSystem.reset_column_information
db/migrate/20160317194215_remove_miq_user_role_from_miq_groups.rb:    MiqGroup.reset_column_information
db/migrate/20160406190650_add_cross_manager_association_to_configured_system.rb:    add_reference(:configured_systems, :counterpart, :polymorphic => true, :type => :bigint)
db/migrate/20160412173518_add_inventory_group_relation_to_configuration_script.rb:    add_reference :configuration_scripts, :inventory_root_group, :type => :bigint
db/migrate/20160428215808_add_filters_to_entitlements.rb:    Entitlement.reset_column_information
db/migrate/20160610124506_add_lives_on_to_middleware_server.rb:    add_reference :middleware_servers, :lives_on, :type => :bigint, :polymorphic => true
db/migrate/20160811081235_add_source_to_tenants.rb:    add_reference :tenants, :source, :type => :bigint, :polymorphic => true
db/migrate/20161211162552_add_generating_ems_to_event_streams.rb:    add_reference :event_streams, :generating_ems, :type => :bigint
db/migrate/20170109190411_make_job_belong_to_miq_task.rb:    add_reference :jobs, :miq_task, :type => "bigint", :index => true
db/migrate/20170313170754_add_mas_assignee.rb:    add_reference :miq_alert_statuses, :assignee, :type => :bigint, :index => true
db/migrate/20171026114327_add_deleted_on_to_container_quota_and_items.rb:    ContainerQuotaItem.reset_column_information
db/migrate/20180213204846_link_queue_and_task.rb:    add_reference :miq_queue, :miq_task, :type => :bigint
db/migrate/20180607134817_add_network_router_id_to_floating_ip.rb:    add_reference(:floating_ips, :network_router, :type => :bigint, :index => true)
db/migrate/20180618084757_add_zone_before_pause_id_to_ext_management_system.rb:    add_reference :ext_management_systems, :zone_before_pause, :type => :bigint, :index => true
db/migrate/20180817152200_add_network_router_id_to_security_group.rb:    add_reference :security_groups, :network_router, :type => :bigint, :index => true
db/migrate/20180817152201_add_cloud_subnet_id_to_security_group.rb:    add_reference :security_groups, :cloud_subnet, :type => :bigint, :index => true
db/migrate/20181003122633_add_canister_id_and_ems_ref_to_physical_disks.rb:    add_reference :physical_disks, :canister, :type => :bigint
db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb:    add_reference :orchestration_stacks, :evm_owner, :type => :bigint, :index => true
db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb:    add_reference :orchestration_stacks, :miq_group, :type => :bigint, :index => true
db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb:    add_reference :orchestration_stacks, :tenant,    :type => :bigint, :index => true
db/migrate/20181023171353_add_tenant_id_and_tenant_node_to_miq_product_features.rb:    add_reference :miq_product_features, :tenant, :type => :bigint, :index => true
db/migrate/20190201173247_add_host_id_to_switch.rb:    add_reference :switches, :host, :type => :bigint, :index => true
db/migrate/20190513155251_add_price_to_service_templates.rb:    add_reference :service_templates, :currency, :type => :bigint
db/migrate/20190520174739_add_price_to_services.rb:    add_reference :services, :currency, :type => :bigint
```

d-m-u:
@bdunne @juliancheal @jrafanie

bdunne:
@d-m-u Can you write a test to reproduce the problem?

spec/migrations/20181010134649_add_evm_owner_to_orchestration_stacks_spec.rb
```diff
@@ -0,0 +1,18 @@
+require_migration
+
+describe AddEvmOwnerToOrchestrationStacks do
+  let(:orchestration_stack) { migration_stub(:OrchestrationStack) }
+
+  migration_context :up do
+    it "reset column information to avoid RangeError" do
+      expect(self).to receive(:clear_caches).at_least(:once)
```
jrafanie: spec migration helper automatically resets the column information before you run migrate so this test would not fail before the fix... this change makes it fail

d-m-u:
@miq-bot add_label bug, ivanchuck/yes

d-m-u:
@miq-bot add_label ivanchuk/yes

db/migrate/20181010134649_add_evm_owner_to_orchestration_stacks.rb
```diff
@@ -1,7 +1,12 @@
 class AddEvmOwnerToOrchestrationStacks < ActiveRecord::Migration[5.0]
+  class OrchestrationStack < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
   def change
     add_reference :orchestration_stacks, :evm_owner, :type => :bigint, :index => true
     add_reference :orchestration_stacks, :miq_group, :type => :bigint, :index => true
     add_reference :orchestration_stacks, :tenant,    :type => :bigint, :index => true
+    OrchestrationStack.reset_column_information
```
Fryguy: This implies that a *previous* migration did not use an OrchestrationStack stub properly, I think.

jrafanie: To catch up with what our discussion was in gitter yesterday:

1) Using a stub at all causes the table schema/columns for that stub to be cached
2) If subsequent migration changed the schema of this table...
3) and another migration after this schema change tries to use a stub for this table, it would get cached information

It's unclear which migration author would know to reset column information other than the maybe the first two:
1) after you use any stub, you need to reset column information, YUCK
2) if you change the schema of a table, you should add a stub for the model and reset column information of that stub after you changed the schema (less yuck)
3) This migration is a victim... it seems odd to reset column information before using any stub... we can but YUCK

jrafanie:
An alternative approach that @d-m-u and I played with after we showed @carbonin the test code that automatically resets column information before using a stub model in migration and he suggested "why don't we do that in regular migrations... it seems like schema caching can't actually be helping us much when we're constantly changing schema (that's what migrations do)"...


```diff
diff --git a/lib/ensure_latest_column_information_helper.rb b/lib/ensure_latest_column_information_helper.rb
new file mode 100644
index 0000000..c088d2a
--- /dev/null
+++ b/lib/ensure_latest_column_information_helper.rb
@@ -0,0 +1,23 @@
+module EnsureLatestColumnInformationHelper
+  def migrate(*args)
+    reset_column_cache do
+      super
+    end
+  end
+
+  private
+
+  # Clears any cached column information on stubs before we use them
+  def reset_column_cache
+    ar_stubs.each(&:reset_column_information)
+    ActiveRecord::Base.connection.schema_cache.clear!
+    yield
+  end
+
+  def ar_stubs
+    self.class
+      .constants
+      .collect { |c| self.class.const_get(c) }
+      .select  { |c| c.respond_to?(:ancestors) && c.ancestors.include?(ActiveRecord::Base) }
+  end
+end
diff --git a/lib/manageiq/schema.rb b/lib/manageiq/schema.rb
index 472ca20..84a6474 100644
--- a/lib/manageiq/schema.rb
+++ b/lib/manageiq/schema.rb
@@ -7,6 +7,7 @@ require 'more_core_extensions/all' # TODO: Move this into specific migrations th
 
 require 'migration_helper'
 require 'migration_stub_helper'
+require 'ensure_latest_column_information_helper'
 
 require 'reserve'
 require 'reserved_shared_mixin'
diff --git a/lib/manageiq/schema/engine.rb b/lib/manageiq/schema/engine.rb
index afe3950..938f167 100644
--- a/lib/manageiq/schema/engine.rb
+++ b/lib/manageiq/schema/engine.rb
@@ -3,6 +3,11 @@ module ManageIQ
     class Engine < ::Rails::Engine
       isolate_namespace ManageIQ::Schema
 
+      ActiveSupport.on_load(:active_record) do
+        require 'ensure_latest_column_information_helper'
+        ActiveRecord::Migration.prepend(EnsureLatestColumnInformationHelper)
+      end
+
       initializer :append_migrations do |app|
         unless app.root.to_s.match root.to_s
           config.paths["db/migrate"].expanded.each do |expanded_path|
diff --git a/spec/support/migration_helper.rb b/spec/support/migration_helper.rb
index 024b392..cd3ec3c 100644
--- a/spec/support/migration_helper.rb
+++ b/spec/support/migration_helper.rb
@@ -10,7 +10,6 @@ module Spec
 
             around do |example|
               Rails.logger.debug("============================================================")
-              clearing_caches(&example)
             end
 
             it("with empty tables") { migrate }
@@ -44,12 +43,10 @@ module Spec
       end
 
       def migrate(options = {})
-        clearing_caches do
-          if options[:verbose] || ENV['VERBOSE_MIGRATION_TESTS']
-            migrate_under_test
-          else
-            suppress_migration_messages { migrate_under_test }
-          end
+        if options[:verbose] || ENV['VERBOSE_MIGRATION_TESTS']
+          migrate_under_test
+        else
+          suppress_migration_messages { migrate_under_test }
         end
       end
 
@@ -61,20 +58,6 @@ module Spec
 
       private
 
-      # Clears any cached column information on stubs, since the migrations
-      # themselves will not expect anything to be cached.
-      def clear_caches
-        ar_stubs.each(&:reset_column_information)
-        ActiveRecord::Base.connection.schema_cache.clear!
-      end
-
-      def clearing_caches
-        clear_caches
-        yield
-      ensure
-        clear_caches
-      end
-
       def ar_stubs
         described_class
           .constants
```

jrafanie:
@Fryguy ^ we probably need to investigate if there's a rails bug but considering this happens in rails 5.0, and we've been on 5.0 for a long time, I don't think the behavior has changed... the above diff is really just a POC to see if it works, which is seems to... if we should automatically do it since developers won't realize they need to do it is quite a different thing

carbonin:
> he suggested "why don't we do that in regular migrations... it seems like schema caching can't actually be helping us much when we're constantly changing schema (that's what migrations do)"...

This is a fairly accurate quote for what it's worth :laughing: 

My general opinion is that the "models" that we use in migrations should have schema caching disabled entirely. The patch @jrafanie proposed was the way I suggested we could get closest to this "ideal" situation. I don't think it's realistic to actually implement my request as stated ("disable schema caching for migration stub models").

Fryguy:
@jrafanie I would almost prefer a more blunt-instrument approach.  Since the schema cache is apparently model independent, then having the layer to find all of the stubs seems overkill when one can just call `schema_cache.clear!` after every migration

jrafanie:
> @jrafanie I would almost prefer a more blunt-instrument approach. Since the schema cache is apparently model independent, then having the layer to find all of the stubs seems overkill when one can just call `schema_cache.clear!` after every migration

Sure, testing

PR #404, opened 2019-08-01
Add initiator_type to miq_requests

lfu:
Blocks https://github.com/ManageIQ/manageiq/pull/19073

https://bugzilla.redhat.com/show_bug.cgi?id=1700692

@miq-bot add_label enhancement, ivanchuk/no


carbonin:
`initiator_type` makes it feel like a polymorphic reference. Can we choose a different name?

@bdunne looks like you were involved in the core PR, any ideas?

lfu:
How about initiated_by?

bdunne:
> `initiator_type` makes it feel like a polymorphic reference. Can we choose a different name?

Sorry, that's my fault :smile: #NamingIsHard

PR #413, opened 2019-09-09
Set zone's visible column default true and update existing rows

jrafanie:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1749694

With hammer (rails 5.0), Zone.seed was not creating the default zone with visible => true, causing it to remain nil. The hammer UI code wasn't checking this when displaying zones in the UI.

When we upgraded to Ivanchuk/rails 5.1, the UI now checks for visible and these visible => nil zones created in hammer would not be visible.  Note, Ivanchuk (rails 5.1),  Zone.seed does create the default zone with visible => true.

To be safe, this migration not only relies on postgresql's column default to set Zone visible but also updates existing records from visible => nil to visible => true.

### DIFF
--- a/db/migrate/20190909195908_set_zone_visible_column_default_true.rb
+++ b/db/migrate/20190909195908_set_zone_visible_column_default_true.rb
@@ -1,0 +1,9 @@
+class SetZoneVisibleColumnDefaultTrue < ActiveRecord::Migration[5.0]
+  class Zone < ActiveRecord::Base
+  end
+
+  def change
+    change_column_default(:zones, :visible, :from => nil, :to => true)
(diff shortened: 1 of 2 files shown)


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 3860ccd17e44391eb7e365931265f88a717e3117
Date:   Mon Sep 9 18:10:32 2019 -0400

    Merge pull request #413 from jrafanie/set_zone_visible_column_default_true
    
    Set zone's visible column default true and update existing rows
    
    (cherry picked from commit 859479cadb40cae2dfe58b4d1cf6a6c2e437017a)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1749694
```

PR #418, opened 2019-09-26
Handle faulty v2_key while upgrading ansible

kbrock:
Situation:

When importing a foreign database, the v2_key and therefore the
ansible key is not available. Which is good/ok since we shouldn't
be able to read a customer's passwords anyway.

What is bad is there is no way to work with this scenario. The database
simply can not be migrated.

Solution:

Provide a way to continue even though we can't read all the ansible
passwords.

run migration with env variable to just get bogus passwords

INVALID_ANSIBLE_VALUE=bogus rake db:migrate

original PR:
https://github.com/ManageIQ/manageiq-schema/pull/385

https://bugzilla.redhat.com/show_bug.cgi?id=1755553

### DIFF
--- a/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
+++ b/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
@@ -139,6 +139,10 @@ class MoveAwxCredentialsToAuthentications < ActiveRecord::Migration[5.0]
       return value unless value.include?("$encrypted$")
 
       new(secret_key, value, field_name, primary_key).decrypt
+    rescue Fernet256::InvalidToken
+      raise unless ENV["HARDCODE_ANSIBLE_PASSWORD"]
+
+      ENV["HARDCODE_ANSIBLE_PASSWORD"]
     end
 
     def self.secret_key
@@ -267,6 +271,15 @@ class MoveAwxCredentialsToAuthentications < ActiveRecord::Migration[5.0]
     end
   rescue PG::ConnectionBad
     say("awx database is unreachable, credentials cannot be migrated")
+  rescue Fernet256::InvalidToken
+    say("awx password invalid, credentials cannot be migrated")
+    say("This is often the case when migrating a support database")
+    say("And fix_auth has reset the ansible secret key")
+    say("")
+    say("If you want to continue and set all the invalid passwords")
+    say("to a preknown bogus value then")
+    say("run migrate again with $HARDCODE_ANSIBLE_PASSWORD=bogus")
+    raise
   end
 
   private


kbrock:
Nick, I need some help with the specs here.

@miq-bot add_reviewer NickLaMuro

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -267,6 +270,15 @@ def up
...
   rescue PG::ConnectionBad
     say("awx database is unreachable, credentials cannot be migrated")
+  rescue Fernet256::InvalidToken
+    say("awx password invalid, credentials cannot be migrated")
+    say("This is often the case when migrating a support database")
+    say("And fix_auth has reset the ansible secret key")
+    say("")
+    say("If you want to continue and set all the invalid passwords")
+    say("to a preknown bogus value then")
+    say("run migrate again with $INVALID_ANSIBLE_VALUE=bogus"
```
NickLaMuro: Looks like you are short a `)` at the end here.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -139,6 +139,9 @@ def self.decrypt(field_name, value, primary_key)
       return value unless value.include?("$encrypted$")
 
       new(secret_key, value, field_name, primary_key).decrypt
+    rescue Fernet256::InvalidToken
```
jrafanie: Is this a very specific class exception?  Is there a higher level invalid token class we can use?

NickLaMuro: @jrafanie look higher up in the file 😉 

https://github.com/ManageIQ/manageiq-schema/blob/a3a2dcfe7cd63e65b3b7ba70634247519183408e/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb#L23

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -139,6 +139,9 @@ def self.decrypt(field_name, value, primary_key)
...
+      raise unless ENV["INVALID_ANSIBLE_VALUE"]
+      ENV["INVALID_ANSIBLE_VALUE"]
```
jrafanie: Does this value need to be known later for fix_auth?  Can it be any value?

Words are hard.  Maybe HARDCODE_ANSIBLE_PASSWORD or something like that?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
jrafanie: Oh wow, I forgot we have that implemented here.  👍

jrafanie:
This workflow seems ok to me.  I'm not sure how to word this but it seems reasonable to get a database dump from someone and not caring about the actual ansible password, you just want to migrate the DB using a fake or hardcoded invalid password.

NickLaMuro:
@jrafanie @kbrock For simplicity, I created a new version of this PR with the specs added:

https://github.com/ManageIQ/manageiq-schema/pull/419

I ended up rebased my specs into the original commit, so didn't think that merging my changes into this PR would work out.  @kbrock should still have credit as the author of the commit though since it was the one that I merged into.

Hope that is okay.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
kbrock: can be any value. In fixauth, we typically choose `"bogus"`. But it need not be the same as fix_auth

PR #419, opened 2019-09-26
Handle faulty v2_key while upgrading ansible (with specs)

NickLaMuro:
This is an update to #418 which adds specs (and a few minor refactor commits to support) to ensure we are working with the new code properly.  (mostly) original PR description below:

Situation:
----------

When importing a foreign database, the v2_key and therefore the ansible key is not available.  Which is good/ok since we shouldn't be able to read a customer's passwords anyway.

What is bad is there is no way to work with this scenario. The database simply can not be migrated.

Solution:
---------

Provide a way to continue even though we can't read all the ansible passwords.

run migration with env variable to just get bogus passwords

```console
$ HARDCODE_ANSIBLE_PASSWORD=bogus rake db:migrate
```

Also adds some specs around the use cases where the `secret_key` provided doesn't exist or does not match what is provided, as well as testing the workaround in those cases.

Links
-----

- original PR: https://github.com/ManageIQ/manageiq-schema/pull/385
- https://bugzilla.redhat.com/show_bug.cgi?id=1755553

### DIFF
--- a/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
+++ b/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
@@ -139,6 +139,10 @@ class MoveAwxCredentialsToAuthentications < ActiveRecord::Migration[5.0]
       return value unless value.include?("$encrypted$")
 
       new(secret_key, value, field_name, primary_key).decrypt
+    rescue Fernet256::InvalidToken
+      raise unless ENV["HARDCODE_ANSIBLE_PASSWORD"]
+
+      ENV["HARDCODE_ANSIBLE_PASSWORD"]
     end
 
     def self.secret_key
@@ -260,13 +264,25 @@ class MoveAwxCredentialsToAuthentications < ActiveRecord::Migration[5.0]
         next
       end
 
-      say_with_time("Migrating credential #{auth.name} from awx to vmdb") do
+      say_with_time("Migrating credential #{auth.name} from AWX to vmdb") do
         awx_info = awx_credential_info(auth.manager_ref)
         update_authentication(auth, awx_info)
       end
     end
   rescue PG::ConnectionBad
-    say("awx database is unreachable, credentials cannot be migrated")
+    say("AWX database is unreachable, credentials cannot be migrated")
+  rescue Fernet256::InvalidToken
+    say("AWX password invalid, credentials cannot be migrated!")
+    say("")
+    say("This is often the case when migrating a database that was backed up")
+    say("and then restored and fix_auth has reset the ansible secret key")
+    say("")
+    say("If you want to continue and set all the invalid passwords to a")
+    say("preknown 'dummy' value then run migrate again with:")
+    say("")
+    say("   $ $HARDCODE_ANSIBLE_PASSWORD=bogus")
+    say("")
+    raise
   end
 
   private
--- a/spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
+++ b/spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
@@ -12,6 +12,7 @@ describe MoveAwxCredentialsToAuthentications do
   let(:authentication) { migration_stub(:Authentication) }
   let(:miq_database)   { migration_stub(:MiqDatabase) }
 
+  let(:awx_conn)        { instance_double(PG::Connection) }
   let(:secret_key)      { "ecad5764714a254a619d74ccc1c4387b" }
   let(:miq_database_id) { miq_database.create.id }
 
@@ -117,8 +118,6 @@ describe MoveAwxCredentialsToAuthentications do
     end
 
     context "with an awx database connection" do
-      let(:awx_conn) { instance_double(PG::Connection) }
-
       before do
         allow(PG::Connection).to receive(:new).with(a_hash_including(:dbname => "awx")).and_return(awx_conn)
       end
--- a/spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
+++ b/spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
@@ -259,34 +259,77 @@ describe MoveAwxCredentialsToAuthentications do
 
         expect_authentications_migrated
       end
+
+      context "with a 'bogus' secret key (DB restore use case)" do
+        let(:secret_key) { "invalid" }
+
+        before do
+          # load the initial set of authentication records
+          auths = YAML.load_file(data_dir.join("vmdb_authentications.yaml"))
+          auths.each { |auth| authentication.create!(auth["initial"]) }
+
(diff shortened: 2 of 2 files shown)


db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -139,6 +139,10 @@ def self.decrypt(field_name, value, primary_key)
       return value unless value.include?("$encrypted$")
 
       new(secret_key, value, field_name, primary_key).decrypt
+    rescue Fernet256::InvalidToken
```
NickLaMuro: So after thinking about this, I don't _**think**_ we need to handle the case where a `secret_key` doesn't exist in our database, since the `def up` should not even be trying to decrypt anything (even with an existing `awx` DB) since there is nothing in the `vmdb_production` to migrate.

But wanted to post this here since I ran into this as a "failing case" when trying to write the specs for this, and for a bit thought it might be relevant.

kbrock: An InvalidToken happens when the ansible key is invalid.

we have an invalid ansible key when we `fix_auth --invalid bogus` a database from a customer.

NickLaMuro: @kbrock I think you are missing the point of the concern...

> An InvalidToken happens when the ansible key is invalid.

Wasn't concerned about this error that is already there, I am concerned about a different error that I was able to "reproduce" in the specs where it tries to do the look up in `AnsibleDecrypt.secret_key`, but doesn't find anything, so a `NoMethodError` is returned on `nil`.

Again, I don't think this scenario would be possible since I don't think you can have a `EmbeddedAnsible` authentication(s) record (which would be what allows it to get that far to hit `.secret_key` in the migration) without having secret key to decrypt with.  But I wouldn't mind a second opinion on that.

NickLaMuro:
@jrafanie did you want to look at this and sign off as well?

@kbrock I assume you are okay with this one and we can close the other one, correct?

NickLaMuro:
Also, I assume this is something we want in the `ivanchuk` release, right?

kbrock:
@miq-bot add_label ivanchuk/yes, bug,

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -267,6 +271,15 @@ def up
     end
   rescue PG::ConnectionBad
     say("awx database is unreachable, credentials cannot be migrated")
+  rescue Fernet256::InvalidToken
+    say("awx password invalid, credentials cannot be migrated")
```
Fryguy: AWX should be capitalized through these sentences.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -267,6 +271,15 @@ def up
...
+    say("This is often the case when migrating a support database")
```
Fryguy: This sentence is weird in upstream, since a support database isn't a thing.  Perhaps...

```
This is often the case when migrating a database that was backed up and then restored
```

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -267,6 +271,15 @@ def up
...
+    say("And fix_auth has reset the ansible secret key")
```
Fryguy: Lower case the `A` in And.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
```diff
@@ -267,6 +271,15 @@ def up
...
+    say("")
+    say("If you want to continue and set all the invalid passwords")
+    say("to a preknown bogus value then")
```
Fryguy: The word bogus is strange....not sure of an alternative.  Perhaps "a preknown replacement password" ?

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
```diff
@@ -251,34 +259,77 @@
 
         expect_authentications_migrated
       end
+
+      context "with a 'bogus' secret key (support use case)" do
```
Fryguy: Remove the "support use case" text because it doesn't make sense in upstream.

Fryguy:
Overall this looks good to me.  Mostly commenting on the text as presented to the user.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: @Fryguy 'dummy'?

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: 👍 Fixed.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: Done.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: I #blameKeenan™

Fixed.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
NickLaMuro: I went with "dummy".  But I can change it to what you suggested if you prefer.

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
NickLaMuro: switched it to "(DB restore use case)" since I think it is still worth mentioning a distinction the spec description.

NickLaMuro:
@Fryguy Comments should be addressed now.

db/migrate/20190617191109_move_awx_credentials_to_authentications.rb
Fryguy: dummy is perfect

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit ccd4d14df7431d288763bcba5c967bcbf2a63328
Author: Jason Frey <jfrey@redhat.com>
Date:   Mon Sep 30 18:20:21 2019 +0200

    Merge pull request #419 from NickLaMuro/credentials_issue
    
    Handle faulty v2_key while upgrading ansible (with specs)
    
    (cherry picked from commit 2e27544078da01ade0b1f2205f68def6ca2a0d3c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1755553
```

PR #420, opened 2019-09-30
Connect CloudNetwork, NetworkPort and SecurityGroup to ResourceGroup

mancabizjak:
In this PR we provide three simple migrations that add a reference to a `ResourceGroup` for `CloudNetwork`, `NetworkPort` and `SecurityGroup` models. So far, only VMs have a reference to the resource group they belong to. 

This change is proposed to allow simpler inventorying process in AzureStack provider later on. In Azure and AzureStack, these entities are always managed as a part of a resource group, and we aim to support a resource group refresh target. Without this change, we would need to infer the related targets based on individual resource's `ems_ref`, which contains the resource group name.

@miq-bot assign @agrare

### DIFF
--- a/db/migrate/20190930054405_add_resource_group_id_to_cloud_network.rb
+++ b/db/migrate/20190930054405_add_resource_group_id_to_cloud_network.rb
@@ -1,0 +1,5 @@
+class AddResourceGroupIdToCloudNetwork < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :cloud_networks, :resource_group, :type => :bigint, :index => true
+  end
+end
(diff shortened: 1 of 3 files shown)


miha-plesko:
Hi @agrare some context: there was `ResourceGroup` model introduced when you guys were doing Azure provider, but it feels like the new model was never really used because:

- other models are not having foreign keys to it
- it has no UI implemented
- it has no operations implemented (like "delete entire resource group" etc.)
- it has no REST API implemented (like /api/resource_groups)

But since the Azure/AzureStack is very much resource group oriented, it only makes sense for us to support all those enhancement ASAP.

Would that make sense?

agrare:
Interestingly orchestration_stack also has a "reference" to resource_groups, but instead of a foreign_key it is a string column.  We might want to look at migrating that to a proper relationship in the future but it'll require a data-migration.

PR #421, opened 2019-10-01
Rename table chargeback_rate_detail_currencies to currencies

lpichler:
Currencies are not related just to chargeback

https://github.com/ManageIQ/manageiq-schema/pull/367#discussion_r283829528


d-m-u:
Does the replication exclude tables list need to change with this?

PR #426, opened 2019-10-24
Remove ems_ref_obj YAML column

agrare:
Currently we store a YAML serialized `VimString` `ManagedObjectReference` that includes the `vimType` (e.g. `VirtualMachine`/`HostSystem`/`ClusterComputeResource`) because the `ems_ref` and model are not enough information to accurately represent a VMware object, you need the VIM class name as well.

Rather than serializing a full `VMwareWebService/VimType` we can store the VIM class in a string e.g. `'VirtualMachine`.  This gives us enough information to indicate a `VMware` `ManagedObject` without requiring the `VMwareWebService` gem in core to allow for deserialization of `ems_ref_obj`.

https://github.com/ManageIQ/manageiq/issues/18428

### DIFF
--- a/db/migrate/20191023180130_update_openstack_hosts_uid_ems.rb
+++ b/db/migrate/20191023180130_update_openstack_hosts_uid_ems.rb
@@ -5,9 +5,7 @@ class UpdateOpenstackHostsUidEms < ActiveRecord::Migration[5.1]
   end
 
   def up
-    Host.in_my_region.where(:type => "ManageIQ::Providers::Openstack::InfraManager::Host").find_each do |host|
-      next if host.ems_ref_obj.nil?
-
+    Host.in_my_region.where(:type => "ManageIQ::Providers::Openstack::InfraManager::Host").where.not(:ems_ref_obj => nil).find_each do |host|
       host.uid_ems = YAML.safe_load(host.ems_ref_obj)
       host.save!
     end
--- a/db/migrate/20191023180130_update_openstack_hosts_uid_ems.rb
+++ b/db/migrate/20191023180130_update_openstack_hosts_uid_ems.rb
@@ -1,0 +1,23 @@
+class UpdateOpenstackHostsUidEms < ActiveRecord::Migration[5.1]
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    Host.in_my_region.where(:type => "ManageIQ::Providers::Openstack::InfraManager::Host").find_each do |host|
+      next if host.ems_ref_obj.nil?
+
+      host.uid_ems = YAML.safe_load(host.ems_ref_obj)
+      host.save!
+    end
+  end
+
+  def down
+    Host.in_my_region.where(:type => "ManageIQ::Providers::Openstack::InfraManager::Host").find_each do |host|
+      host.ems_ref_obj = YAML.dump(host.uid_ems)
+      host.uid_ems     = host.ems_ref
+      host.save!
+    end
+  end
+end
--- a/db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
+++ b/db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
@@ -45,8 +45,13 @@ class MigrateEmsRefObjToEmsRefType < ActiveRecord::Migration[5.1]
   def down
     MODELS_WITH_EMS_REF_OBJ.each do |klass|
       say_with_time("Converting ems_ref_type to ems_ref_obj for #{klass.name}") do
-        # Skip any records that have ems_ref_obj set already
-        klass.in_my_region.where(:ems_ref_type => nil, :ems_ref_obj => nil).find_each do |obj|
+        # Skip any records that have ems_ref_obj set already and don't have ems_ref_type set
+        query_params = {:ems_ref_type => nil, :ems_ref_obj => nil}
+
+        # Skip cloud VMs which never had ems_ref_obj set
+        query_params[:cloud] = [nil, false] if klass == VmOrTemplate
+
+        klass.in_my_region.where(query_params).find_each do |obj|
           obj.update!(:ems_ref_obj => "--- #{obj.ems_ref}\n")
         end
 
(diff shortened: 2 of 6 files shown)


db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
```diff
@@ -0,0 +1,61 @@
...
+    include ActiveRecord::IdRegions
+  end
+
+  MODELS_WITH_EMS_REF_OBJ = [EmsCluster, EmsFolder, Host, ResourcePool, Snapshot, Storage, VmOrTemplate].freeze
+
+  def up
+    MODELS_WITH_EMS_REF_OBJ.each do |klass|
+      say_with_time("Converting ems_ref_obj to ems_ref_type for #{klass.name}") do
+        klass.in_my_region.where("ems_ref_obj LIKE '%VimString%'").find_each do |obj|
+          ems_ref_obj  = YAML.load(obj.ems_ref_obj)
```
Fryguy: This bit is going to require that the VimString class is deserializable forever.  Is it possible to do this migration *without* the YAML parsing?

db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def down
+    MODELS_WITH_EMS_REF_OBJ.each do |klass|
+      say_with_time("Converting ems_ref_type to ems_ref_obj for #{klass.name}") do
+        klass.in_my_region.find_each do |obj|
+          ems_ref_obj = if obj.ems_ref_type.nil?
+            obj.ems_ref
+          else
+            VimString.new(obj.ems_ref, obj.ems_ref_type, :ManagedObjectReference)
```
Fryguy: Same here 😆 .

We can always manually construct the YAML if needed.

Alternately, we could copy the classes into this migration though that may actually make things worse.

db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
agrare: Definitely, I want to get the rest of the changes working then I'm going to parse the string manually so we don't need to yaml load it

db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb
agrare: Same as above, that is the plan

spec/migrations/20191023180330_migrate_ems_ref_obj_to_ems_ref_type_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      migrate
+
+      expect(vm.reload.ems_ref_type).to      eq("VirtualMachine")
+      expect(host.reload.ems_ref_type).to    eq("HostSystem")
+      expect(cluster.reload.ems_ref_type).to eq("ClusterComputeResource")
+      expect(respool.reload.ems_ref_type).to eq("ResourcePool")
+      expect(storage.reload.ems_ref_type).to eq("Datastore")
+    end
+
+    it "ignores ems_ref_obj if it isn't a VimString" do
```
Fryguy: Is it possible there are entires will nil values in ems_ref_obj?  Should there be a test for nil values?

spec/migrations/20191023180330_migrate_ems_ref_obj_to_ems_ref_type_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        str: vm-1234
+        xsiType: :ManagedObjectReference
+        vimType: :VirtualMachine
+      EMS_REF_OBJ
+    end
+
+    it "sets to a simple yaml string if ems_ref_type is nil" do
+      vm = vm_stub.create!(:ems_ref => "b7212ae9-e968-4431-bb17-cc16d5095cd0")
+      migrate
+      expect(vm.reload.ems_ref_obj).to eq("--- b7212ae9-e968-4431-bb17-cc16d5095cd0\n")
```
Fryguy: Did this happen for all providers or only specific ones?

spec/migrations/20191023180330_migrate_ems_ref_obj_to_ems_ref_type_spec.rb
agrare: It would most likely be `"--- \n"` but I'll add a test for `nil` as well just in case

agrare: [here](https://github.com/ManageIQ/manageiq-schema/pull/426/files#diff-e15ea0fdfded664940afbedd79eb324cR82)

spec/migrations/20191023180330_migrate_ems_ref_obj_to_ems_ref_type_spec.rb
agrare: Good point, cloud vms wouldn't have ems_ref_obj set.  I'll skip those

agrare: Okay skipping cloud vms now

db/migrate/20191023180130_update_openstack_hosts_uid_ems.rb
```diff
@@ -0,0 +1,23 @@
+class UpdateOpenstackHostsUidEms < ActiveRecord::Migration[5.1]
+  class Host < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    Host.in_my_region.where(:type => "ManageIQ::Providers::Openstack::InfraManager::Host").find_each do |host|
+      next if host.ems_ref_obj.nil?
```
Fryguy: You can probably push this into the query... `.where.not(:ems_ref_obj => nil)`

agrare: :+1:

agrare: Done [here](https://github.com/ManageIQ/manageiq-schema/pull/426/files#diff-13bdd75d8c9d067286df49a626dc45e1R8) and test added

PR #433, opened 2019-11-14
Remove MiqWidget with content_type set to rss

ZitaNemeckova:
There's no code that handles them but they stayed in DB.

Removed in https://github.com/ManageIQ/manageiq-ui-classic/pull/5271 and https://github.com/ManageIQ/manageiq/pull/18478 .

@miq-bot add_label ivanchuk/no, technical debt

@lpichler please have a look,  thanks :)


ZitaNemeckova:
@miq-bot add_label ivanchuk/no, technical debt

PR #435, opened 2019-11-18
Subclass ResourcePools

agrare:
There are some provider specifics to ResourcePools which should not live in core

Core issue: https://github.com/ManageIQ/manageiq/pull/19529

### DIFF
--- a/spec/migrations/20191118161319_subclass_resource_pools_spec.rb
+++ b/spec/migrations/20191118161319_subclass_resource_pools_spec.rb
@@ -1,0 +1,60 @@
+require_migration
+
+describe SubclassResourcePools do
+  let(:ext_management_system_stub) { migration_stub(:ExtManagementSystem) }
+  let(:resource_pool_stub)         { migration_stub(:ResourcePool) }
+
+  migration_context :up do
+    it "migrates resource pools from all supported providers" do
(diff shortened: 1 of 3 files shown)


spec/migrations/20191118161319_subclass_resource_pools_spec.rb
```diff
@@ -0,0 +1,42 @@
+require_migration
+
+describe SubclassResourcePools do
+  let(:ext_management_system_stub) { migration_stub(:ExtManagementSystem) }
+  let(:resource_pool_stub)         { migration_stub(:ResourcePool) }
+
+  migration_context :up do
+    it "migrates resource pools from all supported providers" do
```
bdunne: It would probably also be good to have a test for a record that isn't changed by the migration.

agrare: :+1: done

skateman:
This broke the UI :cry:, see https://github.com/ManageIQ/manageiq-ui-classic/issues/6425

agrare:
@skateman we tested this with the other PRs linked to make sure the UI was still working with all of the PRs applied.  Just got caught in the middle of a set of PRs

PR #438, opened 2019-11-20
Run bundle install in `bin/setup`

agrare:
Running manageiq-schema from cross-repo tests fails because we do not run `bundle install` from `bin/setup`

Ref: https://travis-ci.org/ManageIQ/manageiq-cross_repo-tests/jobs/614581288#L293

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,7 +1,10 @@
 #!/bin/bash
 if [[ -n "$CI" ]]; then
   echo 'gem: --no-ri --no-rdoc --no-document' > ~/.gemrc
+  bundle config --local path $(pwd)/vendor/bundle
+
   psql -c "CREATE USER root SUPERUSER PASSWORD 'smartvm';" -U postgres
-else
-  bundle exec rake spec:setup
 fi
+
+bundle update --jobs=3
+bundle exec rake spec:setup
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
+echo 'gem: --no-ri --no-rdoc --no-document' > ~/.gemrc
```
Fryguy: This needs to stay in $CI, because we don't want to mess with local .gemrc files on personal machine.

bin/setup
```diff
@@ -1,7 +1,9 @@
...
+bundle update --jobs=3
```
Fryguy: If we are doing this here, we should stop travis from doing it also i the .travis.yml

bin/setup
agrare: :+1:

Fryguy:
We should also remove the `bundle exec rake spec:setup` from .travis.yml, otherwise it will run twice.

agrare:
:+1: yeah this was just a test to see if it fixed running -schema in -cross-repo

bin/setup
agrare: I might be going blind but I don't see in .travis.yml where we call out to do bundle install, but it definitely is done https://travis-ci.org/ManageIQ/manageiq-schema/jobs/614674353#L413

agrare: :man_facepalming: nevermind, I was diff'ing the same file and wondering why nothing was different

agrare: Okay done, only bundle install once now but it looks like it isn't using the bundler cache anymore ;(

PR #440, opened 2019-12-02
Add restart_needed field to vms

gekorob:
Sometimes VMs need to be restarted to apply not hotpluggable configuration
changes and the user needs to be aware of that.
Changes to a VM can be applied through ManageIQ but also by the
external the external system

This is part of the RFE
https://bugzilla.redhat.com/show_bug.cgi?id=1572649

### DIFF
--- a/db/migrate/20191128111630_add_restart_needed_to_vms.rb
+++ b/db/migrate/20191128111630_add_restart_needed_to_vms.rb
@@ -1,0 +1,5 @@
+class AddRestartNeededToVms < ActiveRecord::Migration[5.1]
+  def change
+    add_column :vms, :restart_needed, :boolean
+  end
+end


db/migrate/20191128111630_add_restart_needed_to_vms.rb
```diff
@@ -0,0 +1,5 @@
+class AddRestartNeededToVms < ActiveRecord::Migration[5.1]
+  def change
+    add_column :vms, :restart_needed, :boolean, :null => false, :default => false
```
agrare: We should drop the `:null => false` because most providers won't set this, also we should drop the `:default => false`, we try to keep the "logic" at the rails not the db level and if it is nil it will behave falsy anyway `if nil => false`

gekorob: Yes, absolutely, I was used to that to avoid the three state problem. I'll modify that and re-push

agrare: Yeah actually personally I prefer having db-level defaults and no nils and we do that in other projects (topological-inventory for example) but in MIQ we avoid db-level defaults

PR #444, opened 2019-12-10
Fix splitting storages for archived hosts

agrare:
The spec test for the archived host would have caught this, but the
`let()` block wasn't invoked so it never created the archived host.

https://bugzilla.redhat.com/show_bug.cgi?id=1394909


agrare:
@himdel can you test with this applied?

himdel:
Works now, thanks! :)

agrare:
Awesome thanks for confirming!

PR #446, opened 2019-12-25
[wip] Add format to disks

borod108:
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1726590

### DIFF
--- a/db/migrate/20191119111806_split_disk_type_to_thin_and_format_for_disk.rb
+++ b/db/migrate/20191119111806_split_disk_type_to_thin_and_format_for_disk.rb
@@ -1,0 +1,64 @@
+class SplitDiskTypeToThinAndFormatForDisk < ActiveRecord::Migration[5.1]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Storage < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+    belongs_to :ext_management_system, foreign_key: "ems_id"
+  end
+
+  class Disk < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+    belongs_to :storage
+    has_one :ext_management_system, through: :storage
+  end
+
+  HASH_UP = {
+    'Redhat::InfraManager' => [ { disk_type_query: "disk_type = 'thin'", thin_value: true, format_value: "NULL" }, { disk_type_query: "disk_type = 'thick'", thin_value: false, format_value: "NULL"}],
+    'Vmware::InfraManager' => [ { disk_type_query: "disk_type = 'thin'", thin_value: true, format_value: "'vmdk'" }, { disk_type_query: "disk_type = 'thick'", thin_value: false, format_value: "'vmdk'"},
+                  { disk_type_query: "disk_type NOT IN ('thick', 'thin')", thin_value: false, format_value: "disk_type"}],
+    'Vmware::CloudManager' => [ { disk_type_query: "true", thin_value: true, format_value: "disk_type" } ],
+    'Microsoft::InfraManager' => [ { disk_type_query: "disk_type = 'thin'", thin_value: true, format_value: "'vhdx'" }, { disk_type_query: "disk_type = 'thick'", thin_value: false, format_value: "'vhdx'"},
+                  { disk_type_query: "disk_type NOT IN ('thick', 'thin')", thin_value: false, format_value: "'unknown'"}],
+    'Azure::CloudManager' => [ { disk_type_query: "true", thin_value: true, format_value: "disk_type" } ]
+  }
+
+  def up
+    add_column :disks, :format, :string
+    add_column :disks, :thin, :boolean
+    HASH_UP.each do |provider, queries|
+      queries.each do |query_opts|
+        Disk.in_my_region.joins(:storage).joins(:ext_management_system).where(:ext_management_systems => {type: "ManageIQ::Providers::#{provider}"})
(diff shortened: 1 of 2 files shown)


db/migrate/20191119111806_add_format_to_disk.rb
```diff
@@ -0,0 +1,5 @@
+class AddFormatToDisk < ActiveRecord::Migration[5.1]
+  def change
+    add_column :disks, :format, :string
```
agrare: @borod108 format might be a little too general, what are some of the options for this value for RHV so we can get a feeling for how it might be used by other providers?

borod108: Hi!
So format in this case is either qcow2 or raw.
We need this to validate in pre provision that we are trying to prevision something that fits the options described under: "Permitted Storage Combinations" here:
https://www.ovirt.org/documentation/admin-guide/chap-Virtual_Machine_Disks.html

agrare: Okay, that sounds like the `:disk_type` column, on vmware that stores if it is thin, thick, or rdm.

borod108: @agrare actually we use disk_type for that exactly.
These are more or less orthogonal for rhv - We can have "thin raw" disks on "file" type storage.

agrare: Oh you're right I thought I checked the RHV parser and saw it wasn't used...I must have been looking at the wrong file

agrare: I wonder if we should split up `:disk_type` into `:thin` and `:format` columns since currently it is a little overloaded.

```
          disk_type   |      thin      |     format
----------------------------------------------------
VMware |    thin      |      true      |     vddk
VMware |    thick     |     false      |     vddk
VMware | rdm-virtual  |     false      | virtual-rdm
VMware | rdm-physical |     false      | physical-rdm
Ovirt  |    thin      |     true       |     qcow
Ovirt  |    thick     |     false      |     qcow
Ovirt  |    thin      |     true       |     raw
Ovirt  |    thick     |     false      |     raw
```

borod108: sounds good to me, but I just wonder if maybe we should have "sparsity" and in it "thin/preallocated" - just to keep it generic.

agrare: Is preallocated the same thing as thick?  Trying to stay away from string columns because they tend to get less normalized aka different providers just put their own exact string there.

borod108: @agrare I thought that is a good thing, anyway - yes thick is preallocated. so should the disk_type column just be deleted?

agrare: Yes I think we should migrate disk_type into the new columns

borod108: @agrare should the "thin" value be set to false for all the VMware formats other then "thin"?

borod108: @agrare updated the PR, would you say this can work?

agrare: @borod108 I think the logic is right we just need to try to make it more performant

borod108: @agrare how about now?

agrare:
@borod108 take a look at a migration like https://github.com/ManageIQ/manageiq-schema/pull/435/files that should give you an idea of what a data migration should look like.  We also need a spec test for data migrations.

db/migrate/20191119111806_split_disk_type_to_thin_and_format_for_disk.rb
```diff
@@ -0,0 +1,17 @@
+class SplitDiskTypeToThinAndFormatForDisk < ActiveRecord::Migration[5.1]
+  def up
+    add_column :disks, :format, :string
+    add_column :disks, :thin, :boolean
+    Disk.connection.execute("UPDATE disks SET thin = (CASE disk_type WHEN 'thin' THEN true ELSE false END) WHERE disk_type IN ('thin', 'thick')")
```
agrare: You can define a Disk subclass in this migration e.g. `class Disk < ActiveRecord::Base; end`

agrare: Probably something like `Disk.in_my_region.where(:disk_type => 'thin').update_all(:thin => true)` would work
For the format you probably need to look at the EMS type to set the format e.g. vmdk/qcow/etc...

db/migrate/20191119111806_split_disk_type_to_thin_and_format_for_disk.rb
```diff
@@ -0,0 +1,64 @@
...
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+    belongs_to :ext_management_system, foreign_key: "ems_id"
+  end
+
+  class Disk < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
```
agrare: Disks don't have a `:type` column and aren't STI so you don't need this

db/migrate/20191119111806_split_disk_type_to_thin_and_format_for_disk.rb
```diff
@@ -0,0 +1,64 @@
...
+    belongs_to :storage
+    has_one :ext_management_system, through: :storage
```
agrare: I think the more "correct" association to an EMS would be through vms -> hardwares -> disks.

E.g. we should try to follow https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L72

PR #451, opened 2020-01-27
Remove VimTypes from EmsEvent#full_data

agrare:
EmsEvents have VimTypes (e.g. VimHash/VimString) in the serialized full_data column which requires core workers to require `VMwareWebService/VimTypes` to be able to load the record.

### DIFF
--- a/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
+++ b/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
@@ -1,34 +1,34 @@
-module VimType
-  def vimType
-    @vimType.nil? ? nil : @vimType.to_s
-  end
-
-  def vimType=(val)
-    @vimType = val.nil? ? nil : val.to_sym
-  end
-
-  def xsiType
-    @xsiType.nil? ? nil : @xsiType.to_s
-  end
-
-  def xsiType=(val)
-    @xsiType = val.nil? ? nil : val.to_sym
-  end
-end
-
-class VimHash < Hash
-  include VimType
-end
-
-class VimArray < Array
-  include VimType
-end
-
-class VimString < String
-  include VimType
-end
-
 class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
+  module VimType
+    def vimType
+      @vimType.nil? ? nil : @vimType.to_s
+    end
+
+    def vimType=(val)
+      @vimType = val.nil? ? nil : val.to_sym
+    end
+
+    def xsiType
+      @xsiType.nil? ? nil : @xsiType.to_s
+    end
+
+    def xsiType=(val)
+      @xsiType = val.nil? ? nil : val.to_sym
+    end
+  end
+
+  class VimHash < Hash
+    include VimType
+  end
+
+  class VimArray < Array
+    include VimType
+  end
+
+  class VimString < String
+    include VimType
+  end
+
   class EventStream < ActiveRecord::Base
     include ActiveRecord::IdRegions
 
@@ -36,13 +36,33 @@ class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
   end
 
   def up
+    vim_hash_backup   = const_replace(:VimHash, VimHash)
+    vim_string_backup = const_replace(:VimString, VimString)
+    vim_array_backup  = const_replace(:VimArray, VimArray)
+
     EventStream.in_my_region.where(:source => "VC").find_each do |event|
       full_data = YAML.load(event.full_data)
       event.update!(:full_data => vim_types_to_basic_types(full_data).to_yaml)
     end
+
+    const_replace(:VimHash, vim_hash_backup)
+    const_replace(:VimString, vim_string_backup)
+    const_replace(:VimArray, vim_array_backup)
   end
 
   private
+
+  def const_replace(const_name, new_const)
+    old_const = const_remove(const_name)
+    Object.const_set(const_name, new_const)
+    old_const
+  end
+
+  def const_remove(const_name)
+    return unless Object.const_defined?(const_name)
+    old_const = Object.const_get(const_name)
+    Object.send(:remove_const, const_name)
+  end
 
   def vim_types_to_basic_types(obj)
     case obj
--- a/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
+++ b/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
@@ -1,4 +1,9 @@
 class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  BATCH_SIZE = 5_000
+
   module VimType
     def vimType
       @vimType.nil? ? nil : @vimType.to_s
@@ -17,17 +22,9 @@ class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
     end
   end
 
-  class VimHash < Hash
-    include VimType
-  end
-
-  class VimArray < Array
-    include VimType
-  end
-
-  class VimString < String
-    include VimType
-  end
+  class VimHash < Hash; include VimType; end
+  class VimArray < Array; include VimType; end
+  class VimString < String; include VimType; end
 
   class EventStream < ActiveRecord::Base
     include ActiveRecord::IdRegions
@@ -40,9 +37,18 @@ class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
     vim_string_backup = const_replace(:VimString, VimString)
     vim_array_backup  = const_replace(:VimArray, VimArray)
 
-    EventStream.in_my_region.where(:source => "VC").find_each do |event|
-      full_data = YAML.load(event.full_data)
-      event.update!(:full_data => vim_types_to_basic_types(full_data).to_yaml)
+    say_with_time("Removing Vim Types from EmsEvents") do
+      base_relation = EventStream.in_my_region.where(:source => "VC")
+      say_batch_started(base_relation.size)
+
+      processed_count = 0
+      base_relation.find_each(batch_size: BATCH_SIZE) do |event|
+        full_data = YAML.load(event.full_data)
+        event.update!(:full_data => vim_types_to_basic_types(full_data).to_yaml)
+
+        processed_count += 1
+        say_batch_processed(processed_count) if processed_count % BATCH_SIZE == 0
+      end
     end
 
     const_replace(:VimHash, vim_hash_backup)
--- a/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
+++ b/db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
@@ -35,7 +35,7 @@ class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
   def up
     with_replaced_constants(:VimHash => VimHash, :VimString => VimString, :VimArray => VimArray) do
       say_with_time("Removing Vim Types from EmsEvents") do
-        base_relation = EventStream.in_my_region.where(:source => "VC")
+        base_relation = EventStream.in_my_region.where(:source => "VC").where("full_data LIKE ?", "%hash-with-ivars:VimHash%")
         say_batch_started(base_relation.size)
 
         base_relation.find_in_batches(batch_size: BATCH_SIZE) do |events|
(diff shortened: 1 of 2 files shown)


spec/migrations/20200126203714_remove_vim_types_from_ems_events_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      FULL_DATA
+
+      event = event_stream_stub.create!(:full_data => full_data, :source => "VC")
+
+      migrate
+
+      full_data = YAML.load(event.reload.full_data)
+
+      expect(full_data.class.name).to eq("Hash")
+      expect(full_data["key"].class.name).to eq("String")
```
bdunne: Should the value also be verified here?

bdunne: Also, `expect(full_data["key"].class.name).to eq("String")` can be `expect(full_data["key"]).to be_kind_of(String)`

agrare: :+1: will do

agrare: > Also, expect(full_data["key"].class.name).to eq("String") can be expect(full_data["key"]).to be_kind_of(String)

The problem with that is a `VimString < String` so `expect(VimString).to be_kind_of(String)` passes

agrare: Alright pushed a change to check the values of key and datacenter

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
```diff
@@ -0,0 +1,86 @@
+class RemoveVimTypesFromEmsEvents < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  BATCH_SIZE = 5_000
+
+  module VimType
+    def vimType
+      @vimType.nil? ? nil : @vimType.to_s
```
Fryguy: Minor, but `@vimType&.to_s` will do the same (even though nil responds to .to_s, `&.` short circuits it anyway)

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
```diff
@@ -0,0 +1,86 @@
...
+  class EventStream < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    vim_hash_backup   = const_replace(:VimHash, VimHash)
+    vim_string_backup = const_replace(:VimString, VimString)
+    vim_array_backup  = const_replace(:VimArray, VimArray)
```
Fryguy: Would it be possible to use some yielding method here that `ensure`s that the cleanup occurs?

Picturing this like
```ruby
 def up
   with_replaced_consts(:VimHash => VimHash, :VimString => VimString, :VimArray => VimArray) do
     # stuff
   end
 end

File: db/migrate/20200126203714_remove_vim_types_from_ems_events.rb (hunk 2 above)
Comment (agrare) [reply db/migrate/20200126203714_remove_vim_types_from_ems_events.rb]: I copied exactly what was in `VMwareWebService/VimTypes` but I can change this if you want

File: db/migrate/20200126203714_remove_vim_types_from_ems_events.rb (hunk 3 above)
Comment (agrare) [reply db/migrate/20200126203714_remove_vim_types_from_ems_events.rb]: :+1: yeah definitely

File: db/migrate/20200126203714_remove_vim_types_from_ems_events.rb [hunk 4]
```diff
@@ -0,0 +1,86 @@
...
+
+  def vim_types_to_basic_types(obj)
+    case obj
+    when VimString
+      obj = obj.to_s
+    when VimHash
+      obj = obj.to_h
+      obj.each { |key, val| obj[key] = vim_types_to_basic_types(val) }
+    when VimArray
+      obj = obj.map { |v| vim_types_to_basic_types(v) }
```
Fryguy: `map!` might be a bit more efficient on memory here.

spec/migrations/20200126203714_remove_vim_types_from_ems_events_spec.rb
Fryguy: `.class == String` then?

agrare: Okay

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
Fryguy: Nah this is fine.

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
Fryguy: I thikn these backups are now no longer necessary

agrare: Yeah just forgot to delete them

Fryguy:
@agrare Do you have any performance estimates on how long this will take for a large table?

Fryguy:
I think we need a VimArray in the test too (I don't see one)

spec/migrations/20200126203714_remove_vim_types_from_ems_events_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect(full_data["datacenter"].class).to eq(Hash)
+      expect(full_data["datacenter"]).to include(
+        "name"       => "dev-vc67-DC",
+        "datacenter" => "datacenter-104"
+      )
+    end
+  end
+
+  migration_context :down do
+  end
```
Fryguy: I think you can remove this stub

agrare:
> Do you have any performance estimates on how long this will take for a large table?

I'm currently restoring a large lab DB I'll update here with how long the migration took when it completes

agrare:
> I think we need a VimArray in the test too (I don't see one)

Good point, I'll add a VimArray with a nested VimHash and VimString

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
agrare: Oh yeah, obj.map! still leave it as a VimArray

agrare:
```
>> EventStream.count
=> 892946
>> EmsEvent.count
=> 123341
>> EmsEvent.where(:source => "VC").count
=> 56283

== 20200126203714 RemoveVimTypesFromEmsEvents: migrating ======================
-- Removing Vim Types from EmsEvents
   -> Processing 56283 rows
   -> 5000 rows (8.88% - 5000 total - 23.37s - ETA: 4 minutes)
   -> 5000 rows (17.77% - 10000 total - 23.67s - ETA: 4 minutes)
   -> 5000 rows (26.65% - 15000 total - 24.41s - ETA: 3 minutes)
   -> 5000 rows (35.53% - 20000 total - 23.93s - ETA: 3 minutes)
   -> 5000 rows (44.42% - 25000 total - 24.93s - ETA: 3 minutes)
   -> 5000 rows (53.30% - 30000 total - 24.80s - ETA: 2 minutes)
   -> 5000 rows (62.19% - 35000 total - 23.89s - ETA: 2 minutes)
   -> 5000 rows (71.07% - 40000 total - 24.46s - ETA: 1 minute)
   -> 5000 rows (79.95% - 45000 total - 25.46s - ETA: less than a minute)
   -> 5000 rows (88.84% - 50000 total - 25.14s - ETA: half a minute)
   -> 5000 rows (97.72% - 55000 total - 25.55s - ETA: less than 10 seconds)
   -> 276.3252s
== 20200126203714 RemoveVimTypesFromEmsEvents: migrated (276.3253s) ===========
```

So it takes about 50s / 10,000 records

Fryguy:
I thikn the issue is that since transactions are disabled, you are essentially doing 1 commit per record instead of per batch.  Most other changes that do this usually use something like `.limit(batch_size).update_all`, which is like doing all the records in one transaction, but here you are using find_batch and going one by one.  I think you need to find a way to wrap each batch of N in a transaction.

Fryguy:
I also just noticed that this PR
a) might not be reentrant (which is the purpose of `disable_ddl_transaction!` and batching).  That is, I'm not sure if the conversion handles already converted records...seems like it might by nature of it falling through the case statement.
b) The query doesn't ignore already converted records.  This way you are not re-processing records if you have to re-enter.  I'm not sure there's a nice way to do this though, so we might have to skip it.

agrare:
> I'm not sure if the conversion handles already converted records...seems like it might by nature of it falling through the case statement.

Correct if it is a normal hash it won't do anything, I can make that more explicit and not update the record if full_data isn't a VimType

agrare:
> The query doesn't ignore already converted records. This way you are not re-processing records if you have to re-enter. I'm not sure there's a nice way to do this though, so we might have to skip it.

Let me see if I can get a LIKE condition added to the query that skips anything that doesn't have a VimType in it

Fryguy:
> Let me see if I can get a LIKE condition added to the query that skips anything that doesn't have a VimType in it

That might be expensive in the other direction...probably good for us to get a really large DB with real data, and see if adding it helps or hurts.

agrare:
Testing with the batched transactions now, then I'll run it again after it has been updated with no changes then again with the LIKE with no changes

agrare:
about half the time with transactions around the batch

```
== 20200126203714 RemoveVimTypesFromEmsEvents: migrating ======================
-- Removing Vim Types from EmsEvents
   -> Processing 56283 rows
   -> 5000 rows (8.88% - 5000 total - 8.12s - ETA: 1 minute)
   -> 5000 rows (17.77% - 10000 total - 12.96s - ETA: 2 minutes)
   -> 5000 rows (26.65% - 15000 total - 14.65s - ETA: 2 minutes)
   -> 5000 rows (35.53% - 20000 total - 15.01s - ETA: 2 minutes)
   -> 5000 rows (44.42% - 25000 total - 16.38s - ETA: 1 minute)
   -> 5000 rows (53.30% - 30000 total - 16.39s - ETA: 1 minute)
   -> 5000 rows (62.19% - 35000 total - 15.85s - ETA: 1 minute)
   -> 5000 rows (71.07% - 40000 total - 14.81s - ETA: less than a minute)
   -> 5000 rows (79.95% - 45000 total - 16.11s - ETA: half a minute)
   -> 5000 rows (88.84% - 50000 total - 16.07s - ETA: less than 20 seconds)
   -> 5000 rows (97.72% - 55000 total - 16.37s - ETA: less than 5 seconds)
   -> 5000 rows (106.60% - 60000 total - 4.28s - ETA: less than 20 seconds)
   -> 167.0932s
== 20200126203714 RemoveVimTypesFromEmsEvents: migrated (167.0933s) ===========
```

agrare:
Okay I created 1 million vmware events with VimHash full_data:

```
== 20200126203714 RemoveVimTypesFromEmsEvents: migrating ======================
-- Removing Vim Types from EmsEvents
   -> Processing 1000000 rows
.....
   -> 5000 rows (99.50% - 995000 total - 7.16s - ETA: less than 5 seconds)
   -> 5000 rows (100.00% - 1000000 total - 6.73s - ETA: less than 5 seconds)
   -> 639.0090s
== 20200126203714 RemoveVimTypesFromEmsEvents: migrated (639.0090s) ===========
```

So about 10mins for a million events

Fryguy:
As discussed, we found that replacing the VimHash with Hash (et.al.) seems to work, allowing us to do a pure SQL approach, avoiding the YAML altogether.

db/migrate/20200126203714_remove_vim_types_from_ems_events.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  class EventStream < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Removing Vim Types from EmsEvents") do
+      base_relation = EventStream.in_my_region.where(:source => "VC")
```
Fryguy: If the type column is indexed, we can probably add that into the where clause as well (might even be better than having source)

agrare: Actually neither of them do?
```
>> EventStream.connection.index_exists?(:event_streams, :type)
=> false
>> EventStream.connection.index_exists?(:event_streams, :source)
=> false
```

Fryguy: :(

agrare:
```
== 20200126203714 RemoveVimTypesFromEmsEvents: migrating ======================
-- Removing Vim Types from EmsEvents
   -> Processing 1000000 rows
   -> 50000 rows (5.00% - 50000 total - 8.69s - ETA: 3 minutes)
   -> 50000 rows (100.00% - 1000000 total - 9.14s - ETA: less than 5 seconds)
   -> 184.5047s
== 20200126203714 RemoveVimTypesFromEmsEvents: migrated (184.5048s) ===========
```

So 3 minutes for 1million rows now

PR #454, opened 2020-01-30
[WIP] Add and populate Tenant#ems_ref

lpichler:
We had issue when openstack provider was removed and then re-added with same name with openstack mapping(OM). OM creates tenant structure according to cloud tenants. When those tenants are tied with some group, then removal of openstack provider will leave tenants(tied with groups. ) in system (we cannot remove tenant with group)  and then OM fails when provider was re-added and provider was not able to create tenant structure.
In this case, we need to rejoin existing tenants with new OM but we don't have unique identification how to do it. We have just name(this name is reflected in mapped tenant) of provider  and it can be so we cannot rely on it.

to achieve this we need to add  nique identification  to the mapped tenant from provider.

in mapped tenants of CloudTenant we are using `CloudTenant#ems_ref` in 
in mapped root tenant of Provider we are using  `ExtManagement#guid`.

This PR adds ems_ref to Tenant and populates Tenant#ems_ref  according to ^

After this PR we can update OM sync process.

 @aufi can you review please.  ? can you take look whether I am using proper identification ( ExtManagement#guid., CloudTenant#ems_ref) for mapped tenants ? thanks! 

cc @gtanzillo 

@miq-bot add_label bug

### DIFF
--- a/db/migrate/20200129122439_copy_cloud_tenants_ems_ref_to_tenants.rb
+++ b/db/migrate/20200129122439_copy_cloud_tenants_ems_ref_to_tenants.rb
@@ -1,0 +1,36 @@
+class CopyCloudTenantsEmsRefToTenants < ActiveRecord::Migration[5.1]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Tenant < ActiveRecord::Base; end
+
+  class CloudTenant < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Copying CloudTenant#ems_ref or ExtManagementSystem#guid to Tenant#ems_ref") do
+      Tenant.where.not(:source_id => nil).each do |tenant|
+        case tenant.source_type
+        when "CloudTenant"
+          source_scope = CloudTenant.where(:id => tenant.source_id)
+          if source_scope.exists?
+            tenant.update_attributes(:ems_ref => source_scope.first.ems_ref)
+          end
+        when "ExtManagementSystem"
+          source_scope = ExtManagementSystem.where(:id => tenant.source_id)
+          if source_scope.exists?
+            tenant.update_attributes(:ems_ref => source_scope.first.guid)
+          end
+        end
+      end
+    end
+  end
+
+  def down
+    say_with_time("Nullifying Tenant#ems_ref") do
+      Tenant.where.not(:source_id => nil).where(:source_type => %w[CloudTenant ExtManagementSystem]).update_all(:ems_ref => nil)
+    end
+  end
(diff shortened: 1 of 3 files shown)


aufi:
@lpichler Hi Libor! ```CloudTenant#ems_ref``` is correct identification field for mapping :+1: 

The ```ExtManagement#guid``` doesn't come from OpenStack, it is most propably generated on CF side. If the provider would be deleted and re-added, it will most propably be a different one.

Only thing I can think of is ```keystone_domain_id``` which is stored as ```uid_ems``` attribute on ExtManagementSystem OpenStack subclasses. It is set only when Keystone v3 is used, but that should be required for Tenant Mapping feature. Actually the Domain object on OpenStack side is a parent object of Projects, so it might make sense implement it in this way.

db/migrate/20200129122439_copy_cloud_tenants_ems_ref_to_tenants.rb
```diff
@@ -0,0 +1,36 @@
...
+  class CloudTenant < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Copying CloudTenant#ems_ref or ExtManagementSystem#guid to Tenant#ems_ref") do
+      Tenant.where.not(:source_id => nil).each do |tenant|
+        case tenant.source_type
+        when "CloudTenant"
+          source_scope = CloudTenant.where(:id => tenant.source_id)
```
agrare: Is it possible for there to be more than one CloudTenant record with the same tenant.source_id?  This looks like it should just be `CloudTenant.find_by`

db/migrate/20200129122439_copy_cloud_tenants_ems_ref_to_tenants.rb
```diff
@@ -0,0 +1,36 @@
...
+          if source_scope.exists?
+            tenant.update_attributes(:ems_ref => source_scope.first.ems_ref)
+          end
+        when "ExtManagementSystem"
+          source_scope = ExtManagementSystem.where(:id => tenant.source_id)
+          if source_scope.exists?
+            tenant.update_attributes(:ems_ref => source_scope.first.guid)
```
agrare: I don't think guid is going to help you here, if you destroy and re-create an ExtManagementSystem you'll get a new guid

Fryguy:
Closing as this needs more design...

I feel like adding single ems_ref to a shared platform object is going to end up in collisions.  For example, if 2 openstack providers both have a tenant with the same name, which one will the ems_ref in Tenant map to?  Arguably, it should map to both.  In addition, when tenant mapping is added to other providers, what will happen there.   I feel like this needs a matrix of expectations laid out for these different use cases, before we start coding in new columns.

PR #455, opened 2020-02-10
MiqAeNamespace should use ancestry instead of acts_as_tree

lfu:
Cross Repo Tests: ManageIQ/manageiq-cross_repo-tests#66
Blocks https://github.com/ManageIQ/manageiq/pull/19824

https://github.com/ManageIQ/manageiq-automation_engine/issues/409

@miq-bot add_label performance

cc @tinaafitz

### DIFF
--- a/db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
+++ b/db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
@@ -1,0 +1,31 @@
+require 'ancestry'
+
+class AddAncestryToMiqAeNamespace < ActiveRecord::Migration[5.1]
+  class MiqAeNamespace < ActiveRecord::Base
+    has_ancestry
+  end
+
+  def up
+    add_column :miq_ae_namespaces, :ancestry, :string
+    add_index :miq_ae_namespaces, :ancestry
+
+    say_with_time("Converting MiqAeNamespaces from parent_id to ancestry") do
+      MiqAeNamespace.build_ancestry_from_parent_ids!
+    end
+
+    remove_index :miq_ae_namespaces, :column => :parent_id
+    remove_column :miq_ae_namespaces, :parent_id
+  end
+
+  def down
+    add_column :miq_ae_namespaces, :parent_id, :bigint
+    add_index :miq_ae_namespaces, :parent_id
+
+    say_with_time("Converting MiqAeNamespaces from ancestry to parent_id") do
+      MiqAeNamespace.update_all("parent_id = CAST(regexp_replace(ancestry, '.*/', '') AS bigint)")
+    end
+
+    remove_index :miq_ae_namespaces, :column => :ancestry
+    remove_column :miq_ae_namespaces, :ancestry
+  end
+end
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{db,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
 
+  s.add_dependency "ancestry"
   s.add_dependency "activerecord-id_regions", "~> 0.3.0"
   s.add_dependency "linux_admin", "~> 2.0"
   s.add_dependency "manageiq-password", "~> 0.3"
--- a/spec/migrations/20200221200855_add_ancestry_to_miq_ae_namespace_spec.rb
+++ b/spec/migrations/20200221200855_add_ancestry_to_miq_ae_namespace_spec.rb
@@ -1,0 +1,54 @@
+require_migration
+
+describe AddAncestryToMiqAeNamespace do
+  let(:ns_stub) { Class.new(ActiveRecord::Base) { self.table_name = "miq_ae_namespaces" } }
+
+  migration_context :up do
+    # nodes:
+    # ns1
+    #   ns11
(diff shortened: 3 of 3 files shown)


db/migrate/20200203200855_add_ancestry_to_miq_ae_namespace.rb
```diff
@@ -0,0 +1,40 @@
+class AddAncestryToMiqAeNamespace < ActiveRecord::Migration[5.0]
+  class MiqAeNamespace < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def update_ns_parent(parent)
```
Fryguy: Lucty, as discussed you should be able to use the built-in `build_ancestry_from_parent_ids!` method.

db/migrate/20200203200855_add_ancestry_to_miq_ae_namespace.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def down
+    add_column :miq_ae_namespaces, :parent_id, :bigint
+    add_index :miq_ae_namespaces, :parent_id
+
+    say_with_time("Converting MiqAeNamespaces from ancestry to parent_id") do
+      MiqAeNamespace.all.each do |ns|
+        parent_ns_id = ns.ancestry.split("/").last.to_i if ns.ancestry.present?
+        ns.update(:parent_id => parent_ns_id)
+      end
```
Fryguy: I think this can be done in a single query... looking for my original code change.

db/migrate/20200203200855_add_ancestry_to_miq_ae_namespace.rb
```diff
@@ -0,0 +1,31 @@
+require 'ancestry'
+
+class AddAncestryToMiqAeNamespace < ActiveRecord::Migration[5.0]
```
jrafanie: If this is for master, we should be generating the migration using rails 5.1.  I think the 5.0 here would still work but going forward, we'll be doing 5.1, hopefully 5.2 soon.

manageiq-schema.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{db,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
 
+  s.add_dependency "ancestry"
```
jrafanie: @Fryguy what do you think of this dependency in schema?  I believe requiring ancestry in schema migration was a thing that was possibly too much work to try to avoid.

kbrock: ancestry buys us `build_ancestry_from_parent_id`

db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
```diff
@@ -0,0 +1,31 @@
...
+  class MiqAeNamespace < ActiveRecord::Base
+    has_ancestry
+  end
+
+  def up
+    add_column :miq_ae_namespaces, :ancestry, :string
+    add_index :miq_ae_namespaces, :ancestry
+
+    say_with_time("Converting MiqAeNamespaces from parent_id to ancestry") do
+      MiqAeNamespace.build_ancestry_from_parent_ids!
```
bdunne, line 13: Is there any way to scope this to only the current region?

db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
```diff
@@ -0,0 +1,31 @@
...
+    remove_index :miq_ae_namespaces, :column => :parent_id
+    remove_column :miq_ae_namespaces, :parent_id
+  end
+
+  def down
+    add_column :miq_ae_namespaces, :parent_id, :bigint
+    add_index :miq_ae_namespaces, :parent_id
+
+    say_with_time("Converting MiqAeNamespaces from ancestry to parent_id") do
+      MiqAeNamespace.update_all("parent_id = CAST(regexp_replace(ancestry, '.*/', '') AS bigint)")
```
bdunne, line 25: Same here

db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
bdunne: nevermind, it's excluded.  carry on

manageiq-schema.gemspec
Fryguy: I'm ok with this being in schema...if it changes later, we'll have to deal with it.

otherwise we have to copy the guts of `build_ancestry_from_parent_id`

db/migrate/20200221200855_add_ancestry_to_miq_ae_namespace.rb
Fryguy: I believe that method is scope aware, so it would be worth it to try.

simaishi:
Jansa backport details:

```
$ git log -1
commit 7051f7cbfd163947468a6e2238125f4c54784c2e
Date:   Thu Apr 2 13:29:23 2020 -0400

    Merge pull request #455 from lfu/add_ancestry_to_miq_ae_namespace

    MiqAeNamespace should use ancestry instead of acts_as_tree

    (cherry picked from commit 1cab5a114a1b8ae1bbeb0231a498383c95f7bedf)
```

PR #458, opened 2020-02-12
Convert OvirtSDK4 objects to Hash in EventStream full_data

agrare:
The `Ovirt` event catcher was populating the `event_streams` `:full_data` column with `OvirtSDK4` objects like `OvirtSDK4::Event`.  This was causing the event handler to throw an exception because it does not load the `OvirtSDK4` gem.

Similar to: https://github.com/ManageIQ/manageiq-schema/pull/451

### DIFF
--- a/db/migrate/20200212171537_remove_ovirt_sdk4_from_event_stream.rb
+++ b/db/migrate/20200212171537_remove_ovirt_sdk4_from_event_stream.rb
@@ -1,0 +1,27 @@
+class RemoveOvirtSdk4FromEventStream < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  class EventStream < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Removing OvirtSDK4 types from EmsEvents") do
+      base_relation = EventStream.in_my_region.where(:source => "RHEVM")
+        .where("full_data LIKE ?", "%object:OvirtSDK4::Event%")
+
+      say_batch_started(base_relation.size)
+
+      loop do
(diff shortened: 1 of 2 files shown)


Fryguy:
Are these objects originally a subclass of Hash or OpenStruct?

db/migrate/20200212171537_remove_ovirt_sdk4_from_event_stream.rb
```diff
@@ -0,0 +1,27 @@
...
+  def up
+    say_with_time("Removing OvirtSDK4 types from EmsEvents") do
+      base_relation = EventStream.in_my_region.where(:source => "RHEVM")
+        .where("full_data LIKE ?", "%object:OvirtSDK4::Event%")
+
+      say_batch_started(base_relation.size)
+
+      loop do
+        count = base_relation.limit(50_000)
+          .update_all("full_data = REGEXP_REPLACE(full_data, '!ruby/object:OvirtSDK4::\\w+', '!ruby/object:Hash', 'g')")
```
Fryguy: Are the objects always directly under the OvirtSDK namespace?  That is, could there ever be more than one set of `::`

agrare:
> Are these objects originally a subclass of Hash or OpenStruct?

Looks like they're Structs, https://github.com/oVirt/ovirt-engine-sdk-ruby/blob/master/sdk/lib/ovirtsdk4/type.rb#L95-L126

db/migrate/20200212171537_remove_ovirt_sdk4_from_event_stream.rb
agrare: Doesn't look like it:
```
>> OvirtSDK4::Identified.descendants.map(&:name).any? { |n| n.scan("::").count > 1 }
=> false
```

(`OvirtSDK4::Identified` is the base class that all others are derived from)

PR #459, opened 2020-02-18
Add "thin" and "format" columns to disk

borod108:
Required for: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/453
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1726590

### DIFF
--- a/db/migrate/20200219125312_add_thin_and_format_to_disk.rb
+++ b/db/migrate/20200219125312_add_thin_and_format_to_disk.rb
@@ -1,0 +1,6 @@
+class AddThinAndFormatToDisk < ActiveRecord::Migration[5.1]
+  def change
+    add_column :disks, :thin, :boolean
+    add_column :disks, :format, :string
+  end
+end


borod108:
@agrare is this ok?
I know now I removed all other providers it is over engineered, but if someone would like to use this as a bases to do the other ones - it is there.

agrare:
Hey @borod108 , if you just add the `:thin` and `:format` columns I'll take care of the data migration for all of the providers for you.  This should simplify it down to a simple `#change` with the `add_column`s

borod108:
@agrare done.

db/migrate/20200219125312_add_thin_and_format_to_disk.rb
```diff
@@ -0,0 +1,6 @@
+class AddThinAndFormatToDisk < ActiveRecord::Migration[5.1]
+  def change
+    add_column :disk, :thin, :boolean
```
agrare: Ah this needs to be `:disks` not `:disk`

borod108: done

PR #464, opened 2020-04-03
Renamed foreman features to ems_configuration and adjusted roles

skateman:
Adjusting the user roles to the renamed foreman features. This isn't the final form yet as the name of the features might change in the core PR. 

Core PR: https://github.com/ManageIQ/manageiq/pull/19949
UI PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/6782

### DIFF
--- a/db/migrate/20200331150436_rename_foreman_features.rb
+++ b/db/migrate/20200331150436_rename_foreman_features.rb
@@ -1,0 +1,85 @@
+class RenameForemanFeatures < ActiveRecord::Migration[5.1]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+  class User < ActiveRecord::Base
+    serialize :settings, Hash
+  end
+
+  FEATURE_MAPPING = {
+    'providers_accord'                   => 'ems_configuration',
+    'provider_foreman_view'              => 'ems_configuration_view',
+    'provider_foreman_control'           => 'ems_configuration_control',
+    'configuration_manager_provider_tag' => 'ems_configuration_tag',
+    'provider_admin'                     => 'ems_configuration_admin',
+    'provider_foreman_delete_provider'   => 'ems_configuration_delete_provider',
+    'provider_foreman_edit_provider'     => 'ems_configuration_edit_provider',
+    'provider_foreman_add_provider'      => 'ems_configuration_add_provider',
+    'configured_systems_filter_accord'   => 'configured_system',
+  }.freeze
+
+  ROLES_FEATURE_MAPPING = {
+    'provider_foreman_explorer'             => %w[ems_configuration configured_system configuration_profile],
+    'configured_systems_filter_accord'      => %w[configured_system],
+    'configured_systems_filter_accord_view' => %w[configured_system_view],
+    'configuredsystem_control'              => %w[configured_system_control],
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    say_with_time 'Renaming old foreman explorer features to the non-explorer ones' do
+      # Direct renaming of features
+      FEATURE_MAPPING.each do |from, to|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+
+      # Mapping to already existing features
+      ROLES_FEATURE_MAPPING.each do |from, to|
+        from_feature = MiqProductFeature.find_or_create_by!(:identifier => from)
+        to_features = to.map { |identifier| MiqProductFeature.find_or_create_by!(:identifier => identifier) }
+
+        MiqRolesFeature.where(:miq_product_feature_id => from_feature.id).each do |roles_feature|
+          to_features.each do |to_feature|
+            MiqRolesFeature.create!(:miq_product_feature_id => to_feature.id, :miq_user_role_id => roles_feature.miq_user_role_id)
+          end
+        end
+      end
+    end
+
+    say_with_time 'Updating starting page for users who had foreman explorer set' do
+      # Update starting page for users who had it on provider_foreman/explorer
+      User.select(:id, :settings).each do |user|
+        if user.settings&.dig(:display, :startpage) == 'provider_foreman/explorer'
+          user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'ems_configuration/show_list'}))
+        end
+      end
+    end
+  end
+
+  def down
+    return if MiqProductFeature.none?
+
+    say_with_time 'Renaming new ems configuration features to old foreman explorer ones' do
+      FEATURE_MAPPING.each do |to, from|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+
+      ROLES_FEATURE_MAPPING.each do |to, from|
+        next if from.length > 1 # The explorer one is redundant if all its 3 children are checked
+
+        from_feature = MiqProductFeature.find_or_create_by!(:identifier => from.first)
+        to_feature = MiqProductFeature.find_or_create_by!(:identifier => to.first)
+
+        MiqRolesFeature.where(:miq_product_feature_id => from_feature.id)&.update(:miq_product_feature_id => to_feature.id)
+      end
+    end
+
+    say_with_time 'Updating starting page for users who had non-explorer ems_configuration pages set' do
+      User.select(:id, :settings).each do |user|
+        if %w[ems_configuration/show_list configuration_profile/show_list configured_system/show_list].include?(user.settings&.dig(:display, :startpage))
+          user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'provider_foreman/explorer'}))
+        end
+      end
+    end
+  end
+end
--- a/spec/migrations/20200331150436_rename_foreman_features_spec.rb
+++ b/spec/migrations/20200331150436_rename_foreman_features_spec.rb
@@ -1,0 +1,92 @@
+require_migration
+
+describe RenameForemanFeatures do
+  let(:user_role_id) { anonymous_class_with_id_regions.id_in_region(1, anonymous_class_with_id_regions.my_region_number) }
+  let(:feature_stub) { migration_stub :MiqProductFeature }
+  let(:roles_feature_stub) { migration_stub :MiqRolesFeature }
+  let(:user_stub) { migration_stub :User }
+
+  migration_context :up do
+    describe 'product features update' do
+      it 'renames the features' do
+        accord_feature = feature_stub.create!(:identifier => 'providers_accord')
+        view_feature = feature_stub.create!(:identifier => 'provider_foreman_view')
+
+        migrate
+
+        expect(accord_feature.reload.identifier).to eq('ems_configuration')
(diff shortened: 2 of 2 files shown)


skateman:
@Fryguy this is how far I could get, I am totally unsure why that one case is failing :confused:

Fryguy:
@bdunne Can you also lend some eyes to this?

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,56 @@
+class RenameForemanFeatures < ActiveRecord::Migration[5.1]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+
+  def up
+    MiqProductFeature.where(:identifier => 'provider_foreman_explorer').update(:identifier => 'ems_configuration')
+
+    MiqProductFeature.where(:identifier => 'provider_foreman_view').update(
+      :name        => 'View',
+      :description => 'View Configuration Manager Providers',
```
Fryguy, line 10: For simplicity I think you can leave out the name and description from the migration, as the seed with fill out the rest of the details (unless of course you need it because of other uniqueness contraints).  The identifier is really the only thing that matters from a seeding perspective.

Fryguy:
The reason the specs are failing is because data migrations should leave a previously empty database empty afterwards.  This is the base case for a brand new installation, where the database is completely empty and when you migrate up it should still be empty.  This is important for things like the test database.

The way to ensure this is to see if there are existing records that need to be migrated before migrating anything.  If there are no records, there's nothing to "fix", so you can skip the rest.

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,56 @@
+class RenameForemanFeatures < ActiveRecord::Migration[5.1]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+
+  def up
+    MiqProductFeature.where(:identifier => 'provider_foreman_explorer').update(:identifier => 'ems_configuration')
```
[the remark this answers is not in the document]
Fryguy: There should only be one...Perhaps `find_by` would be better than `where` to make that clear?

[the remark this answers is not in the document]
skateman: Sure, I just thought that `update` doesn't work on a `nil` but `where` worst case returns an empty array.

skateman:
@Fryguy I understand what's wrong with the specs, but I don't really understand what should I do to fix them.

Fryguy:
@skateman Oh, just add this as a first line in both the `up` and the `down`

`return if MiqProductFeature.none?`

skateman:
@Fryguy ok, now it makes sense what you wrote before, thanks.

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,35 @@
+class RenameForemanFeatures < ActiveRecord::Migration[5.1]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+
+  def up
+    return if MiqProductFeature.none?
+
+    MiqProductFeature.find_by(:identifier => 'provider_foreman_explorer')&.update(:identifier => 'ems_configuration')
```
Fryguy, line 8: I think we should do `update!` as opposed to `update` here and elsewhere in the file.

spec/migrations/20200331150436_rename_foreman_features_spec.rb
```diff
@@ -0,0 +1,58 @@
+require_migration
+
+describe RenameForemanFeatures do
+  class MiqRolesFeature < ActiveRecord::Base; end
+
+  let(:user_role_id) { rand(1..1000) }
```
Fryguy, line 6: To deal with regions accurately, prefer something like

`ActiveRecord::IdRegions::Migration.anonymous_class_with_id_regions.rails_sequence_start`

Fryguy: If you really want a random id just add `+ rand(1...1000)` to the end of that

Fryguy: @bdunne We could really use a cleaner helper method for stuff like this... maybe a spec/support thing in activerecord-id_regions gem?  I'm thinking a handful of helpers for the most common stuff.

bdunne: Yeah, that would make it much easier to use

spec/migrations/20200331150436_rename_foreman_features_spec.rb
```diff
@@ -0,0 +1,58 @@
+require_migration
+
+describe RenameForemanFeatures do
+  class MiqRolesFeature < ActiveRecord::Base; end
```
Fryguy, line 4: This should not be needed. as you have the roles_feature_stub below, which is preferable.

Fryguy:
Found one last thing, otherwise LGTM

skateman:
@h-kataria found some issues, moving this back to WIP

skateman:
@Fryguy we had to adjust the migration a little bit more after revisiting the core PR, can you please re-review?

h-kataria:
tested, LGTM.

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,77 @@
...
+    'configured_systems_filter_accord_view' => %w[configured_system_view],
+    'configuredsystem_control'              => %w[configured_system_control],
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    # Direct renaming of features
+    FEATURE_MAPPING.each do |from, to|
+      MiqProductFeature.find_by(:identifier => from)&.update(:identifier => to)
```
Fryguy, line 32: Please use `update!`

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,77 @@
...
+    'configured_systems_filter_accord'      => %w[configured_system],
+    'configured_systems_filter_accord_view' => %w[configured_system_view],
+    'configuredsystem_control'              => %w[configured_system_control],
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    # Direct renaming of features
+    FEATURE_MAPPING.each do |from, to|
```
Fryguy, line 31: For data migrations, we also tend to use `say_with_time` to give feedback to the user about how long it takes.

Not sure of how granular each block should be,  It could be one giant block around the whole thing, but it might be better to have 3 sections  since you've broken up the code that way as well.

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+      end
+    end
+  end
+
+  def down
+    return if MiqProductFeature.none?
+
+    say_with_time 'Renaming new ems configuration features to old foreman explorer ones' do
+      FEATURE_MAPPING.each do |to, from|
```
Fryguy, line 63: There a handful of query-in-a-loop patterns here that can be really inefficient, but we're not talking a ton of identifiers here, so maybe its not really an issue, and the code is pretty elegant to read as is.  I'll leave it up to you.  I think the alternative would look something like...

```ruby
features = MiqProductFeature.all.index_by(&:identifier)
...
FEATURE_MAPPING.each do |from, to|
  features[from]&.update!(...)
end
...
ROLES_FEATURE_MAPPING.each do |from, to|
  from_feature = (features[from] ||= MiqProductFeature.create!(...)
  to_features = to.map { |identifier| (features[to] ||= MiqProductFeature.create!(...)  }
...
```

db/migrate/20200331150436_rename_foreman_features.rb
```diff
@@ -0,0 +1,85 @@
...
+        to_feature = MiqProductFeature.find_or_create_by!(:identifier => to.first)
+
+        MiqRolesFeature.where(:miq_product_feature_id => from_feature.id)&.update(:miq_product_feature_id => to_feature.id)
+      end
+    end
+
+    say_with_time 'Updating starting page for users who had non-explorer ems_configuration pages set' do
+      User.select(:id, :settings).each do |user|
+        if %w[ems_configuration/show_list configuration_profile/show_list configured_system/show_list].include?(user.settings&.dig(:display, :startpage))
+          user.update(:settings => user.settings.deep_merge(:display => {:startpage => 'provider_foreman/explorer'}))
```
Fryguy, line 80: Prefer `update!` ;)

spec/migrations/20200331150436_rename_foreman_features_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(roles_feature_stub.where(:miq_user_role_id => user_role_id).count).to eq(1)
+
+      migrate
+
+      expect(roles_feature_stub.where(:miq_user_role_id => user_role_id).count).to eq(4)
+    end
+
+    it 'replaces user starting page if foreman' do
+      feature_stub.create!(:identifier => 'provider_foreman_explorer')
+      user = user_stub.create!(:settings => {:display => {:startpage => 'provider_foreman/explorer'}})
```
Fryguy, line 37: Does there need to be a test for a user that doesn't have the settings at all, or perhaps doesn't have the :settings => :display section?  Is that even a possibility?

Fryguy:
This PR is awesome...just a couple more minor comments.

Fryguy:
Also seems the Travis issues are legit.

spec/migrations/20200331150436_rename_foreman_features_spec.rb
h-kataria: User always have default settings https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/application_controller.rb#L100, default (startpage setting)[https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/application_controller.rb#L193]

Fryguy: @h-kataria Those are defined as defaults in the controller and so are not available at migration time.  My question is whether those values are always present in the column.  That is, can the column value be nil or an empty hash, thus causing this code to blow up?

skateman: @Fryguy I am doing the check for the settings in the safest way possible:
```ruby
if user.settings&.dig(:display, :startpage) == 'provider_foreman/explorer'
```

db/migrate/20200331150436_rename_foreman_features.rb
skateman: I'd like to leave it as it is, for readability. We don't really have millions of features.

simaishi:
Jansa backport details:

```
$ git log -1
commit c8a9c45aa717255e43022b7290011b47194c4986
Date:   Thu May 7 22:05:16 2020 -0400

    Merge pull request #464 from skateman/rename-foreman-features

    Renamed foreman features to ems_configuration and adjusted roles

    (cherry picked from commit e3e59070665932b94e0b2a5bbd0bd28f7713b346)
```

PR #468, opened 2020-04-24
add system flag to dialogs

d-m-u:
we have one built-in dialog, we could use something like this to differentiate it
really i just want to see what people have to say about it

requires data migration for it: 
`Dialog.find_by(:label => 'Transform VM')`

that issue mentions the azure one, `manageiq-providers-azure/content/service_dialogs/azure-single-vm-from-user-image.yml` but I think we got rid of that dialog in https://github.com/ManageIQ/manageiq-providers-azure/pull/107

see https://github.com/ManageIQ/manageiq/issues/15144

### DIFF
--- a/db/migrate/20200424183342_add_read_only_to_dialogs.rb
+++ b/db/migrate/20200424183342_add_read_only_to_dialogs.rb
@@ -1,0 +1,5 @@
+class AddReadOnlyToDialogs < ActiveRecord::Migration[5.1]
+  def change
+    add_column :dialogs, :system, :boolean
+  end
+end


d-m-u:
@himdel could I please get you to look at this?

himdel:
I guess it makes sense :+1: 

Though, on the other hand, I don't see why we would stop the user from changing that dialog if they really want to.

And on the third hand, I was under the impression that "vm_transform" is the operation that should have been replaced by the whole v2v repo.

himdel:
Oh, reading back the original issue... the motivation was that some dialogs are Magic now, and the UI should know about them being special.

So, maybe we should go with something less specific, like `builtin` instead of `readonly`. Read only can be the effect, or we can hide it completely, I don't know what we should do with that one dialog (if it can't be removed in favor of v2v).

But if we call it readonly, sooner or later there will be a way to mark dialogs read only and we lose the info again.

d-m-u:
` the motivation was that some dialogs are Magic now`
🤣

db/migrate/20200424183342_add_read_only_to_dialogs.rb
```diff
@@ -0,0 +1,5 @@
+class AddReadOnlyToDialogs < ActiveRecord::Migration[5.1]
+  def change
+    add_column :dialogs, :read_only, :boolean
```
bdunne, line 3: Just a warning that Rails has a concept of `readonly`, this may cause some confusion.  In some other places I think we've used `system`.

d-m-u:
I guess it feels like this could potentially be useful functionality to have even if that one dialog that this is going to to effectively change isn't being used.

bdunne:
Wouldn't they have to copy and edit?  It looks like seeding would overwrite attributes the next time it is run.

bdunne:
cc @Fryguy for naming

Fryguy:
I didn't look at the issue, but I've personally always liked the column name `system` to imply an ootb, un-editable thing that is added/updated/deleted via seeding.

d-m-u:
soooo this is like super low prio and i wouldn't even look at it if i were you all cause it doesn't really matter

d-m-u:
oh, cool, okay, well, thanks @bdunne

PR #471, opened 2020-04-27
set dialog system flag

d-m-u:
we have one built-in dialog, we could use something like this to differentiate them
really i just want to see what people have to say about it

requires data migration for it: 
`Dialog.find_by(:label => 'Transform VM')`

that issue mentions the azure one, `manageiq-providers-azure/content/service_dialogs/azure-single-vm-from-user-image.yml` but I think we got rid of that dialog in https://github.com/ManageIQ/manageiq-providers-azure/pull/107

see https://github.com/ManageIQ/manageiq/issues/15144

where.not doesn't work with nils but we've the name/label can't be nil anyway

## depends on 
data migration required for https://github.com/ManageIQ/manageiq-schema/pull/468

### DIFF
--- a/db/migrate/20200427175819_set_read_only_flag.rb
+++ b/db/migrate/20200427175819_set_read_only_flag.rb
@@ -1,0 +1,13 @@
+class SetReadOnlyFlag < ActiveRecord::Migration[5.1]
+  class Dialog < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time("Updating Dialogs") do
+      Dialog.in_my_region.where(:label => 'Transform VM').update_all(:system => true)
+      Dialog.in_my_region.where.not(:label => 'Transform VM').update_all(:system => false)
+    end
+  end
+end
--- a/spec/migrations/20200427175819_set_read_only_flag_spec.rb
+++ b/spec/migrations/20200427175819_set_read_only_flag_spec.rb
@@ -1,0 +1,17 @@
+require_migration
+
+describe SetReadOnlyFlag do
+  let(:dialog) { migration_stub(:Dialog) }
+
+  migration_context :up do
+    it "sets the flag" do
+      dialog1 = dialog.create(:label => "anything else")
+      dialog2 = dialog.create(:label => 'Transform VM')
+
+      migrate
+
+      expect(dialog1.reload.system).to be(false)
+      expect(dialog2.reload.system).to be(true)
+    end
+  end
+end


db/migrate/20200427175819_set_read_only_flag.rb
```diff
@@ -0,0 +1,12 @@
+class SetReadOnlyFlag < ActiveRecord::Migration[5.1]
+  class Dialog < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time("Updating Dialogs") do
+      Dialog.in_my_region.all.each { |dialog| dialog.label == 'Transform VM' ? dialog.update(:system => true) : dialog.update(:system => false) }
```
bdunne, line 9: What do you think of this?  It prevents loading every record and enumerating through them one-by-one.
```suggestion
      Dialog.in_my_region.where(:label => 'Transform VM').update_all(:system => true)
      Dialog.in_my_region.where.not(:label => 'Transform VM').update_all(:system => false)
```

d-m-u: 👍 thanks

bdunne:
Can you also add a `default_value_for` and update the seeding logic in the model?

bdunne:
@d-m-u Can you take care of the default value and take a look at the test failure?

d-m-u:
@bdunne I had asked you about the comment about "update the seeding logic in the model" and still am not sure what is required for that part.

bdunne:
> [ManageIQ/manageiq#20198](https://github.com/ManageIQ/manageiq/pull/20198) is already merged. did i miss something there?

Oh, no I missed that one :+1: 

> @bdunne I had asked you about the comment about "update the seeding logic in the model" and still am not sure what is required for that part.

It looks like the dialogs loaded from `file` https://github.com/ManageIQ/manageiq/blob/f124ea91dde7a7e4fec3e752e3e8e4f9cc47950a/app/models/dialog.rb#L27 need a `"system" => true` merged in before it's run through https://github.com/ManageIQ/manageiq/blob/f124ea91dde7a7e4fec3e752e3e8e4f9cc47950a/lib/services/dialog_import_service.rb#L145

d-m-u:
@bdunne okay but I was under the impression that the dialogs in question were in the fixtures dir, and there's a noticeable absence of dialogs there. Can you please clarify?

bdunne:
> @bdunne okay but I was under the impression that the dialogs in question were in the fixtures dir, and there's a noticeable absence of dialogs there. Can you please clarify?

No, they come from `content/service_dialogs` in the plugins. https://github.com/ManageIQ/manageiq/blob/f124ea91dde7a7e4fec3e752e3e8e4f9cc47950a/app/models/dialog.rb#L26

PR #452, opened 2020-01-28
Add type column to service_order table

gmcculloug:
NEEDS TO BE MERGED ALONG WITH https://github.com/ManageIQ/manageiq/pull/19766

This adds a type column to the service_order table. The associated type would be plucked from its associated requests e.g. ServiceTemplateProvisionRequest or ServiceTemplateTransformationPlanRequest.

**Still need to update ServiceOrder with type on creation so new `ServiceOrder` instances use STI.**
Needs to be merged along with  https://github.com/ManageIQ/manageiq/pull/19795

Attempts to solve ManageIQ/manageiq-ui-service#1463

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1565049

PR #475, opened 2020-05-01
Move database.yml to database.tmpl.yml and copy it for CI

bdunne:
- Allow people to maintain their own database.yml
- Copy database.yml from template if it doesn't exist

[I think I told GM that I would do this](https://github.com/ManageIQ/manageiq-schema/pull/452#issuecomment-622374578)


agrare:
:+1:

bdunne:
> I think the database.yml needs to be added to .gitignore. also should we update the readme since the instructions don't quite match?

All of `spec/dummy/` is already in the `.gitignore`.  README updated.

README.md
```diff
@@ -24,9 +24,10 @@ The gem is available as open source under the terms of the [Apache License 2.0](
 5. Create new Pull Request
 
 ## Run Tests in dev environment
-1. Modify `spec/dummy/config/database.yml` to access local database
+1. If necessary, copy `spec/dummy/config/database.tmpl.yml` to `spec/dummy/config/database.yml` and modify it to access your local database
 2. Run: `bin/setup`
-    1. Performs `bundle update`
-    2. Generates random database region number
-    3. Creates/resets `dummy_test` database
+    1. Copies `spec/dummy/config/database.tmpl.yml` to `spec/dummy/config/database.yml` it it doesn't exist
```
chessbyte, line 29: typo: `it it`

bdunne: Fixed, thanks!

PR #477, opened 2020-05-08
Remove most of VmdbDatabase and friends

jrafanie:
Fixes https://github.com/ManageIQ/manageiq/issues/20103 (backend issue)
Related to https://github.com/ManageIQ/manageiq-ui-classic/issues/6995 (frontend
issue)
- [x] Goes together with the frontend PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/6998 [MERGED]
- [ ] Goes together with the core PR: https://github.com/ManageIQ/manageiq/pull/20148
- [x] [cross repo tests](https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/119)

We're leaving VmdbDatabaseConnection and VmdbDatabaseLock models because there is
application code using the connection information, etc., so we'll need
replacements for those before we can remove them.

### DIFF
--- a/db/migrate/20200508150740_remove_vmdb_database_tables.rb
+++ b/db/migrate/20200508150740_remove_vmdb_database_tables.rb
@@ -1,0 +1,78 @@
+class RemoveVmdbDatabaseTables < ActiveRecord::Migration[5.2]
+  def up
+    drop_table :vmdb_database_metrics
+    drop_table :vmdb_databases
+    drop_table :vmdb_indexes
+    drop_table :vmdb_metrics
+    drop_table :vmdb_tables
+  end
+
+  def down
+    create_table "vmdb_database_metrics", force: :cascade do |t|
+      t.bigint "vmdb_database_id"
+      t.integer "running_processes"
+      t.integer "active_connections"
+      t.datetime "timestamp"
+      t.string "capture_interval_name"
+      t.bigint "disk_total_bytes"
+      t.bigint "disk_free_bytes"
+      t.bigint "disk_used_bytes"
+      t.bigint "disk_total_inodes"
+      t.bigint "disk_used_inodes"
+      t.bigint "disk_free_inodes"
+    end
+
+    create_table "vmdb_databases", force: :cascade do |t|
+      t.string "name"
+      t.string "ipaddress"
+      t.string "vendor"
+      t.string "version"
+      t.string "data_directory"
+      t.datetime "last_start_time"
+      t.string "data_disk"
+    end
+
+    create_table "vmdb_indexes", force: :cascade do |t|
+      t.bigint "vmdb_table_id"
+      t.string "name"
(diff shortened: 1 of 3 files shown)


carbonin:
> We're leaving VmdbDatabaseConnection and VmdbDatabaseLock

Those are not actual tables in our schema though, right? Pretty sure those are backed by PG system tables/views.

db/migrate/20200508150740_remove_vmdb_database_tables.rb
```diff
@@ -0,0 +1,78 @@
...
+  def up
+    drop_table :vmdb_database_metrics
+    drop_table :vmdb_databases
+    drop_table :vmdb_indexes
+    drop_table :vmdb_metrics
+    drop_table :vmdb_tables
+  end
+
+  def down
+    create_table "vmdb_database_metrics", force: :cascade do |t|
```
jrafanie, line 11: I grabbed the following from my schem.rb for the tables that were deleted.  I think that's good enough for a down migration.

jrafanie:
> Those are not actual tables in our schema though, right? Pretty sure those are backed by PG system tables/views.

Correct.  Right, sorry, I copied the comment from the core PR.  The models are remaining for those two models as we use the information from their backing `pg_stat_activity` and `pg_locks` tables in our application code, such as the connection counts.  We can replace that behavior if we want at another time.

jrafanie:
I'm adding a data migration to remove the various settings deleted in https://github.com/ManageIQ/manageiq/pull/20148

db/migrate/20200512161742_remove_vmdb_database_settings.rb
```diff
@@ -0,0 +1,18 @@
...
+    say_with_time('Removing RemoveVmdbDatabaseSettings') do
+      SettingsChange.where(:key => '/database/metrics_collection/collection_schedule').destroy_all
+      SettingsChange.where(:key => '/database/metrics_collection/daily_rollup_schedule').destroy_all
+      SettingsChange.where(:key => '/database/metrics_history/keep_daily_metrics').destroy_all
+      SettingsChange.where(:key => '/database/metrics_history/keep_hourly_metrics').destroy_all
+      SettingsChange.where(:key => '/database/metrics_history/purge_schedule').destroy_all
+      SettingsChange.where(:key => '/database/metrics_history/purge_window_size').destroy_all
+      SettingsChange.where(:key => '/database/metrics_history/queue_timeout').destroy_all
+      SettingsChange.where(:key => '/ui/display_ops_database').destroy_all
+      SettingsChange.where(:key => '/workers/worker_base/schedule_worker/log_database_statistics_interval').destroy_all
```
jrafanie, line 15: As @Fryguy mentioned in the UI PR, there were some settings that could be in the database, so I added a removal of all of the settings removed in the backend PR.  See above.

PR #485, opened 2020-05-19
change hardware mem usage to bigint

d-m-u:
At least memory_mb ought to not be an int cause we convert this as Keenan said into bytes and multiply by 1024*1024 which with aggregation will overflow and instead of a model cast we should be doing something ... something else. 

this changes the type, and has a size check on the down 

@miq-bot assign @kbrock 
or um, no bot after all

### DIFF
--- a/db/migrate/20200519213913_change_hardware_mem_types.rb
+++ b/db/migrate/20200519213913_change_hardware_mem_types.rb
@@ -1,0 +1,16 @@
+class ChangeHardwareMemTypes < ActiveRecord::Migration[5.2]
+  class Hardware < ActiveRecord::Base
+  end
+
+  def up
+    change_column :hardwares, :memory_mb, :bigint
+  end
+
+  def down
+    say_with_time "clamp oversized memory_mb values" do
+      Hardware.where("memory_mb > 2147483647").update_all(:memory_mb => 2_147_483_647)
+    end
+
+    change_column :hardwares, :memory_mb, :integer
+  end
+end
--- a/spec/migrations/20200519213913_change_hardware_mem_types_spec.rb
+++ b/spec/migrations/20200519213913_change_hardware_mem_types_spec.rb
@@ -1,0 +1,16 @@
+require_migration
+
+describe ChangeHardwareMemTypes do
+  migration_context :down do
+    let(:hardware) { migration_stub(:Hardware).create(:memory_mb => 2_147_483_648) }
+
+    it "caps > 2147483647" do
+      expect(hardware.memory_mb).to eq(2_147_483_648)
+
+      migrate
+      hardware.reload
+
+      expect(hardware).to have_attributes(:memory_mb => 2_147_483_647)
+    end
+  end
+end


db/migrate/20200519213913_change_hardware_mem_types.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeHardwareMemTypes < ActiveRecord::Migration[5.2]
+  def up
+    change_column :hardwares, :memory_mb,    :bigint
+    change_column :hardwares, :memory_usage, :bigint
+  end
+
+  def down
+    change_column :hardwares, :memory_mb,    :integer
+    change_column :hardwares, :memory_usage, :integer
```
agrare, line 9: If either of these overflows an integer type then the down migration fails:
```
Caused by:
PG::NumericValueOutOfRange: ERROR:  integer out of range
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/postgresql/database_statements.rb:75:in `exec'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/postgresql/database_statements.rb:75:in `block (2 levels) in execute'
/home/grare/adam/.gems/2.7.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies/interlock.rb:48:in `block in permit_concurrent_loads'
/home/grare/adam/.gems/2.7.0/gems/activesupport-5.2.4.3/lib/active_support/concurrency/share_lock.rb:187:in `yield_shares'
/home/grare/adam/.gems/2.7.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies/interlock.rb:47:in `permit_concurrent_loads'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/postgresql/database_statements.rb:74:in `block in execute'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/abstract_adapter.rb:581:in `block (2 levels) in log'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/abstract_adapter.rb:580:in `synchronize'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/abstract_adapter.rb:580:in `block in log'
/home/grare/adam/.gems/2.7.0/gems/activesupport-5.2.4.3/lib/active_support/notifications/instrumenter.rb:23:in `instrument'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/abstract_adapter.rb:571:in `log'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/postgresql/database_statements.rb:73:in `execute'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/connection_adapters/postgresql/schema_statements.rb:427:in `change_column'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/migration.rb:871:in `block in method_missing'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/migration.rb:840:in `block in say_with_time'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/migration.rb:840:in `say_with_time'
/home/grare/adam/.gems/2.7.0/gems/activerecord-5.2.4.3/lib/active_record/migration.rb:860:in `method_missing'
/home/grare/adam/src/manageiq/manageiq-schema/db/migrate/20200519213913_change_hardware_mem_types.rb:9:in `down'
```

kbrock: I don't think people will store bigints in this column. (famous last words)

Up until now, things have been working fine storing an integer in here. Is the statement that if a user starts storing a bigger number in here after we upgrade and then needs to downgrade, their db will freak out?

I'm ok with this edge case, especially since the edge case is on the rollback case 
(when they are stressed and it is after hours and it will become a sev 1)

agrare: The probability of us running a down migration _and_ someone storing >2147483647 is pretty low.  @Fryguy are you okay with this or do you think we need a data migration on the down to cap any values over the new max?

kbrock:
@agrare did put out there that it would be nice to store bytes in this column since that is what we store in other tables. I would like this since i basically create a virtual column to do the conversion. but we both don't understand the implications and it feels a lot bigger than just expanding the precision of this column

kbrock:
@Fryguy you good with this?

db/migrate/20200519213913_change_hardware_mem_types.rb
Fryguy: I actually this needs to be fixed properly in the down, especially since we know it's a problem.  I thought there was a way, when downcasting a column, to tell it to fix itself.

Fryguy: > I don't think people will store bigints in this column. (famous last words)

Isn't the whole point of this PR to store bigints in this column? :joy:

bdunne: Wow, almost 2 petabytes of RAM would be amazing. :smile:

agrare: Ought to be enough for anyone

kbrock: the actual point of the pr is to allow `sum(this column *1024*1024)` to not blow up. not the actual value in the column itself.

we do weird casting to convert this into bigint when we are doing this stuff. wanted it to work without the hoops.

Do people tend to rollback databases weeks after migrating to a new version?
Depending upon that window, I can't see someone introducing a really big number in a short term (unless as a bug, they put bytes instead of megs into that field)

Fryguy: Ah ok, that makes sense then.

I guess I'll ask the dumb question which is why we aren't storing bytes to begin with like every other column.  If this were bytes as stored it would be already a bigint.  Granted, there are likely users already using this column, so if we did this, we'd need an upgrade path.

agrare: > I guess I'll ask the dumb question which is why we aren't storing bytes to begin with like every other column.

This was my thought also, if we're going to change this we might as well change it to store bytes.  Maybe add a legacy method `memory_mb` that converts it assuming we can migrate all of the performance oriented callers (reporting/metrics)

Fryguy: Yeah, it probably won't happen, but it's also not hard to add a:

```
say_with_time "clamp oversized memory_mb values" do
  Hardware.where("memory_mb > 2147483647").update_all(:memory_mb => 2147483647)
end
```

kbrock:
I can't merge. @agrare do you want to take this?
Think @Fryguy punted (very busy today)

db/migrate/20200519213913_change_hardware_mem_types.rb
kbrock: that sounds reasonable. especially since it is a low probability event.

kbrock:
ok, so new proposal:

- rename `memory_mb` to `ram_size_in_bytes`
- `ram_size_in_bytes` is a bigint
- get the value `* 1024 * 1024`
- convert all virtual attributes that referenced memory_mb to use `ram_size_n_bytes`
- define `memory_mb` virtual attribute / ruby


Our current code defines `ram_size_in_bytes` based upon `memory_mb`.
We'll end up defining `memory_mb` based upon `ram_size_in_bytes`.

```ruby
  def ram_size_in_bytes
    memory_mb.to_i * 1.megabyte
  end
  # resulting sql: "(CAST("hardwares"."memory_mb" AS bigint) * 1048576)"
  virtual_attribute :ram_size_in_bytes, :integer, :arel => (lambda do |t|
    t.grouping(Arel::Nodes::Multiplication.new(Arel::Nodes::NamedFunction.new("CAST", [t[:memory_mb].as("bigint")]),
                                               1.megabyte))
  end)

```

kbrock:
cool. now all we need is to update it.

```ruby
def up
  Hardware.update_all("ram_size_in_bytes = ram_size_in_bytes * 1048576")
end

def down
  Hardware.update_all("ram_size_in_bytes = ram_size_in_bytes / 1048576")
end
```

db/migrate/20200519213913_change_hardware_mem_types.rb
```diff
@@ -0,0 +1,17 @@
+class ChangeHardwareMemTypes < ActiveRecord::Migration[5.2]
+  class Hardware < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  def up
+    change_column :hardwares, :memory_mb, :bigint
+    rename_column :hardwares, :memory_mb, :ram_size_in_bytes
+    Hardware.update_all("ram_size_in_bytes = ram_size_in_bytes * 1048576")
```
kbrock, line 9:
```suggestion
    Hardware.update_all("ram_size_in_bytes = COALESCE(ram_size_in_bytes * 1048576, 0)")
```

we'll get rid of the problematic null values.
and adding the default_value_for will fix it for already created records.

agrare:
So I really like moving this to bytes from mb but it does make this a bigger change overall and increases the risk.

How about we change the column type to bigint and clip large values on the down migration in this PR, cleanup just the casting in the aggregates in core.  Then separately we change the column to bytes, update callers in core, update the provider parsers, etc... in another patch set

kbrock:
@agrare yes, I would like this field to be a big int for casting reasons. but putting that aside

- changing this to a big int doesn't buy us much, nor does it introduce risk
- changing the field name introduces minimal? (moderate?) risk for ruby using our models
- changing the field name introduces much risk for ruby/others not using our models

I hear:
- You don't want to deal with the risk of changing the name at this time.
- You are acknowledging that the risk of bigint is small, so lets go ahead with that one.

code freeze is soonish (agreeing with why you may not want to change things now)

I think I'm good with your thoughts. but I wanted to be clear what I heard to make sure i didn't misunderstand you

agrare:
> changing this to a big int doesn't buy us much

Wasn't this the original purpose?

> You don't want to deal with the risk of changing the name at this time.

Yeah I realized that container providers use the hardware table (through nodes) and they use the batch strategy for refresh so the compatibility methods wouldn't work.

Seems like the best way forward is to take this in small chunks.

db/migrate/20200519213913_change_hardware_mem_types.rb
```diff
@@ -0,0 +1,17 @@
+class ChangeHardwareMemTypes < ActiveRecord::Migration[5.2]
+  class Hardware < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
```
agrare, line 3: I don't think hardware has a :type column for STI

PR #486, opened 2020-05-20
Migration for Startpage url updates for CB Assignments

h-kataria:
UI PR https://github.com/ManageIQ/manageiq-ui-classic/pull/7049

PR #487, opened 2020-05-20
Chargeback Reports features and startpage data migrations

h-kataria:
Data migration changes for product feature changes and startpage changes in https://github.com/ManageIQ/manageiq/pull/20190

UI PR https://github.com/ManageIQ/manageiq-ui-classic/pull/7058

### DIFF
--- a/db/migrate/20200520210200_update_chargeback_reports_startpage.rb
+++ b/db/migrate/20200520210200_update_chargeback_reports_startpage.rb
@@ -1,0 +1,25 @@
+class UpdateChargebackReportsStartpage < ActiveRecord::Migration[5.2]
+  class User < ActiveRecord::Base
+    serialize :settings, Hash
+  end
+
+  def up
+    say_with_time 'Updating starting page for users who had chargeback reports set' do
+      User.select(:id, :settings).each do |user|
+        if user.settings&.dig(:display, :startpage) == 'chargeback_reports/explorer'
+          user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'chargeback_report/show_list'}))
+        end
+      end
+    end
+  end
+
+  def down
+    say_with_time 'Updating starting page for users who had non-explorer chargeback reports pages set' do
+      User.select(:id, :settings).each do |user|
+        if user.settings&.dig(:display, :startpage) == 'chargeback_report/show_list'
+          user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'chargeback_reports/explorer'}))
+        end
+      end
+    end
+  end
+end
--- a/db/migrate/20200520210200_update_chargeback_reports_startpage.rb
+++ b/db/migrate/20200520210200_update_chargeback_reports_startpage.rb
@@ -1,11 +1,12 @@
 class UpdateChargebackReportsStartpage < ActiveRecord::Migration[5.2]
   class User < ActiveRecord::Base
     serialize :settings, Hash
+    include ActiveRecord::IdRegions
   end
 
   def up
     say_with_time 'Updating starting page for users who had chargeback reports set' do
-      User.select(:id, :settings).each do |user|
+      User.in_my_region.select(:id, :settings).each do |user|
         if user.settings&.dig(:display, :startpage) == 'chargeback_reports/explorer'
           user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'chargeback_report/show_list'}))
         end
@@ -15,7 +16,7 @@ class UpdateChargebackReportsStartpage < ActiveRecord::Migration[5.2]
 
   def down
     say_with_time 'Updating starting page for users who had non-explorer chargeback reports pages set' do
-      User.select(:id, :settings).each do |user|
+      User.in_my_region.select(:id, :settings).each do |user|
         if user.settings&.dig(:display, :startpage) == 'chargeback_report/show_list'
           user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'chargeback_reports/explorer'}))
         end
--- a/db/migrate/20200520211047_adjust_chargeback_reports_features.rb
+++ b/db/migrate/20200520211047_adjust_chargeback_reports_features.rb
@@ -32,8 +32,6 @@ class AdjustChargebackReportsFeatures < ActiveRecord::Migration[5.2]
     say_with_time 'Adjusting chargeback reports features to explorer views' do
       admin_feature = MiqProductFeature.find_or_create_by!(:identifier => 'chargeback_reports')
 
-      return if MiqRolesFeature.none?
-
       MiqRolesFeature.where(:miq_product_feature_id => admin_feature.id).each do |feature|
         old_feature_ids.each do |id|
           MiqRolesFeature.create!(:miq_product_feature_id => id, :miq_user_role_id => feature.miq_user_role_id)
--- a/db/migrate/20200520211047_adjust_chargeback_reports_features.rb
+++ b/db/migrate/20200520211047_adjust_chargeback_reports_features.rb
@@ -1,0 +1,52 @@
+class AdjustChargebackReportsFeatures < ActiveRecord::Migration[5.2]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+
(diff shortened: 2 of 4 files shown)


db/migrate/20200520210200_update_chargeback_reports_startpage.rb
```diff
@@ -0,0 +1,26 @@
+class UpdateChargebackReportsStartpage < ActiveRecord::Migration[5.2]
+  class User < ActiveRecord::Base
+    serialize :settings, Hash
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time 'Updating starting page for users who had chargeback reports set' do
+      User.in_my_region.select(:id, :settings).each do |user|
```
skateman, line 9: Why just in your region? Maybe I'm wrong, but that's not all the users :confused:

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
```diff
@@ -0,0 +1,62 @@
+class AdjustChargebackReportsFeatures < ActiveRecord::Migration[5.2]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+  class User < ActiveRecord::Base
```
skateman, line 4: You probably don't need this.

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
```diff
@@ -0,0 +1,62 @@
...
+    serialize :settings, Hash
+  end
+
+  OLD_FEATURES = %w[
```
skateman, line 8: Isn't this just `FEATURE_MAPPING.keys`?

h-kataria: @skateman made requested change

db/migrate/20200520210200_update_chargeback_reports_startpage.rb
h-kataria: @skateman i made this change to be consistent with changes requested here https://github.com/ManageIQ/manageiq-schema/pull/486#pullrequestreview-416181237

Fryguy: it's better to not touch records from another region in data migrations because it can potentially break replication (though usually it's ok)

spec/migrations/20200520211047_adjust_chargeback_reports_features_spec.rb
```diff
@@ -0,0 +1,65 @@
+require_migration
+
+describe AdjustChargebackReportsFeatures do
+  let(:user_role_id) { anonymous_class_with_id_regions.id_in_region(1, anonymous_class_with_id_regions.my_region_number) }
+  let(:feature_stub) { migration_stub :MiqProductFeature }
+  let(:roles_feature_stub) { migration_stub :MiqRolesFeature }
+  let(:user_stub) { migration_stub :User }
```
Fryguy, line 7: This is not used and should be removed

h-kataria: done

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
```diff
@@ -0,0 +1,52 @@
...
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    say_with_time 'Adjusting chargeback reports features for the non-explorer views' do
+      new_feature = MiqProductFeature.find_or_create_by!(:identifier => 'chargeback_reports_view')
+      MiqRolesFeature.where(:miq_product_feature_id => old_feature_ids).each do |feature|
+        MiqRolesFeature.create!(:miq_product_feature_id => new_feature.id, :miq_user_role_id => feature.miq_user_role_id)
+      end
```
Fryguy, line 19: I don't think this logic is correct.  What I read here is that for _each_ of csv, pdf, text, and reports, create an extra roles-features link of chargeback_reports_view.  So, if the role has, say, all 4 features, then this will create 4 duplicate rows with the chargeback_reports_view in roles-features, which I think violates some uniqueness somewhere.

If you change your specs and attach more than one productfeature to a role you should see it fail.

I think the correct code would look something like:

user_role_ids_with_old_features = MiqRolesFeature.where(:miq_product_feature_id => old_feature_ids).pluck(:miq_user_role_id).uniq
user_role_ids_with_old_features.each do |user_roles_id|
  MiqRolesFeature.create!(:miq_product_feature_id => new_feature.id, :miq_user_role_id => user_role_id)
end
```

File: db/migrate/20200520211047_adjust_chargeback_reports_features.rb [hunk 6]
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def down
+    return if MiqProductFeature.none?
+
+    say_with_time 'Adjusting chargeback reports features to explorer views' do
+      admin_feature = MiqProductFeature.find_or_create_by!(:identifier => 'chargeback_reports')
+
+      return if MiqRolesFeature.none?
```
Fryguy, line 34: This will end up skipping the reverse rename [below](https://github.com/ManageIQ/manageiq-schema/pull/487/files#diff-7c5b5ce01487a3575fbf23a171257674R42)...can you verify this in a spec?

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
h-kataria: @Fryguy you were right it did create multiple records for the new parent feature. Fixed migration and spec tests to verify changes, plus addressed https://github.com/ManageIQ/manageiq-schema/pull/487#discussion_r429449922

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
h-kataria: addressed [here](https://github.com/ManageIQ/manageiq-schema/pull/487/commits/b7108ad7b1b96e7fbf66a1fa022f8357ab529349#diff-05b72a996a9f2a2185bd92e1d7f3016dR65-R70)

h-kataria:
@Fryguy please re-review

db/migrate/20200520211047_adjust_chargeback_reports_features.rb
Fryguy: I don't see it addressed... basically, even if there are no MiqRolesFeatures, doesn't line 44 still have to happen?  I think this line can actually be removed completely

h-kataria: @Fryguy you are right, with that check the code would not have executed line 44, and ideally there will always be MiqRoleFeatures from seeding so i have removed that line. please take a look.

spec/migrations/20200520211047_adjust_chargeback_reports_features_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect(assigned.count).to eq(5)
+
+        expect(view_feature1.reload.identifier).to eq('chargeback_download_csv')
+        expect(view_feature2.reload.identifier).to eq('chargeback_download_pdf')
+        expect(view_feature3.reload.identifier).to eq('chargeback_download_text')
+        expect(view_feature4.reload.identifier).to eq('chargeback_report_only')
+      end
+
+      it 'skips feature renaming when no role features are set' do
+        migrate
```
Fryguy, line 66: @h-kataria Now that you change the logic, this test also needs to change to show that the features _are_ renamed when no role features are set

h-kataria: @Fryguy updated test to verify product feature renaming, please take a look

simaishi:
Jansa backport details:

```
$ git log -1
commit 728a94096acb011f23a8db62e3851217b59fb618
Date:   Mon Jun 1 10:59:24 2020 -0400

    Merge pull request #487 from h-kataria/cb_reports_startpage_update

    Chargeback Reports features and startpage data migrations

    (cherry picked from commit 3a66ffb1d5bb98e51a167aa5305c2c3fef77d0b4)
```

PR #492, opened 2020-06-10
set vm ancestry from relationship resource info

d-m-u:
we are working on converting vm genealogy to use ancestry. In order to do this, we need to use the existing relationship resource information on all vms with genealogy rels of type VmOrTemplate to convert those resource ids into vm ids. 

This PR 1) adds the necessary index and 2) gets the ancestry information from the relationships, splitting on slashes, then maps those back to the vm ancestry records. 

the first commit's the original sql which joined to vms
the second commit removes that join so we no longer have a potential issue with the ordering 

needs https://github.com/ManageIQ/manageiq/pull/20788, tests are green: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/225

### DIFF
--- a/db/migrate/20200607025146_add_ancestry_to_vm.rb
+++ b/db/migrate/20200607025146_add_ancestry_to_vm.rb
@@ -9,40 +9,73 @@ class AddAncestryToVm < ActiveRecord::Migration[5.2]
   class Relationship < ActiveRecord::Base
   end
 
-  #
-  # build sql that converts relationship ancestry with the id of the relationship to model-side ancestry with the id of the model
-  #
+  def up
+    add_column :vms, :ancestry, :string
+    add_index :vms, :ancestry
 
-  # def ancestry_resource_ids(relationship, model_class_name, id_range)
-  #   Relationship.where(
-  #     :relationship => relationship,
-  #     :resource_type => model_class_name,
-  #     :resource_id => id_range
-  #   ).where.not(:ancestry => nil).flat_map do |a_rels|
-  #     a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
-  #   end
-  # end
+    say_with_time("set vm ancestry from existing genealogy relationship resource information") do
+      connection.execute(transfer_relationships_to_ancestry(VmOrTemplate, 'VmOrTemplate', 'genealogy'))
+    end
 
-  def ancestry_resource_ids(relationship, model_class_name, id_range)
+    Relationship.where(:relationship => 'genealogy', :resource_type => 'VmOrTemplate', :resource_id => VmOrTemplate.all.select(:id)).delete_all
+  end
+
+  def down
+    say_with_time("create relationship records from vm ancestry") do
+      create_relationships_from_ancestry(VmOrTemplate.where(:template => false), Relationship, 'VmOrTemplate', 'genealogy')
+    end
+
+    remove_column :vms, :ancestry
+  end
+
+  # src is relationship, dest is vm
+  def transfer_relationships_to_ancestry(dest_model, resource_type, relationship)
+    ancestry_resources = ancestry_resource_ids(relationship, resource_type, dest_model.rails_sequence_range(dest_model.my_region_number))
+    ancestry_sources = ancestry_src_ids(ancestry_resources)
+    new_ancestors = ancestry_of_src_ids_for_src(ancestry_sources)
+    update_src(new_ancestors, dest_model)
+  end
+
+  # src is vm, dest is relationship
+  def create_relationships_from_ancestry(src_model, dest_model, resource_type, relationship)
+    children = src_model.select("ancestry::bigint").where("ancestry not like '%/%'")
+    rels = src_model.where.not(:ancestry => nil).or(src_model.where(:id => children)).map do |obj|
+      dest_model.create!(:relationship  => relationship,
+                         :resource_type => resource_type,
+                         :resource_id   => obj.id,
+                         :ancestry      => obj.ancestry)
+    end.index_by(&:resource_id)
+    rels.each do |_, r|
+      next if r.ancestry.nil?
+
+      ancestry = r.ancestry.split('/').map { |rel| rels[rel.to_i].id }.join('/')
+      r.update!(:ancestry => ancestry)
+    end
+  end
+
+  private
+
+  def ancestry_resource_ids(relationship, resource_type, id_range)
     <<-SQL
     SELECT a_rels.resource_id AS src_id, relationships.id AS rel_id, relationships.indx AS rel_indx
     FROM   relationships a_rels
     LEFT JOIN LATERAL UNNEST(STRING_TO_ARRAY(a_rels.ancestry, '/')::BIGINT[])
     WITH ORDINALITY AS relationships(id, indx) ON TRUE
     WHERE  a_rels.relationship = '#{relationship}'
-    AND    a_rels.resource_type = '#{model_class_name}'
+    AND    a_rels.resource_type = '#{resource_type}'
     AND    a_rels.resource_id BETWEEN #{id_range.first} AND #{id_range.last}
     SQL
-  end
 
-  # def ancestry_src_ids(ancestry_resources)
-  #   ancestry_resources.map do |rec|
-  #     {
-  #       src_id: rec.src_id,
-  #       ancestry_src_id: Relationship.find(rec[1]).resource_id
-  #     }
-  #   end
-  # end
+    # The above is a pure-SQL version of this similar Ruby code:
+    #
+    #   Relationship.where(
+    #     :relationship => relationship,
+    #     :resource_type => resource_type,
+    #     :resource_id => id_range
+    #   ).where.not(:ancestry => nil).flat_map do |a_rels|
+    #     a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
+    #   end
+  end
 
   def ancestry_src_ids(ancestry_resources)
     <<-SQL
@@ -52,16 +85,16 @@ class AddAncestryToVm < ActiveRecord::Migration[5.2]
     ON res_rels.id = ancestry_resources.rel_id
     ORDER BY ancestry_resources.src_id, ancestry_resources.rel_indx
     SQL
-  end
 
-  # def ancestry_of_src_ids_for_src(ancestry_sources)
-  #   ancestry_sources.group_by { |rec| rec.src_id }.map do |src_id, recs|
-  #     {
-  #       src_id: src_id,
-  #       new_ancestry: recs.map { |rec| rec.ancestry_src_id }.join('/')
-  #     }
-  #   end
-  # end
+    # The above is a pure-SQL version of this similar Ruby code:
+    #
+    #   ancestry_resources.map do |rec|
+    #     {
+    #       src_id: rec.src_id,
+    #       ancestry_src_id: Relationship.find(rec[1]).resource_id
+    #     }
+    #   end
+  end
 
   def ancestry_of_src_ids_for_src(ancestry_sources)
     <<-SQL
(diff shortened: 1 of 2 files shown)


Fryguy (dismissed):
This is awesome.  Great idea to get genealogy out of there!

d-m-u:
hey @Fryguy could I please ask you to review this again?

d-m-u:
could I please get someone to review this?

d-m-u:
commenting so I don't forget: it's work in progress because I think Keenan's right about what's missing: "... is [it] possible that not all the ids are in the database - could that be the issue?"

kbrock:
aside: we have some notes in [gist](https://gist.github.com/kbrock/3ee8e721f0e50218621652ac55b81652) - may want to bring that into this PR for future reference (or not...)

NickLaMuro:
Ran this migration against a large customer database and it passed without issue.

Does that mean I understand what this SQL is doing?  Nope!

But it didn't blow up.  🎉

d-m-u:
@kbrock I'm pretty convinced that the order (and everything else) on this is correct and at this point I'm not sure how else to test that.

d-m-u:
hey @agrare could I please ask you to review this? I think you might have a stake here.

d-m-u:
Okay, well. Even though we consensus'd on closing [the pr from hell](https://github.com/ManageIQ/manageiq/pull/20274) because it has too much overhead to realistically be a viable next step, I still think this PR was both an excellent idea and low enough overhead to make it worth merging. I don't know what else I need to do to get anyone else to believe me about that (thanks for the review, @NickLaMuro).

kbrock:
outstanding:

we need to fix the "PR from hell"
To do that, we need to remove relationships from vms
To do that, we need to introduce relationships to `ems_folders`

There is a possible issue around loading resource_pools - but I think this will resolve itself.

d-m-u:
thanks @kbrock

Fryguy:
Ignore the rubocops...I've added them to the ignore list here - https://github.com/ManageIQ/manageiq-style/pull/8

spec/migrations/20200607025146_add_ancestry_to_vm_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+
+    context "tree with depth 3" do
+      #         parent
+      #         child
+      #   second_level_child
+
+      it 'updates ancestry' do
+        parent_rel = rel_stub.create!(:relationship => 'genealogy', :ancestry => nil, :resource_type => 'VmOrTemplate', :resource_id => root.id)
+        child_rel = rel_stub.create!(:relationship => 'genealogy', :ancestry => ancestry_for(parent_rel), :resource_type => 'VmOrTemplate', :resource_id => vm.id)
+        rel_stub.create!(:relationship => 'genealogy', :ancestry => ancestry_for(child_rel, parent_rel), :resource_type => 'VmOrTemplate', :resource_id => vm2.id)
```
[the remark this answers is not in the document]
kbrock: They say comments are a code smell.
The same folks also say to comment your code.
Sigh.

So the only way to do that is to create self documenting code.

https://github.com/ManageIQ/manageiq/blob/master/spec/models/mixins/relationship_mixin_spec.rb#L3-L7

Sorry, just remembered that code when I was typing the last response. It was amusing that the code was for relationship mixin, since we are ... yea

If you look at the evolution from your first version to the second version, and you extrapulate, you'll see how someone ended up here.

db/migrate/20200607025146_add_ancestry_to_vm.rb
```diff
@@ -0,0 +1,112 @@
...
+  #   Relationship.where(
+  #     :relationship => relationship,
+  #     :resource_type => model_class_name,
+  #     :resource_id => id_range
+  #   ).where.not(:ancestry => nil).flat_map do |a_rels|
+  #     a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
+  #   end
+  # end
+
+  def ancestry_resource_ids(relationship, model_class_name, id_range)
```
kbrock, line 26: I know we had to put in ruby code that mimics the sql because it is just too much.

but (at least now when the code is fresh in my mind) like the end result.

getting this into a migration routine may be nice - not sure if we want something like this in ancestry.
Definitely want a better way to populate ancestry based upon sql. (sorry tangent)
^ ignore the ancestry comment. want to keep the thought out there but  don't want us to move towards that right now

kbrock (commented):
sorry - I know we went over the specs a few times. this link was itching at the back of my mind and only surfaced today

db/migrate/20200607025146_add_ancestry_to_vm.rb
d-m-u: `but (at least now when the code is fresh in my mind) like the end result.`
Are you missing a verb? Is this saying that you do like the end result here now, or that you do not?

kbrock: sorry, missed a noun. I **do** like the end result

kbrock:
this looks nice.

I was thinking that useing https://github.com/ManageIQ/manageiq/blob/master/spec/models/mixins/relationship_mixin_spec.rb#L870-L886 would allow you to pass in a hash and not need to create any of the objects.

If this is more clear for you then this is good

d-m-u:
@kbrock sorry, yes. the reason i chose to not go with that suggestion is that that particular code you linked to relies on 1) the parent setter which relies on 2) with_relationship_type which relies on 3) caching methods that I didn't really want to bring into all this. My choice was either to rewrite that and keep the functionality without the extra code, or this approach, which felt ... easier.

db/migrate/20200607025146_add_ancestry_to_vm.rb
kbrock: um, I missed the "I" which is a pro noun.

I guess I then added the do to answer your question of do or do not (that is the question).

So I stand by my noun statement.

kbrock:
hey @d-m-u - where did we leave this? were the spec changes good to go?
I figure we're going to duplicate those tests across a few migrations so good to have a pattern together.

Also, did you decide to pull the migration out into a separate class?

d-m-u:
marking as work in progress: i'm rewriting the specs per the last comment

kbrock (commented):
sorry - could you pull out the ancestry_resource_ids into an AncestryMigrationhelper

then we can ship this

kbrock:
sigh - sorry that I had you change perfectly working code

db/migrate/20200607025146_add_ancestry_to_vm.rb
```diff
@@ -0,0 +1,38 @@
+class AddAncestryToVm < ActiveRecord::Migration[5.2]
+  include RelationshipAncestryConverter
```
kbrock, line 2: this should be loaded via [spec_helper]
Probably want to scope this

```suggestion
  include ::RelationshipAncestryConverter
```

[spec_helper]: https://github.com/ManageIQ/manageiq-schema/blob/d30ccc15125a7c19b23182ba46bd96a05a12e9fb/spec/spec_helper.rb#L14

if that doesn't work, wonder if using extend `instead` would do what you need - but 
include` seems to be the right thing to me right now

kbrock (commented):
thanks for agreeing to refactor code that was working

db/migrate/20200607025146_add_ancestry_to_vm.rb
d-m-u: thank you  👍

d-m-u:
marking [WIP] because I'm mildly concerned about the test failure (on https://github.com/ManageIQ/manageiq/pull/20788): we can't serialize objects with unbound methods, so anything previously that got queued that has our ancestry binding on latest code will be broken.

d-m-u:
https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/225
yay

lib/relationship_ancestry_converter_helper.rb
```diff
@@ -0,0 +1,105 @@
+module RelationshipAncestryConverterHelper
```
Fryguy, line 1: Are you planning on using this for other tables in future migrations?  If not, might just be better to embed this module directly into the migration that uses it.

d-m-u: Yes. We're planning on using this for https://github.com/ManageIQ/manageiq-schema/pull/497 as well.

Fryguy: Is it possible to write specs for it as a standalone module or is that kind of impossible?

lib/relationship_ancestry_converter_helper.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  # def ancestry_resource_ids(relationship, resource_type, id_range)
+  #   Relationship.where(
+  #     :relationship => relationship,
+  #     :resource_type => resource_type,
+  #     :resource_id => id_range
+  #   ).where.not(:ancestry => nil).flat_map do |a_rels|
+  #     a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
+  #   end
+  # end
```
Fryguy, line 41: I'm curious about these commented bits...are they for the other PR in the future?  If so, should they just be added at that time?

d-m-u: I wanted ruby versions of the sql for each method just to make it easier to understand.

Fryguy: ohhhh I see - I didn't realize they were the same/related method but Ruby - wonder if it's better to organize them together - Something like:

Before:

```ruby
  # def ancestry_resource_ids(relationship, resource_type, id_range)
  #   Relationship.where(
  #     :relationship => relationship,
  #     :resource_type => resource_type,
  #     :resource_id => id_range
  #   ).where.not(:ancestry => nil).flat_map do |a_rels|
  #     a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
  #   end
  # end

  def ancestry_resource_ids(relationship, resource_type, id_range)
    <<-SQL
    SELECT a_rels.resource_id AS src_id, relationships.id AS rel_id, relationships.indx AS rel_indx
    FROM   relationships a_rels
    LEFT JOIN LATERAL UNNEST(STRING_TO_ARRAY(a_rels.ancestry, '/')::BIGINT[])
    WITH ORDINALITY AS relationships(id, indx) ON TRUE
    WHERE  a_rels.relationship = '#{relationship}'
    AND    a_rels.resource_type = '#{resource_type}'
    AND    a_rels.resource_id BETWEEN #{id_range.first} AND #{id_range.last}
    SQL
  end
```

After:

```ruby
  def ancestry_resource_ids(relationship, resource_type, id_range)
    <<-SQL
    SELECT a_rels.resource_id AS src_id, relationships.id AS rel_id, relationships.indx AS rel_indx
    FROM   relationships a_rels
    LEFT JOIN LATERAL UNNEST(STRING_TO_ARRAY(a_rels.ancestry, '/')::BIGINT[])
    WITH ORDINALITY AS relationships(id, indx) ON TRUE
    WHERE  a_rels.relationship = '#{relationship}'
    AND    a_rels.resource_type = '#{resource_type}'
    AND    a_rels.resource_id BETWEEN #{id_range.first} AND #{id_range.last}
    SQL

    # The above is a pure-SQL version of this similar Ruby code:
    #
    # Relationship.where(
    #   :relationship => relationship,
    #   :resource_type => resource_type,
    #   :resource_id => id_range
    # ).where.not(:ancestry => nil).flat_map do |a_rels|
    #   a_rels.ancestry.split('/').each_with_index.map { |rel_id, indx| [a_rels.resource_id, rel_id, indx] }
    # end
  end
```

lib/relationship_ancestry_converter_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+    LEFT JOIN LATERAL UNNEST(STRING_TO_ARRAY(a_rels.ancestry, '/')::BIGINT[])
+    WITH ORDINALITY AS relationships(id, indx) ON TRUE
+    WHERE  a_rels.relationship = '#{relationship}'
+    AND    a_rels.resource_type = '#{resource_type}'
+    AND    a_rels.resource_id BETWEEN #{id_range.first} AND #{id_range.last}
+    SQL
+
+    # The above is a pure-SQL version of this similar Ruby code:
+    #
+    # def ancestry_resource_ids(relationship, resource_type, id_range)
```
chessbyte, line 46: redundant method definition in the comment.  In other words, suggest to remove the `def ancestry_resource_ids(...)` in the comments and the correlating `end`

d-m-u: I don't agree. I think the definitions are useful here because tracking the method params has been really important for catching bugs and understanding what this is doing.

lib/relationship_ancestry_converter_helper.rb
d-m-u: Even with the redundant method signatures, @Fryguy, you struggled to understand what this is doing, so I don't exactly think taking them out is going to help. This PR's kind of fragile, and this is about the third attempt of this PR.

Fryguy: No I think combining them together makes the SQL easier to understand, which was my biggest struggle...I can use the two together to understand the intent.

lib/relationship_ancestry_converter_helper.rb
kbrock: it is going to be used 3 times max? I'd honestly prefer copy/past over spending the money/time on testing this as an individual component. The output of the code is already tested for this particular use case. and will be tested for the other use case as well.

kbrock (approved):
Feel like this has dragged a bit and I have already asked to rewrite the sql a few times and then extract it.

Lets get this to a reasonable point and get it in

Fryguy:
We can't merge it until the other side is available (or is it done? don't see any linked PRs in the OP, and there are a lot of comments here, so not sure)

d-m-u:
https://github.com/ManageIQ/manageiq/pull/20788

d-m-u:
Could I get this merged, please?

kbrock:
@Fryguy I too would like to bump this.

Premise:

```ruby
add_column :vms, :ancestry, :string
Vm.update(:ancestry => vms.relationships.ancestry)
Vm.relationships.delete_all
```

Of course the devil in the details has a cascading effect:

1. Change ancestry from `relationships.id` to `vms.id`.
2. Change `O(N + log(N))` to `O(1)` since the substitution thing mentioned above is is more a tad complex than it looks.
3. The sql needed for `O(1)` is more obtuse than the ruby version.

d-m-u:
Okay, I'm marking this and the core pr (https://github.com/ManageIQ/manageiq/pull/20788) as work in progress again because I'd like to rewrite the tests.

d-m-u:
@kbrock could I get another look at this please?

kbrock (approved):
this looks nice.

ship it

Fryguy (approved):
I think this is good, but would like @agrare to also review and approve.

spec/migrations/20200607025146_add_ancestry_to_vm_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      yield(parent, tree)
+    when Array
+      tree.each { |node| traverse(node, parent, &block) }
+    when Hash
+      tree.each do |key, children|
+        yield(parent, key)
+        traverse(children, parent + [key], &block)
+      end
+    else
+      puts "curious type: #{tree.class.name}"
```
Fryguy, line 171: Minor, but should this raise an Exception instead?  Either way, this is a debugging statement so it should probably be removed as per the bot's request.

db/migrate/20200607025146_add_ancestry_to_vm.rb
```diff
@@ -0,0 +1,130 @@
...
+  def transfer_relationships_to_ancestry(dest_model, resource_type, relationship)
+    ancestry_resources = ancestry_resource_ids(relationship, resource_type, dest_model.rails_sequence_range(dest_model.my_region_number))
+    ancestry_sources = ancestry_src_ids(ancestry_resources)
+    new_ancestors = ancestry_of_src_ids_for_src(ancestry_sources)
+    connection.execute(update_src(new_ancestors, dest_model))
+  end
+
+  # src is vm, dest is relationship
+  def create_relationships_from_ancestry(src_model, dest_model, resource_type, relationship)
+    children = src_model.select("ancestry::bigint").where("ancestry not like '%/%'")
```
Fryguy, line 41: SUPER minor, but 
```suggestion
    children = src_model.select("ancestry::bigint").where("ancestry NOT LIKE '%/%'")
```

Fryguy:
From a testing perspective, have you run this against a "real" customer database?  I'm curious if it migrates correctly.  If so, that would give a huge confidence boost.  If it fails, then that's more test cases.

d-m-u:
@Fryguy yes, @NickLaMuro was awesome and provided the customer database that was used to test this. Please see his earlier comment: https://github.com/ManageIQ/manageiq-schema/pull/492#issuecomment-659679138

d-m-u:
@agrare sorry yes, it's schema then core then the providers

agrare (commented):
:+1: LGTM after @Fryguy 's two suggestions are fixed

PR #76, opened 2017-09-26
Fix migration name typo: s/shorback/showback/

Fryguy:
cc @aljesusg @sergio-ocon 

@chessbyte Please merge.

PR #75, opened 2017-09-25
Change the table used for LXCA config patterns

skovic:
Recently, a new table named _configuration_templates_ was created to hold data related to configuration patterns. Based on feedback received from Red Hat, this PR updates an existing table, _customization_scripts_, to hold configuration pattern data instead.

PR #73, opened 2017-09-22
Change dialog field description to text field to accomodate larger sizes

d-m-u:
Text type allows us to use dialog field descriptions for help text and eventually, longer messages and maybe even links.

PR #493, opened 2020-06-29
bump more core extensions

d-m-u:
more core ext's on 4.1.0

update for update sake

https://github.com/ManageIQ/more_core_extensions/compare/179bf40..e5b4501
 - Added Ruby 2.7 support [[#79](https://github.com/ManageIQ/more_core_extensions/pull/79)]
 - Added Process#pause, Process#resume, and Process#alive? [[#73](https://github.com/ManageIQ/more_core_extensions/pull/73)]

array added * `#compact_map` - Collect non-nil results from the block
array added `#tabular_sort` - Sorts an Array of Hashes by specific columns

hierarchy added `#descendant_get` - Returns the descendant with a given name

the two breaking changes:
- **BREAKING**: Moved Object#descendant_get to Class#descendant_get [[#75](https://github.com/ManageIQ/more_core_extensions/pull/75)]
- **BREAKING**: Removed deprecated Enumerable#stable_sort_by [[#76](https://github.com/ManageIQ/more_core_extensions/pull/76)]

a minor header output change was made that hasn't been released yet to make tableize more markdown compliant

see https://github.com/ManageIQ/linux_admin/pull/221

### DIFF
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activerecord-id_regions", "~> 0.3.0"
   s.add_dependency "linux_admin", "~> 2.0"
   s.add_dependency "manageiq-password", "~> 0.3"
-  s.add_dependency "more_core_extensions", "~> 3.5"
+  s.add_dependency "more_core_extensions", ">= 3.5", "< 5"
   s.add_dependency "pg"
   s.add_dependency "pg-pglogical", "~> 2.1.1"
   s.add_dependency "rails", "~>5.2.4", ">=5.2.4.3"


manageiq-schema.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activerecord-id_regions", "~> 0.3.0"
   s.add_dependency "linux_admin", "~> 2.0"
   s.add_dependency "manageiq-password", "~> 0.3"
-  s.add_dependency "more_core_extensions", "~> 3.5"
+  s.add_dependency "more_core_extensions", "~> 4.0"
```
bdunne, line 22: Maybe this should be `">= 3.5", "< 5"`?  In #244 it sounds like we only really care that the minimum is 3.5.

PR #498, opened 2020-07-28
add hidden flag to clusters

d-m-u:
Part of the effort to sort out folders involves adding this hidden flag thing to clusters. I don't know what we're wanting to call it really.


kbrock:
@agrare I'm thinking this could be separate from other changes with relationships and ancestry.

Do you already have a branch for "having an `ems_cluster` for all hosts"?

agrare:
> Do you already have a branch for "having an ems_cluster for all hosts"?

No I don't yet but I'll create one after this is in

PR #499, opened 2020-07-28
add folder references

d-m-u:
as part of the effort to clean up ems folders, we need references to folders on a bunch of objects.

### DIFF
--- a/db/migrate/20200728151356_add_references_to_ems_folder.rb
+++ b/db/migrate/20200728151356_add_references_to_ems_folder.rb
@@ -1,0 +1,10 @@
+class AddReferencesToEmsFolder < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :vms,            :resource_pool, :type => :bigint, :index => true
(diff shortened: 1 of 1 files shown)


kbrock (commented):
@d-m-u thanks for starting this
This will be a very cool addition
I'm guessing we will create a pr on the miq side with the ugly ancestry patch


I was thinking that ems folders would be an ancestry model
And resource pool would have ancestry as well

```Ruby
Vm.belongs_to :resource_pool
Vm.belongs_to :ems_folder
Host.belong_to :ems_ref # better name here
ResourcePool.belongs_to :host
ResourceRool.belongs_to :ems_cluster
```

In the future, @agrare will change our code to not delete the default cluster form resource pool and all hosts will have a cluster (or it is the other way around - but it is for another time)

d-m-u:
> I was thinking that ems folders would be an ancestry model
And resource pool would have ancestry as well

https://github.com/ManageIQ/manageiq/pull/20392

d-m-u:
same with this: @kbrock could I get a comment from you here about what I need to go here to get this moving forward?

PR #503, opened 2020-08-24
Add relationship from sessions to users

gtanzillo:
Related to https://github.com/ManageIQ/manageiq/issues/20476

### DIFF
--- a/db/migrate/20200819184941_add_user_to_sessions.rb
+++ b/db/migrate/20200819184941_add_user_to_sessions.rb
@@ -1,0 +1,5 @@
+class AddUserToSessions < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :sessions, :user
+  end
+end


db/migrate/20200819184941_add_user_to_sessions.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserToSessions < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :sessions, :user
```
Fryguy, line 3: Can you verify this creates a bigint column? We have other PRs where I'm not sure.

gtanzillo: Yes, just verified it in my local DB

gtanzillo: Same as the `id` column, `int8`

Issue #502, opened 2020-08-19
[RFE] Speed up `rake test:vmdb:setup`


NickLaMuro:
Migrations take a while to run (see benchmarks), and probably could be collapsed to save time at this point.

We have 7ish years of migrations built up that have to run on every build that a `test:vmdb:setup` or `evm:db:reset` task is run.  This takes a lot of time for devs and build time on CI with no real value in doing that beyond them having them run semi-often in this repo.

Benchmarks
----------

Note:  Times below don't include `db:seed`, which could also be improved, but not the focus this RFE.

### `rake db:migrate` (before)

```
$ time rake test:vmdb:setup

real    1m36.398s
user    1m6.130s
sys     0m4.525s
```

### `rake db:shema:load` (after)

```console
$ time rake evm:db:destroy db:schema:load

...

real    0m14.210s
user    0m6.666s
sys     0m3.044
```

(Note:  `metrics_rollups_*` style tables are commented out for these to properly run, but the performance impact of those extra would be minimal, as most of the 14sec above is booting the Rails env)

Proposals
---------

One of the following probably could be done to solve this:

- Fix `rake db:schema:load` (in other words, fix the `db/schema.rb` file) so it can be run (hard)
- Do another `_collapse_initial_migration.rb` that is up to date (hard?)

The advantage to the first is that we can execute the "faster" code pretty much any time after a `bin/update` has been done.  So we can get to a state of re:seeding much quicker, not just after a "collapse" has happened.  That said, updating how the schema is generated to support this would be probably a significant bit of monkey patching to accomplish, and not a quick turn around.

That said, the "collapse" migration is basically a schema, so figuring out how collapse would probably lead to how we could generate the schema.  It also might be worth "archiving" migrations after every release so that migrations are "versioned" to some degree.

Testing is also something we probably would want to consider with either approach.  The advantage again with the first is that we don't have to be concerned using it in production, since it is only a dev speed up.  Making sure we get the collapse correct would probably require some testing interface that runs the full collection of migrations from the beginning, and then tests them from a snapshot onward (if something like this doesn't already exist).

Links
-----

* https://github.com/ManageIQ/manageiq-schema/blob/62d91b62/db/migrate/20130923182042_collapsed_initial_migration.rb

NickLaMuro:
[...]

PR #504, opened 2020-08-26
Fix `db/schema.rb` generation

NickLaMuro:
This is a first pass at the "Fix `rake db:schema:load`" option for https://github.com/ManageIQ/manageiq-schema/issues/502

What this does
--------------

The first part of the changes in this PR are a refactoring that takes much of what is coded into individual migrations and the `lib/migration_helper.rb`, and applies them to the `ActiveRecord::Base.connection`.

The other part of the changes is creating introspection for the `db/schema.rb` generation via the MiqSchemaDumper, which is a collection of additions for everything that was refactored into the `.connection` to generate the `schema.rb` with our custom functions/contrainst/views/etc.

The functions that were created on (moved to) the `.connection`:

- `.add_trigger`
- `.create_miq_metric_view`
- `.add_miq_metric_table_inheritance`

(and the `drop_*` equivalent functions for each of those)

The in addtion, there were a few introspection methods added there as well:

- `.connection.triggers`
- `.connection.views`

Which can be used to query the database to determine what exist for the `triggers` and `views`.

In addtion there is a additional introspection method in the `MiqSchemaDumper` called `determine_table_parent`, which is used to determine the table parent for a given metrics partition table.

Things for Reviewers to consider
---------------------------------

* Do we just want to leverage gems for things (like `scenic` for VIEWS) instead, or are the `MIQ` specifics enough to warrant me rolling it ourselves?
* Could some of my introspection queries be done better in a more "PostgreSQL DB Admin Approved" fashion?
* I used two different naming conventions for methods (with/without the miq_ prefix). Should we standardize?
* Are things like the `METRIC_ROLLUP_TABLE_REGEXP` too specific and do we want to make them more generic, or is it fine to specifically target tables for these additions to be more "surgical" about these changes?

Testing
-------

I have been testing by just running the following two commands:

```console
$ RAILS_ENV=development rake db:environment:set evm:db:destroy db:migrate > /dev/null
$ RAILS_ENV=development rake db:environment:set evm:db:destroy db:schema:load db:seed > /dev/null
```

And also inspected the contents of the `db/schema.rb`, which should now include entries for the above functions.  It should look something like this at the bottom:

```ruby
  add_miq_metric_table_inheritance "metric_rollups_01", "metric_rollups_base", :conditions => ["capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?", "realtime", "01"]
  add_miq_metric_table_inheritance "metric_rollups_02", "metric_rollups_base", :conditions => ["capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?", "realtime", "02"]
  add_miq_metric_table_inheritance "metric_rollups_03", "metric_rollups_base", :conditions => ["capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?", "realtime", "03"]
  # ...
  add_miq_metric_table_inheritance "metrics_00", "metrics_base", :conditions => ["capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?", "realtime", "00"]
  add_miq_metric_table_inheritance "metrics_01", "metrics_base", :conditions => ["capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?", "realtime", "01"]
  add_miq_metric_table_inheritance "metrics_02", "metrics_base", :conditions => ["capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?", "realtime", "02"]
  add_miq_metric_table_inheritance "metrics_03", "metrics_base", :conditions => ["capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?", "realtime", "03"]
  add_miq_metric_table_inheritance "metrics_04", "metrics_base", :conditions => ["capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?", "realtime", "04"]
  # ...

  create_miq_metric_view "metrics"

  create_miq_metric_view "metric_rollups"

  add_trigger "insteadof", "metrics", "metrics_partition", <<-SQL
          CASE EXTRACT(HOUR FROM NEW.timestamp)
            WHEN 0 THEN
              INSERT INTO metrics_00 VALUES (NEW.*);
           ...
  SQL

  add_trigger "insteadof", "metric_rollups", "metric_rollups_partition", <<-SQL
          CASE EXTRACT(MONTH FROM NEW.timestamp)
            WHEN 1 THEN
              INSERT INTO metric_rollups_01 VALUES (NEW.*);
          ....
  SQL
```

### DIFF
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -3,6 +3,38 @@ module ManageIQ
     module SchemaDumper
       METRIC_ROLLUP_TABLE_REGEXP = /metrics?_.*\d\d$/.freeze
 
+      def tables(stream)
+        super
+        triggers(stream)
+      end
+
+      def triggers(stream)
+        @connection.triggers.each do |(direction, table, name, body)|
+          # Inverse of the case statement in add_trigger_hook
+          direction_key = direction.downcase.gsub(/\s/, "")
+
+          # Formats the SQL we get from the database to do that following:
+          #
+          # - Removes the "BEGIN" and "END;" stanzas
+          # - Chomps the new lines from the beginning (keeps existing indent)
+          # - Strips whitespace from the end (no need to worry about indent)
+          # - Indents by four spaces
+          #
+          # For formatting, we want to ensure at least some bit of an indent,
+          # regardless of how the SQL string was originally added.
+          #
+          formatted_body = body.gsub(/(BEGIN$|^END;$)/, "")
+                               .reverse.chomp.reverse.rstrip
+                               .indent(4)
+
+          stream.puts "  add_trigger #{direction_key.inspect}, "       \
+                                    "#{table.inspect}, "               \
+                                    "#{name.inspect}, "                \
+                                    "<<-SQL\n#{formatted_body}\n  SQL"
+          stream.puts
+        end
+      end
+
       private
 
       def column_spec_for_primary_key(column)
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -12,7 +12,19 @@ module ManageIQ
 
       def table(table, stream)
         super
+        add_id_column_comment(table, stream)
         track_miq_metric_table_inheritance(table)
+      end
+
+      def add_id_column_comment(table, stream)
+        pk      = @connection.primary_key(table)
+        pkcol   = @connection.columns(table).detect { |c| c.name == pk }
+        comment = pkcol.comment
+
+        return unless comment
+
+        stream.puts "  change_column_comment #{remove_prefix_and_suffix(table).inspect}, #{pk.inspect}, #{comment.inspect}"
+        stream.puts
       end
 
       # Must be done after all of the table definitions since `metrics_01` is
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -1,12 +1,40 @@
 module ManageIQ
   module Schema
     module SchemaDumper
-      METRIC_ROLLUP_TABLE_REGEXP = /metrics?_.*\d\d$/.freeze
+      METRIC_ROLLUP_TABLE_REGEXP = /(?<METRIC_TYPE>metrics?_?.*)_(?<CHILD_TABLE_NUM>\d\d)$/.freeze
 
       def tables(stream)
         super
+        miq_metric_table_contraints(stream)
         miq_metric_views(stream)
         triggers(stream)
+      end
+
+      def table(table, stream)
+        super
+        track_miq_metric_table_inheritance(table)
+      end
+
+      # Must be done after all of the table definitions since `metrics_01` is
+      # dumpped prior to `metrics_base`, etc.
+      #
+      def miq_metric_table_contraints(stream)
+        inherited_metrics_tables.each do |(table, inherit_from)|
+          match             = table.match(METRIC_ROLLUP_TABLE_REGEXP)
+          child_table       = remove_prefix_and_suffix(table).inspect
+          primary_condition = if inherit_from.include?("rollup")
+                                "capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?"
+                              else
+                                "capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?"
+                              end
+          conditions        = [primary_condition, "realtime", match[:CHILD_TABLE_NUM]]
+
+          stream.puts "  add_miq_metric_table_inheritance #{child_table}, " \
+                          "#{inherit_from.inspect}, "                       \
+                          ":conditions => #{conditions.inspect}"
+        end
+
+        stream.puts
       end
 
       def miq_metric_views(stream)
@@ -44,6 +72,26 @@ module ManageIQ
       end
 
       private
+
+      def track_miq_metric_table_inheritance(table)
+        return unless table.match?(METRIC_ROLLUP_TABLE_REGEXP)
+
+        inherit_from              = determine_table_parent(table)
+        inherited_metrics_tables << [table, inherit_from]
+      end
+
+      def inherited_metrics_tables
+        @inherited_metrics_tables ||= []
+      end
+
+      def determine_table_parent(table_name)
+        @connection.execute(<<-SQL).first["parent_table"]
+          SELECT pg_class.relname AS parent_table
+          FROM   pg_catalog.pg_inherits
+          JOIN pg_class ON pg_class.oid = pg_inherits.inhparent
+          WHERE inhrelid = '#{table_name}'::regclass
+        SQL
+      end
 
       def column_spec_for_primary_key(column)
         result = super
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -1,0 +1,15 @@
+module ManageIQ
+  module Schema
+    module SchemaDumper
+      METRIC_ROLLUP_TABLE_REGEXP = /metrics?_.*\d\d$/.freeze
+
+      private
+
+      def column_spec_for_primary_key(column)
+        result = super
+        result.delete(:default) if column.table_name.match?(METRIC_ROLLUP_TABLE_REGEXP)
+        result
+      end
+    end
+  end
+end
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -5,7 +5,15 @@ module ManageIQ
 
       def tables(stream)
         super
+        miq_metric_views(stream)
         triggers(stream)
+      end
+
+      def miq_metric_views(stream)
+        @connection.views.each do |view|
+          stream.puts "  create_miq_metric_view #{view["name"].inspect}"
+          stream.puts
+        end
       end
 
       def triggers(stream)
--- a/lib/migration_helper.rb
+++ b/lib/migration_helper.rb
@@ -54,51 +54,10 @@ module MigrationHelper
   # Triggers
   #
 
+  # Keeping this helper to suppress the SQL body from the `say_with_time` output
   def add_trigger(direction, table, name, body)
     say_with_time("add_trigger(:#{direction}, :#{table}, :#{name})") do
-      add_trigger_function(name, body)
-      add_trigger_hook(direction, name, table, name)
-    end
-  end
-
-  def add_trigger_function(name, body)
-    quoted_name = connection.quote_table_name(name)
-    quoted_body = connection.quote("BEGIN\n#{body}\nEND;\n")
-
-    connection.execute <<-EOSQL, 'Create trigger function'
-      CREATE OR REPLACE FUNCTION #{quoted_name}()
-      RETURNS TRIGGER AS #{quoted_body}
-      LANGUAGE plpgsql;
-    EOSQL
-  end
-
-  def add_trigger_hook(direction, name, table, function)
-    quoted_name = connection.quote_column_name(name)
-    quoted_table = connection.quote_table_name(table)
-    quoted_function = connection.quote_table_name(function)
-    safe_direction = case(direction.downcase)
-                     when 'before'
-                       'BEFORE'
-                     when 'after'
-                       'AFTER'
-                     when 'insteadof'
-                       'INSTEAD OF'
-                     end
-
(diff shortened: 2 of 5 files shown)


lib/manageiq/schema/miq_command_recorder.rb
```diff
@@ -0,0 +1,21 @@
+module ManageIQ
+  module Schema
+    module MiqCommandRecorder
```
NickLaMuro, line 3: Note:  Most of these are _**not**_ used, but I think this is intended for `def change` methods in a migration, so that they can be inverted if needed.  So there most likely is more to make this "feature complete" for future migrations, if any of these methods were to be used again, but arguably that is outside the scope of this PR.

Fryguy, line 3: super minor, but I think we can remove the Miq prefix from all of these since they are namespaced under the ManageIQ namespace anyway.

lib/manageiq/schema/miq_schema_dumper.rb
```diff
@@ -0,0 +1,110 @@
...
+                            "realtime",
+                            match[:CHILD_TABLE_NUM]
+                          ]
+                        else
+                          [
+                            "capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?",
+                            "realtime",
+                            match[:CHILD_TABLE_NUM]
+                          ]
+                        end
```
Fryguy, line 37: can this be collapsed to:

```ruby
primary_condition = 
  if inherit_from.include?("rollup")
    "capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?"
  else
    "capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?"
  end

conditions = [
  primary_condition,
  "realtime",
  match[:CHILD_TABLE_NUM]
]
```

lib/migration_helper.rb
```diff
@@ -54,57 +54,20 @@ def estimate_batch_complete(start_time, progress)
   # Triggers
   #
 
+  # Keeping this helper to supress the SQL body from the `say_with_time` output
```
Fryguy, line 57: Typo `supress` -> `suppress`

Fryguy:
I love this PR...let's get this over the finish line.

Fryguy:
Is it possible for us to remove the miq-specific stuff from the schema dumper?  So, for example, we have `add_miq_metric_table_inheritance`, but nothing inthat method seems miq specific, so can we name is `add_table_inheritance`?  The only thing seems to be the name of the actual constraint, but generically that can be a `constraint_name` parameter that defaults to `#{table}_inheritance_check` if not passed (or we just don't allow passing it at all going the convention-over-configuration route)

> Do we just want to leverage gems for things (like scenic for VIEWS) instead, or are the MIQ specifics enough to warrant me rolling it ourselves?

As I'm reading this PR I'm seeing a gem to do this generically.  So I'd expect we use gems for anything that exists like views, and write our own gem for other things (like maybe triggers if there's no gem for it).  Even so, I like what's here now, so I'd prefer merging what we have, let it get some battle-testing in, then extract / change to gems after.

Fryguy:
> Testing

Can you also verify running db:migrate to a particular `VERSION=` dumps the schema appropriately at that point in time?

lib/manageiq/schema/miq_schema_dumper.rb
```diff
@@ -0,0 +1,110 @@
...
+      def tables(stream)
+        super
+        miq_metric_table_contraints(stream)
+        miq_metric_views(stream)
+        triggers(stream)
+      end
+
+      def table(table, stream)
+        super
+        track_miq_metric_table_inheritence(table)
```
chessbyte, line 15: typo: `inheritence` should be `inheritance`

lib/manageiq/schema/miq_command_recorder.rb
NickLaMuro: 100% agree with that.  Good point.  👍

Fryguy:
I actually want both since they are 2 different things.  This PR, for me, is less about performance, and more about getting back to "normal Rails" where you just use the dumped schema directly.  That you get a performance boost in using normal Rails is more of a side effect.  Completely separately is the performance aspect of a migration from scratch and the collapsed initial migration, which I still think is good to do also.

My intent in https://github.com/ManageIQ/manageiq-schema/issues/502#issuecomment-678280535 was that the collapsed initial migration will, I think, actually remove some of the work in this PR.  That is, the migration I will collapse into is 20190726204302_remove_local_default_embedded_ansible_repos.rb, which comes after 20190122213042_use_views_for_metrics.rb.  So, I was thinking this might then remove or perhaps simplify some of the metrics table stuff, so it would be easier to do this PR after the collapse.

NickLaMuro:
> > Testing
> 
> Can you also verify running db:migrate to a particular `VERSION=` dumps the schema appropriately at that point in time?

@Fryguy So I ran the following:

```console
VERSION=20190726204302 RAILS_ENV=development rake db:environment:set evm:db:destroy db:migrate
```

And then compared it with your collapsed migration.  They were pretty close minus a few things you must have changed manually.

<details>
<summary><code>db/schema.rb</code> vs <code>20190726204302_collapsed_initial_migration.rb</code> diff</summary>

```diff
--- db/schema.rb	2020-09-02 18:06:26.000000000 -0500
+++ json_migration.rb	2020-09-02 18:12:44.000000000 -0500
@@ -1,18 +1,3 @@
-# This file is auto-generated from the current state of the database. Instead
-# of editing this file, please use the migrations feature of Active Record to
-# incrementally modify your database, and then regenerate this schema definition.
-#
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
-#
-# It's strongly recommended that you check this file into your version control system.
-
-ActiveRecord::Schema.define(version: 2019_07_26_204302) do
-
-  # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_table "accounts", force: :cascade do |t|
@@ -1203,6 +1188,7 @@
     t.index ["resource_id", "resource_type"], name: "index_conversion_hosts_on_resource_id_and_resource_type"
     t.index ["resource_type", "resource_id"], name: "index_conversion_hosts_on_resource_type_and_resource_id"
   end
+  change_column_comment "conversion_hosts", "id", "The internal database ID."
 
   create_table "custom_attributes", force: :cascade do |t|
     t.string "section"
@@ -5972,6 +5958,7 @@
     t.string "zone", comment: "The zone in which the task should be executed, or any zone if null."
     t.index ["miq_server_id"], name: "index_miq_tasks_on_miq_server_id"
   end
+  change_column_comment "miq_tasks", "id", "[builtin] The internal record ID. This is the primary key."
 
   create_table "miq_user_roles", force: :cascade do |t|
     t.string "name"
@@ -7503,6 +7490,7 @@
     t.index ["type"], name: "index_vms_on_type"
     t.index ["uid_ems"], name: "index_vms_on_vmm_uuid"
   end
+  change_column_comment "vms", "id", "The internal database ID."
 
   create_table "volumes", force: :cascade do |t|
     t.string "name"
@@ -7579,7 +7567,7 @@
 
   create_miq_metric_view "metric_rollups"
 
-  add_trigger "insteadof", "metrics", "metrics_partition", <<-SQL
+  add_trigger "insteadof", "metrics", "metrics_partition", <<~SQL
           CASE EXTRACT(HOUR FROM NEW.timestamp)
             WHEN 0 THEN
               INSERT INTO metrics_00 VALUES (NEW.*);
@@ -7633,7 +7621,7 @@
           RETURN NEW;
   SQL
 
-  add_trigger "insteadof", "metric_rollups", "metric_rollups_partition", <<-SQL
+  add_trigger "insteadof", "metric_rollups", "metric_rollups_partition", <<~SQL
           CASE EXTRACT(MONTH FROM NEW.timestamp)
             WHEN 1 THEN
               INSERT INTO metric_rollups_01 VALUES (NEW.*);
@@ -7662,5 +7650,3 @@
           END CASE;
           RETURN NEW;
   SQL
-
-end
```

</details>

_Note:  I did delete some lines (the `def down` contents for example) from the collapsed migration file I had locally and unindented it as well so it would `diff -u` nicely, but other than that, I didn't modify anything in the schema.

Fryguy:
Yeah so it turns out Rails schema dumper does dump comments on the id column.  We happened to have a spec for that which is what caught it, so I added it manually.

NickLaMuro:
> Yeah so it turns out Rails schema dumper does dump comments on the id column.

`s/does dump/does drop/` ?

Is this something that I should override as part of #504 ?

kbrock:
> the collapsed initial migration will, I think, actually remove some of the work in this PR
> -- Fryguy

good catch. removing the old metrics migration will buy us a lot

> And then compared it with your collapsed migration. They were pretty close minus a few things you must have changed manually.
> -- NickLaMuro 

that is one beautiful diff

this is a very exciting pr

Fryguy:
Ugh..typing fail  ... `s/does dump/doesn't dump/` (edited original comment)

NickLaMuro:
@Fryguy Sorry for taking a while to fix, but addressed the above with the most most recent commit.  Ran the QA steps again with that code in place to confirm that `db:schema:load` re-adds the comment:

```ruby
$ bin/rails c
** ManageIQ master, codename: Kasparov
Loading development environment (Rails 5.2.4.3)
irb(main):001:0> Vm.columns.detect {|c| c.name == ActiveRecord::Base.connection.primary_key("vms")}.comment
PostgreSQLAdapter#log_after_checkout, connection_pool: size: 5, connections: 1, in use: 1, waiting_in_queue: 0
#=> "The internal database ID."
irb(main):002:0> MiqTask.columns.detect {|c| c.name == ActiveRecord::Base.connection.primary_key("vms")}.comment
#=> "[builtin] The internal record ID. This is the primary key."
irb(main):003:0>
```

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -0,0 +1,114 @@
...
+        super
+        miq_metric_table_contraints(stream)
+        miq_metric_views(stream)
+        triggers(stream)
+      end
+
+      def table(table, stream)
+        super
+        add_id_column_comment(table, stream)
+        track_miq_metric_table_inheritance(table)
```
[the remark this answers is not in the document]
kbrock: a little surprised that you only needed to add comments to the `id`.
If rails already takes care of comments for all columns but the `id`, does it make sense to just drop our comments on the `id` and just use what is built into rails?

Sometimes nice to just go with the flow of rails

kbrock (commented):
As I think I said before, this is one great PR

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+
+      def add_id_column_comment(table, stream)
+        pk    = @connection.primary_key(table)
```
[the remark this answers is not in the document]
NickLaMuro: Table is just a table name `String` here, not some kind of object.

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -0,0 +1,114 @@
...
+        pkcol = @connection.columns(table).detect { |c| c.name == pk }
+
+        return unless (comment = pkcol.comment)
```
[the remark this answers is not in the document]
NickLaMuro: I can move the assignment up a line, since it isn't a big deal to me.

lib/manageiq/schema/schema_dumper.rb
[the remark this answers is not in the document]
NickLaMuro: So, I think the reason `:id` isn't include because also the "table description" is included as part of the `create_table` method arguments _**AND**_ that method signature is usually what defines our `:id` columns.

For example, for `conversion_hosts`:

```ruby
  create_table "conversion_hosts", comment: "Conversion Hosts", force: :cascade do |t|
    t.string "name", comment: "A symbolic name for the conversion host."
    t.string "address", comment: "The IP address for the conversion host. If present, must be one of the associated resource's IP addresses."
    
    # ...

  end

  change_column_comment "conversion_hosts", "id", "The internal database ID."
```

The `comment:` that is there is for the table itself, and the `id` column is just inferred by the table definition.

lib/manageiq/schema/schema_dumper.rb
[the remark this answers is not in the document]
NickLaMuro: This is done now.

PR #505, opened 2020-08-26
Added schema changes for Autosde provider

gregoryb-at-ibm:
# Added StorageResource and related tables

New tables

### Storage Resource (Pool)
Represents an entity that can provide block volumes. Most often a pool, a pool-replication-group, or an io-group.
Holds general information about the pool such as name, free space, etc..

Example
```ruby
<StorageResource id: 1
  name: "9.151.159.178:AutoSDE_Pool_CI"
  ems_id: 2
  ems_ref: "fad8a091-38c5-4320-a575-099e3125a9bd"
  uuid: "fad8a091-38c5-4320-a575-099e3125a9bd"
  logical_free: 1095216660480
  logical_total: 1099511627776
  pool_name: "AutoSDE_Pool_CI"
  storage_system_id: 1
  created_at: "2020-09-09 12:47:43"
  updated_at: "2020-09-09 12:47:43">
```

### Physical Storage Family
Represents a type system which the provider supports.
EMSs that want to allow users to attach a storage system to MIQ will need to build a list of storage systems supported by their provider. 
The storage_system_families are supposed to be scanned from the provider. It's up to the EMS and the provider to determine what to put in these records based on how they want to differentiate different storage systems.
So one provider could have a list like ["SVC", "V7000", "Netapp"], and that could be enough for them. So they would have 3 StorageSystemFamily DB records.
Another provider could need to use advanced functions of SVC which would require to differentiate between different SVC versions, their storage_system_families could look like ["svc_version_lte_x", "svc_version_gt_x", "V8K"]. So they would have 4 StorageSystemFamily DB records. 
Note that the lists don't adhere to a unified naming convention because they use conventions from the EMS's respective providers.

Storage families are used when creating a PhysicalStorage. The user needs to pick one from a drop-down, since the provider needs to know what kind of a system it is in order to connect to it.
In the future we want to use them to determine what features the system supports, and perhaps add uniformity across EMSs.

Example
```ruby
<PhysicalStorageFamily id: 1
  name: "svc"
  version: "1.1"
  ems_id: 2
  ems_ref: "06fd344c-64a1-4b31-a502-a7c5aa01b8ab"
  created_at: "2020-09-09 12:47:43"
  updated_at: "2020-09-09 12:47:43">
```

### Storage Service
This is an abstraction layer for volume creation and allocation.

A storage service represents a type of volume that end-user will consume.
Each storage service is connected to a set of suitable storage resources which fulfill the service criteria.

Storage services will be displayed in the self-service catalog for users requesting new volumes.
Provider may adjust volumes based on changes to their service.

Different storage services can be defined for different user needs.
For example DB storage service. Real-time storage service, etc..
A volume is then created on one of the pools attached to the service.

Example
```ruby
<StorageService id: 1
  name: "9.151.159.178:AutoSDE_Pool_CI"
  description: "auto_created_service"
  uuid: "6c14b03e-572e-4528-a19f-f92e6d91df12"
  ems_id: 2
  ems_ref: "6c14b03e-572e-4528-a19f-f92e6d91df12"
  created_at: "2020-09-09 12:47:43"
  updated_at: "2020-09-09 12:47:43">
```

### Service Resource Attachment
This is the M2M table that attaches storage-services to storage-resources.

### Rough Diagram

image

### DIFF
--- a/db/migrate/20200519141015_create_storage_resources.rb
+++ b/db/migrate/20200519141015_create_storage_resources.rb
@@ -1,0 +1,14 @@
+class CreateStorageResources < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_resources do |t|
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uuid
+      t.bigint :logical_free
+      t.bigint :logical_total
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20200519141015_create_storage_resources.rb
+++ b/db/migrate/20200519141015_create_storage_resources.rb
@@ -4,10 +4,12 @@ class CreateStorageResources < ActiveRecord::Migration[5.2]
       t.string :name
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
-      t.string :uuid
+      t.string :uid_ems
       t.bigint :logical_free
       t.bigint :logical_total
       t.references :physical_storage, :type => :bigint, :index => true
+      t.string :type
+
       t.timestamps
     end
   end
--- a/db/migrate/20200521150213_create_storage_services.rb
+++ b/db/migrate/20200521150213_create_storage_services.rb
@@ -1,0 +1,14 @@
+class CreateStorageServices < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_services do |t|
+      t.string :name
+      t.string :description
+      t.string :uuid
+      t.bigint :version
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20200521150213_create_storage_services.rb
+++ b/db/migrate/20200521150213_create_storage_services.rb
@@ -3,10 +3,11 @@ class CreateStorageServices < ActiveRecord::Migration[5.2]
     create_table :storage_services do |t|
       t.string :name
       t.string :description
-      t.string :uuid
       t.bigint :version
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
+      t.string :uid_ems
+      t.string :type
 
       t.timestamps
     end
--- a/db/migrate/20200524074928_create_storage_service_resource_attachments.rb
+++ b/db/migrate/20200524074928_create_storage_service_resource_attachments.rb
@@ -1,0 +1,12 @@
+class CreateStorageServiceResourceAttachments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_service_resource_attachments do |t|
+      t.references :storage_service, :type => :bigint, :index => {:name => "index_storage_svc_resource_attachments_on_storage_service_id"}
+      t.references :storage_resource, :type => :bigint, :index => {:name => "index_storage_svc_resource_attachments_on_storage_resource_id"}
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20200806112501_expand_cloud_volume.rb
+++ b/db/migrate/20200806112501_expand_cloud_volume.rb
@@ -1,0 +1,6 @@
+class ExpandCloudVolume < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :cloud_volumes, :storage_resource, :type => :bigint, :index => true
+    add_reference :cloud_volumes, :storage_service, :type => :bigint, :index => true
+  end
+end
--- a/db/migrate/20200922180455_add_uuid_and_family_to_physical_storages.rb
+++ b/db/migrate/20200922180455_add_uuid_and_family_to_physical_storages.rb
@@ -1,0 +1,6 @@
+class AddUuidAndFamilyToPhysicalStorages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :physical_storages, :uuid, :string
+    add_reference :physical_storages, :physical_storage_family, :type => :bigint, :index => true
+  end
+end
--- a/db/migrate/20200922180455_add_uuid_and_family_to_physical_storages.rb
+++ b/db/migrate/20200922180455_add_uuid_and_family_to_physical_storages.rb
@@ -1,6 +1,0 @@
-class AddUuidAndFamilyToPhysicalStorages < ActiveRecord::Migration[5.2]
-  def change
-    add_column :physical_storages, :uuid, :string
-    add_reference :physical_storages, :physical_storage_family, :type => :bigint, :index => true
-  end
-end
--- a/db/migrate/20200608163713_create_physical_storage_families.rb
+++ b/db/migrate/20200608163713_create_physical_storage_families.rb
@@ -1,0 +1,11 @@
+class CreatePhysicalStorageFamilies < ActiveRecord::Migration[5.2]
+  def change
+    create_table :physical_storage_families do |t|
+      t.string :name
+      t.string :version
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20200922180455_add_family_to_physical_storages.rb
+++ b/db/migrate/20200922180455_add_family_to_physical_storages.rb
@@ -1,0 +1,5 @@
+class AddFamilyToPhysicalStorages < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :physical_storages, :physical_storage_family, :type => :bigint, :index => true
+  end
+end


db/migrate/20200806112501_expand_cloud_volume.rb
```diff
@@ -0,0 +1,11 @@
+#  https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/
```
Fryguy, line 1: @gregoryb-at-ibm Can you remove this comment and the blank line after?

Fryguy:
Also can deal with the issues raised by the bot? https://github.com/ManageIQ/manageiq-schema/pull/505#issuecomment-680887216

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,16 @@
+class CreateStorageResources < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_resources do |t|
+      t.string :name
+      t.bigint :ems_id
```
agrare, line 5: I'm concerned that these don't have indexes on any of these foreign keys, `:references` is a nice way to do this in one shot e.g. `t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system`

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :ems_ref
+      t.string :uuid
+      t.bigint :logical_free
+      t.bigint :logical_total
+      t.string :pool_name
+      t.bigint :storage_system_id
```
agrare, line 11: And here `t.references :storage_system, :type => :bigint, :index => true`

db/migrate/20200519141015_create_storage_resources.rb
gregoryb-at-ibm: @agrare , t.references is a front to https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference? it's a bit odd, because the documentation seems to say that it's `:ref_name`, not `:references` ? Is the documentation incorrect? Or am I looking in the wrong place?

While I'm doing the changes, could you explain why not use `:foreign_key => true`  instead of `:index => true`? Why don't we want DB-level foreign keys? Because we might want to convert the relations to polymorphic?

Also, doesn't :type default to bignit?

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :storage_resources do |t|
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uuid
+      t.bigint :logical_free
+      t.bigint :logical_total
+      t.string :pool_name
+      t.references :storage_system, :type => :bigint, :index => true, :references => :storage_system
```
agrare, line 11: you only need the `:references => :storage_system` if the name of the model doesn't match the name of the column, e.g. the column name is ems_id but the model name is `ExtManagementSystem`, for anything else it infers the reference from the column name so these can just be `t.references :storage_system, :type => :bigint, :index => true`

db/migrate/20200519141015_create_storage_resources.rb
agrare: >  it's a bit odd, because the documentation seems to say that it's :ref_name, not :references ? Is the documentation incorrect? Or am I looking in the wrong place?

No that's correct but `ref_name` is just the name of the parameter, `add_reference(table_name, ref_name, **options) public` so in this case `:ems` is the `ref_name` parameter and the rest of the options hash goes to the `options` parameter

agrare: > Also, doesn't :type default to bignit?

In rails 5.2 it does, but prior to that it defaulted to `:integer`.  Take a look at https://apidock.com/rails/v5.2.3/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference versus https://apidock.com/rails/v5.1.7/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference and it looks like you generated this on 5.1 so it would have defaulted to `:integer`

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,15 @@
+class CreateStorageResources < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_resources do |t|
```
agrare, line 3: `StorageResource` is a little generic, if this is a storage pool I think that would be a more intuitive name.  We also have `PhysicalDisks` which could belong to this new model.

db/migrate/20200524080344_create_storage_systems.rb
```diff
@@ -0,0 +1,17 @@
+class CreateStorageSystems < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_systems do |t|
```
agrare, line 3: We have an existing `PhysicalStorage` model, `StorageSystem` sounds like the same concept so I don't want to introduce a duplicate model.  Can you add any attributes that you need which are missing to the physical_storages table instead?

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uuid
+      t.bigint :logical_free
+      t.bigint :logical_total
+      t.string :pool_name
```
agrare, line 10: If this is a storage pool, and it has a name, what is the `pool_name` then?

db/migrate/20200524080344_create_storage_systems.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :management_ip
+      t.string :name
+      t.string :secondary_ip
+      t.string :storage_array
+      t.string :storage_family
+      t.references :storage_system_type, :type => :bigint, :index => true
```
Fryguy, line 9: ending a column in `_type` is going to be confusing because `_type` suffix is reserved for polymorphic relationships.  We may want to rename that table to not end in `_types`

Additionally, the storage_systems_type table is created after this point...I expect that other migration to be first.

db/migrate/20200608163713_create_storage_system_types.rb
```diff
@@ -0,0 +1,11 @@
+class CreateStorageSystemTypes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_system_types do |t|
```
Fryguy, line 3: See https://github.com/ManageIQ/manageiq-schema/pull/505/files#r482449221 - I'm concerned this name might create confusion, and I wonder if we can come up with a different name.  @agrare Do we have a precedent for table names ending in `_types`?

db/migrate/20200524074928_create_service_resource_attachments.rb
```diff
@@ -0,0 +1,13 @@
+class CreateServiceResourceAttachments < ActiveRecord::Migration[5.1]
+  def change
+    create_table :service_resource_attachments do |t|
+      t.references :storage_service, :type => :bigint, :index => true
+      t.references :storage_resource, :type => :bigint, :index => true
+      t.boolean :compliant
```
Fryguy, line 6: What is `compliant`?  I'm concerned this will be confused with our existing compliances modeling.  Perhaps we should use `ems_compliant` or `native_compliant` instead to differentiate?  @agrare Thoughts?

db/migrate/20200806112501_expand_cloud_volume.rb
```diff
@@ -0,0 +1,8 @@
+class ExpandCloudVolume < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :cloud_volumes, :storage_resource, :type => :bigint, :index => true
+    add_reference :cloud_volumes, :storage_service, :type => :bigint, :index => true
+    add_column :cloud_volumes, :compliant, :boolean
```
Fryguy, line 5: Similar comment here re: compliant column.

db/migrate/20200806112501_expand_cloud_volume.rb
```diff
@@ -0,0 +1,8 @@
...
+    add_column :cloud_volumes, :volume_source, :string
```
Fryguy, line 6: What does this column mean?

Fryguy:
@agrare @gregoryb-at-ibm I think `t.references` defaults to `index => true`, but I could be wrong... Can we verify, and if it is the default, then we should leave it out.

db/migrate/20200608163713_create_storage_system_types.rb
agrare: @Fryguy we have a few `cloud_volume_types`, `miq_worker_types`, `notification_types`, and `pxe_image_types`

db/migrate/20200524074928_create_service_resource_attachments.rb
agrare: Who is evaluating this compliance?  Is it the native system or MIQ?

If it is the native system then yeah ems_compliant would be less confusing, if it is an MIQ plugin then we should be using the existing compliance infrastructure not introduce new columns

db/migrate/20200608163713_create_storage_system_types.rb
Fryguy: Ok, then I guess it's ok. :+1:

agrare:
@Fryguy it was changed to default to `true` in rails 5.1, that is recent enough that being explicit might not be a bad thing?

db/migrate/20200608163713_create_storage_system_types.rb
agrare: That said you're right another model referencing this does look like one half of a polymorphic association so if we can come up with a better name I'm all for it

db/migrate/20200524074928_create_service_resource_attachments.rb
ereztheodorou: The compliance is evaluated by the Autosde management system. When False, this is to indicate that a specific resource that is already attached to a service no longer matches the capabilities defined by the service. So yes, ems_compliant is probably better

db/migrate/20200806112501_expand_cloud_volume.rb
ereztheodorou: As we are re-using the cloud volume to represent the volumes managed by autosde, this column will indicate what is the origin of this volume, mainly to be easily displayed as info in the UI list of volumes.

db/migrate/20200524080344_create_storage_systems.rb
ereztheodorou: The motivation for creating a new table was that many of the relations and in the physical_storage are not relevant to other storages (rack, chassis, canisters, computer_system,  ...). The propose a design of a storage_system that is general / minimalistic table that includes only the common fields needed for storage system and relations to only general objects common to all storage systems (i.e. Volumes, Ports) . Each type of storage may inherit from this table and add it's own personal relations and fieles. . This way each field and relation in the table is populated with relevant info and we wont have many empty non relevant fields or relation for each storage type.

db/migrate/20200806112501_expand_cloud_volume.rb
agrare: What would some of the possible values be?  E.g. does this indicate that it came from autosde (which we could get from the ext_management_system link) or something else?

db/migrate/20200524080344_create_storage_systems.rb
agrare: Are these `storage_systems` going to be representing actual physical storage systems with hard drives etc...?   The management_ip makes me think that they are.

If the issue is that autosde won't be populating rack/chassis and lenovo won't be populating storage-pools that's perfectly fine, no provider populates every association on the models that it collects and we can always not show relationships that aren't populated.

I think this will be more confusing when we start adding other standalone storage managers, e.g. if we added a NetApp provider that does have physical_storages but also storage-pools, would it populate both physical_storages and storage_systems?

If it would help we could move the associations to e.g. `physical_racks` and `physical_chassis` into a `PhysicalInfraManager::PhysicalStorage` subclass so that your provider's subclass of `PhysicalStorage` wouldn't have them.

agrare: So looking at https://en.wikipedia.org/wiki/IBM_Storwize_family#Storwize_V7000_Gen_3_(or_'Next_Gen') it has drive bays/enclosures (both on physical_storage table) and multiple canisters (a physical_storage has_many canisters)

As for https://en.wikipedia.org/wiki/IBM_SAN_Volume_Controller that almost sounds like its own standalone storage_manager type which aggregates many physical_storages the same way vCenter manages multiple ESXi hosts

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,15 @@
+class CreateStorageResources < ActiveRecord::Migration[5.1]
```
agrare, line 1: Lets update this to 5.2 as that is the current version of rails/active_record that manageiq is using.  I assume you generated this a while ago while we were still on 5.1 but some of the defaults for e.g. references have changed so this could be simplified on the newer version

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,18 @@
+class CreateStorageServices < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_services do |t|
+      t.string :name
+      t.string :description
+      t.bigint :project_id
+      t.bigint :profile_id
```
agrare, line 7: These look like foreign keys to other models, but we don't have project/profile tables.  Are these coming from the native provider?  If so these are badly named

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,18 @@
+class CreateStorageServices < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_services do |t|
```
agrare, line 3: This sounds like it might be closer to a storage profile?  We also have a storage_profiles model (it is pretty sparse right now but could be expanded out if it does match)

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :name
+      t.string :description
+      t.bigint :project_id
+      t.bigint :profile_id
+      t.string :uuid
+      t.bigint :version
+      t.references :parent_service, :type => :bigint, :index => true, :references => :storage_service
+      t.string :capability_values
```
agrare, line 11: What would the value of this look like?  This looks like it might end up being a comma separated mess

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,18 @@
+class CreateStorageServices < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_services do |t|
+      t.string :name
+      t.string :description
+      t.bigint :project_id
+      t.bigint :profile_id
+      t.string :uuid
+      t.bigint :version
+      t.references :parent_service, :type => :bigint, :index => true, :references => :storage_service
```
agrare, line 10: Is there a limitation on how many parents a service can have?  We do have the ancestry gem which helps model tree-like structures in the database

db/migrate/20200524074928_create_service_resource_attachments.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
```
agrare, line 7: You can use join_table notation to create this if it is a simple has and belongs_to many type table, also then you don't need the ems reference since it is going to be references from the services and resources sides

db/migrate/20200524074928_create_service_resource_attachments.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :ems_ref
```
agrare, line 8: Also, an ems_ref on a join table doesn't make sense, are these actually returned by the provider API as its own resource?

db/migrate/20200524080344_create_storage_systems.rb
agrare: Could this just be `:type` and we use STI to represent the different storage types?

db/migrate/20200524080344_create_storage_systems.rb
agrare: Maybe we rename physical_storages to storage_systems since that is a good name for a general storage type, then physical storages can be a subclass of that?

db/migrate/20200524080344_create_storage_systems.rb
```diff
@@ -0,0 +1,17 @@
+class CreateStorageSystems < ActiveRecord::Migration[5.1]
+  def change
+    create_table :storage_systems do |t|
+      t.string :management_ip
+      t.string :name
+      t.string :secondary_ip
```
agrare, line 6: This looks like it should be handled by `has_many :endpoints`, what happens if a system has a third IP addr?

db/migrate/20200608163713_create_storage_system_types.rb
agrare: @gregoryb-at-ibm what would populate this storage_system_types and what are some of the possible values?  Would this be e.g. SVC or V7000?  In that case this can probably just be an STI class.  Or is it something collected from the provider API that is very dynamic and has to be populated by refresh?

agrare:
@gregoryb-at-ibm could we get an example record for each of these new tables so we can get a feel for what the data is going to look like?

db/migrate/20200608163713_create_storage_system_types.rb
gregoryb-at-ibm: @agrare , The storage_system_types are supposed to be scanned from whatever backend the EMS is managing. It's up to the EMS and the backend to determine what to put there, based on how they want to differentiate different storage systems.
So one backend could have a list like ["SVC", "V7000", "Netapp"], and that can be enough for them.
Another backend could need to use advanced functions of SVC which would require to differentiate between different SVC versions, their storage_system_types could look like ["svc_version_lte_x", "svc_version_gt_x", "V8K"]. 
Note that the lists don't adhere to the same naming conventions because they use conventions from the EMS's backend.

They are used when creating a storage system. The user needs to pick one from a drop-down, since the EMS backend needs to know what kind of a system it is in order to connect to it. 
They are also presented to the user when viewing systems.
In the future we want to use them to determine what features the system supports, and perhaps add uniformity across EMSs.

db/migrate/20200524074928_create_service_resource_attachments.rb
gregoryb-at-ibm: Yea, the attachments between the services and the resources are synced from the provider. 
The idea is that the provider can adjust attachments automatically based on rules.
@ereztheodorou, could you elaborate please?

db/migrate/20200524080344_create_storage_systems.rb
gregoryb-at-ibm: The fields are here because we are syncing them from the provider. 
But I'm not sure that we actually need them. I don't think we intend to access the systems directly from MIQ.
@ereztheodorou, what do you think?

gregoryb-at-ibm:
@agrare , I've added an example of 
 
> @gregoryb-at-ibm could we get an example record for each of these new tables so we can get a feel for what the data is going to look like?

gregoryb-at-ibm:
Sorry closed by mistake

db/migrate/20200524080344_create_storage_systems.rb
gregoryb-at-ibm: @agrare, We'll move everything into PhysicalStorage.

db/migrate/20200521150213_create_storage_services.rb
ereztheodorou: This is a key value dictionary. This comes from the ems (autosde provider) and is an abstraction/standarization of capabilities across different kinds of storages. In our current first release there is no practical use for it so we will drop it for now. (later releases will relay on these capabilities for provisioning the correct volume)

db/migrate/20200524080344_create_storage_systems.rb
ereztheodorou: MIQ will not currently use them as they are managed by the provider. They are here for information only. in the future we might want to be able to update these IPs in the provider though MIQ interface (via the REST api) We can drop it for now.

db/migrate/20200519141015_create_storage_resources.rb
ereztheodorou: This was added for convenience as the resource name can possibly be different then the pool name. (but a resource may not necessarily be a pool.) we'll remove the pool_name from the storage_resource.

db/migrate/20200519141015_create_storage_resources.rb
ereztheodorou: StorageResource is an abstract object that may represent a storage pool for certain storages but not necessarily. (It could possibly represent backends, paired pools (for replication) , containers etc)

db/migrate/20200524074928_create_service_resource_attachments.rb
ereztheodorou: we will remove the compliance field for this release

db/migrate/20200524074928_create_service_resource_attachments.rb
ereztheodorou: we originally have a compliant field that will be synced from the provider autosde. We will remove the compliance field as it will not be used in the initial release. We will add it in future release so it may be better to keep this table without the automatic join for easy sync with the ems

db/migrate/20200524074928_create_service_resource_attachments.rb
ereztheodorou: I have elaborated in the prev comment

db/migrate/20200608163713_create_storage_system_types.rb
gregoryb-at-ibm: @agrare , Can this conversation be resolved? I've renamed storage_system_types to storage_system_families so it doesn't look like a polymorphic relationship.

db/migrate/20200521150213_create_storage_services.rb
gregoryb-at-ibm: @agrare , @Fryguy, what do you think? StorageService is also something that's supposed to provision a volume on demand. Would it make sense for there to be `StorageProfile.create_cloud_volume()` method? Or is it better if we have a separate `StorageService` that provides `StorageService.create_cloud_volume()`?

db/migrate/20200524074928_create_service_resource_attachments.rb
agrare: If this is actually something that is actually representing a provider backed object then I'm good with this not being a join table (which typically just represent connections between two other records)

db/migrate/20200608163713_create_storage_system_types.rb
agrare: :+1:

Fryguy:
@gregoryb-at-ibm Thanks for updating the OP with examples, but it seems there are still some `nil` values in the columns we don't understand like capability_values, secondary_ip, and storage_array.  Can you update those?

gregoryb-at-ibm:
> 
> 
> @gregoryb-at-ibm Thanks for updating the OP with examples, but it seems there are still some `nil` values in the columns we don't understand like capability_values, secondary_ip, and storage_array. Can you update those?

@Fryguy , Thanks for pointing this out. We have made some changes since I added the examples. We have agreed to deprecate some fields and the entire StorageSystem model. I've now updated the examples accordingly.
I've also updated the diagram

db/migrate/20200521150213_create_storage_services.rb
agrare: If it is something that's going to be shown in the service catalog I'm okay with not reusing StorageProfile.  But in that case should this be an actual Miq Service (aka what we show in our service catalog)?  cc @Fryguy

db/migrate/20200519141015_create_storage_resources.rb
agrare: :+1: okay.  A little worried that this being so abstract means it isn't actually representing anything specific but I'm okay seeing where this goes  @Fryguy ?

Fryguy: Yeah, I think I'm ok with this for now.  When we get further along and populate it with more than pools, we could always refactor it later.

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,14 @@
+class CreateStorageServices < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_services do |t|
+      t.string :name
+      t.string :description
+      t.string :uuid
+      t.bigint :version
```
Fryguy, line 7: Is this always going to be an integer version or should we hedge and use a string field (I'm thinking weird versions like `1.0a-beta2`)

Fryguy:
>  We have agreed to deprecate some fields and the entire StorageSystem model.

@gregoryb-at-ibm I see you had updated the OP with examples, but there are some fields in the PR that aren't present there, but I think that you were going to remove (including the storage_systems table).  Did you forget to push a commit or is that still pending?

db/migrate/20200521150213_create_storage_services.rb
ereztheodorou: Yes. This will be a incremental integer value to keep track of changes to a service. The idea is that capabilities in a storage service may be changed in the course of time and once changed a new service with a higher value will be created. This way existing resources will be associated with an older version of the service, and any new resources will be associated with the new service with a higher version.

gregoryb-at-ibm:
> @gregoryb-at-ibm I see you had updated the OP with examples, but there are some fields in the PR that aren't present there, but I think that you were going to remove (including the storage_systems table). Did you forget to push a commit or is that still pending?

@Fryguy, the refactoring to deprecate StorageSystem was taking us a long time. I'd updated the fields in the OP first to reflect what we were doing. The refactoring is now done and I've pushed new code with the offending fields removed.

db/migrate/20200521150213_create_storage_services.rb
gregoryb-at-ibm: I think this would be a backend for an actual MIQ service. 
The idea is that we would sync these StorageServices from the provider during inventory scan. 
When a user wants a new volume, the miq-service-catalog would have a dropdown where they select the StorageService.
A request would then be send to the provider, asking it to create a new pool on this service. 

Also, since it is collected as part of an inventory, it needs ems_id.

db/migrate/20200524074928_create_service_resource_attachments.rb
```diff
@@ -0,0 +1,12 @@
+class CreateServiceResourceAttachments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_resource_attachments do |t|
```
Fryguy, line 3: I'm concerned the name here will be confused with service resources (which is an actual thing as services table entries have resources).  I think `storage_service_resource_attachments` would be clearer.

Fryguy (commented):
One comment on the table name, but the rest looks good to me.

db/migrate/20200524074928_create_service_resource_attachments.rb
gregoryb-at-ibm: @Fryguy 
Renaming the table to storage_service_resource_attachments causes the index-name to be too long.
```ArgumentError: Index name 'index_storage_service_resource_attachments_on_storage_service_id' on table 'storage_service_resource_attachments' is too long; the limit is 63 characters```

Maybe we could use shorthand for naming the table `strg_service_resource_attachment`, but rename the ApplicationRecord to `StorageServiceResourceAttachment` ?

Comment (agrare) [reply db/migrate/20200524074928_create_service_resource_attachments.rb]: You can just override the default index name with something shorter, no need to rename the table

File: db/migrate/20200524074928_create_storage_service_resource_attachments.rb [hunk 23]
```diff
@@ -0,0 +1,12 @@
+class CreateStorageServiceResourceAttachments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_service_resource_attachments do |t|
+      t.references :storage_service, :type => :bigint, :index => true
```
agrare, line 4: E.g. `t.references :storage_service, :type => :bigint, :index => {:name => "shorter_name_for_your_index"}`

Fryguy: Specifically, `:name => index_storage_svc_resource_attachments_on_storage_service_id` will work

db/migrate/20200524074928_create_service_resource_attachments.rb
gregoryb-at-ibm: If the table-name is so long that auto generated indexes fail, won't it be the case with every index that involves this table?
In the future if anyone wants to add another index on this table, they'll unexpectedly get an error like I have and they'll have to specify an explicit index name.
Also, if any other table should want to reference this one, the index-name would also be unexpectedly too long. 

So my idea was that shortening the table name would make indexes involving this table "just work" and spare future developers from these surprising errors. I thought this sort of consideration was the reason for having ext_management_system in stead of external_management_system.

db/migrate/20200922180455_add_uuid_and_family_to_physical_storages.rb
```diff
@@ -0,0 +1,6 @@
+class AddUuidAndFamilyToPhysicalStorages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :physical_storages, :uuid, :string
```
agrare, line 3: `PhysicalStorage` already has `:ems_ref` and a `uid_ems` columns

The `ems_ref` is typically the `id` of the object or some equivalent, and the `uid_ems` is typically a UUID if the provider has one, otherwise equal to the `ems_ref`

Unless you need both of the existing ems_ref and uid_ems columns for other properties, can you store the uuid in the uid_ems column?

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :storage_resources do |t|
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uuid
+      t.bigint :logical_free
+      t.bigint :logical_total
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.timestamps
+    end
```
agrare, line 12: Can you add a `:type` string column for STI so we can subclass this model in the provider?

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+
+      t.timestamps
+    end
```
agrare, line 12: Same with this model, STI will be helpful

db/migrate/20200519141015_create_storage_resources.rb
Fryguy: Oh good call, I completely forgot about that. :+1:

db/migrate/20200521150213_create_storage_services.rb
Fryguy: Do we need the type column on physical_storage_families ?

db/migrate/20200521150213_create_storage_services.rb
```diff
@@ -0,0 +1,14 @@
+class CreateStorageServices < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_services do |t|
+      t.string :name
+      t.string :description
+      t.string :uuid
```
Fryguy, line 6: @agrare Based on your comment on the pyhsical_storages table, should this, then, be uid_ems instead of uuid?

db/migrate/20200519141015_create_storage_resources.rb
```diff
@@ -0,0 +1,14 @@
+class CreateStorageResources < ActiveRecord::Migration[5.2]
+  def change
+    create_table :storage_resources do |t|
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uuid
```
Fryguy, line 7: @agrare Same here, should this be uid_ems instead of uuid?

db/migrate/20200521150213_create_storage_services.rb
agrare: Yeah I considered that and I don't think so because this is basically a list of possible types in each record.

If they want to do any `supports?(feature)` on the `physical_storage_family` though then it definitely should.

db/migrate/20200521150213_create_storage_services.rb
agrare: :+1: yeah

db/migrate/20200521150213_create_storage_services.rb
gregoryb-at-ibm: @agrare , we don't intend to have a  list of options in each record. Each record would represent a single family in a single provider. If a provider's families are ['SVC', 'Netapp', 'V7000'], it would have three records. One for each family.

That said I don't think we'll be doing STI with physical_storage_families at this point.

agrare: Yes exactly @gregoryb-at-ibm each one of these records would be the equivalent to what and STI subclass would give us.

I think we still want a `:type` column on this `storage_services` model though right?

gregoryb-at-ibm: Oops, must have missed it.

agrare (approved):

PR #507, opened 2020-09-02
Fix RSpec warning on raise_error

Fryguy:
"expect to raise_error" should not be used without a parameter as any
error, including bugs will end up passing the test. This particular
instance can just be checked with a normal respond_to, so we use that
instead.


Fryguy:
@agrare @bdunne Please review.

PR #508, opened 2020-09-10
Replace Vim* types in CustomizationSpec#spec

agrare:
The CustomizationSpec#spec is a serialized column which on versions prior to jansa contained a VimHash CustomizationSpecItem object.

https://github.com/ManageIQ/manageiq-providers-vmware/pull/614 collects these with RbVmomi as standard hashes

### DIFF
--- a/db/migrate/20200910145925_remove_vim_types_from_customization_specs.rb
+++ b/db/migrate/20200910145925_remove_vim_types_from_customization_specs.rb
@@ -1,0 +1,40 @@
+class RemoveVimTypesFromCustomizationSpecs < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  class CustomizationSpec < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time("Removing Vim Types from CustomizationSpecs") do
+      base_relation = CustomizationSpec.in_my_region.where("spec LIKE ?", "%hash-with-ivars:VimHash%")
+      say_batch_started(base_relation.size)
+
+      loop do
+        count = base_relation
+                .limit(50_000)
+                .update_all("spec = REGEXP_REPLACE(spec, '!ruby/(string|array|hash-with-ivars):Vim(Hash|String|Array)', '!ruby/\\1:\\2', 'g')")
+        break if count == 0
+
+        say_batch_processed(count)
+      end
+    end
+  end
+
+  def down
+    say_with_time("Restoring Vim Types in CustomizationSpecs") do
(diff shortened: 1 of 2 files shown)


spec/migrations/20200910145925_remove_vim_types_from_customization_specs_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+                    ivars:
+                      :@xsiType: :CustomizationPrefixName
+                      :@vimType:
+                    elements:
+                      base: !ruby/string:VimString
+                        str: custom-hostname
+                        xsiType: :SOAP::SOAPString
+                        vimType:
+                  domain: !ruby/string:VimString
+                    str: cloudforms.lab.eng.rdu2.redhat.com
```
kbrock, line 64: think these won't be hard to tweak to something else to appease rubocop

db/migrate/20200910145925_remove_vim_types_from_customization_specs.rb
```diff
@@ -0,0 +1,40 @@
...
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  class CustomizationSpec < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time("Removing Vim Types from CustomizationSpecs") do
+      base_relation = CustomizationSpec.in_my_region.where("spec LIKE ?", "%hash-with-ivars:VimHash%")
```
kbrock, line 11: note: this ensures records already processed won't get processed again

db/migrate/20200910145925_remove_vim_types_from_customization_specs.rb
```diff
@@ -0,0 +1,40 @@
...
+      say_batch_started(base_relation.size)
+
+      loop do
+        count = base_relation
+                .limit(50_000)
+                .update_all("spec = REGEXP_REPLACE(spec, '!ruby/(string|array|hash-with-ivars):Vim(Hash|String|Array)', '!ruby/\\1:\\2', 'g')")
+        break if count == 0
```
kbrock, line 18: note: this bails when there is nothing left to update

db/migrate/20200910145925_remove_vim_types_from_customization_specs.rb
kbrock: I notice that the regular expression in the `update_all` matches more cases than this like in the limiter / `base_relation`.

Is it possible to only have something like a `string:VimString` and it will not trigger the base relation but should be updated by the `update_all`?

kbrock (commented):
Just the one comment.
(I'm not a merger so feel free to ignore)

I'm very excited about these changes

PR #509, opened 2020-09-10
Data migration for control explorer split

h-kataria:
Migration changes are to go with changes in https://github.com/ManageIQ/manageiq/pull/20438.
- startpage url changed to `miq_policy_set`(Policy Profiles explorer) as start page instead of `control explorer`(unified version of Control explorer)
- parent feature identifier was changed from `control_explorer` to `control`, we do not need to worry about other child features in migration because those were hidden features previously. If a user could access control explorer they were allowed to perform all actions in that area. Now to be consistent with other screens, made all features visible so RBAC can be applied on the actions individually.

https://github.com/ManageIQ/manageiq-ui-classic/issues/6996

This PR goes with changes in core [PR](https://github.com/ManageIQ/manageiq/pull/20438)
UI changes in [PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/7259)

### DIFF
--- a/db/migrate/20200910163614_adjust_control_explorer_features.rb
+++ b/db/migrate/20200910163614_adjust_control_explorer_features.rb
@@ -1,0 +1,26 @@
+class AdjustControlExplorerFeatures < ActiveRecord::Migration[5.2]
+  class MiqProductFeature < ActiveRecord::Base; end
+
+  FEATURE_MAPPING = {
+    'control_explorer'  => 'control'
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    say_with_time 'Adjusting control_explorer features for the split explorers' do
+      # Direct renaming of features
+      FEATURE_MAPPING.each do |from, to|
+        MiqProductFeature.find_or_create_by!(:identifier => from)&.update!(:identifier => to)
+      end
+    end
+  end
+
+  def down
+    say_with_time 'Adjusting split explorer features back to control explorer' do
+      FEATURE_MAPPING.each do |to, from|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+    end
+  end
+end
--- a/db/migrate/20200910163718_adjust_control_explorer_startpage_entries.rb
+++ b/db/migrate/20200910163718_adjust_control_explorer_startpage_entries.rb
@@ -1,0 +1,25 @@
+class AdjustControlExplorerStartpageEntries < ActiveRecord::Migration[5.2]
+  class User < ActiveRecord::Base
+    serialize :settings, Hash
+  end
+
+  def up
+    say_with_time 'Updating starting page for users who had control explorer set' do
+      User.select(:id, :settings).each do |user|
+        if user.settings&.dig(:display, :startpage) == 'control/explorer'
+          user.update!(:settings => user.settings.deep_merge(:display => {:startpage => 'miq_policy_set/explorer'}))
+        end
+      end
+    end
+  end
+
+  def down
+    say_with_time 'Updating starting page for users who had miq_policy_set back to unified control explorer' do
+      User.select(:id, :settings).each do |user|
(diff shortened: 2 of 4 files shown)


db/migrate/20200910163614_adjust_control_explorer_features.rb
```diff
@@ -0,0 +1,48 @@
...
+    say_with_time 'Adjusting control_explorer features for the split explorers' do
+      new_feature = MiqProductFeature.find_or_create_by!(:identifier => 'control')
+      user_role_ids_with_old_features = MiqRolesFeature.where(:miq_product_feature_id => old_feature_ids).pluck(:miq_user_role_id).uniq
+      user_role_ids_with_old_features.each do |user_role_id|
+        MiqRolesFeature.create!(:miq_product_feature_id => new_feature.id, :miq_user_role_id => user_role_id)
+      end
+
+      # Direct renaming of features
+      FEATURE_MAPPING.each do |from, to|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
```
gtanzillo, line 21: I'm not sure but, wouldn't this result in creating 2 product features where the identifier is `control`? Because on line 13 will create the new one and this line will find the old one and update the identifier to `control` as well

h-kataria: @gtanzillo https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_product_feature.rb#L20 should not allow duplicates, correct?

Fryguy: @h-kataria Since you stubbing that model than no code in core is being executed, so no.

Fryguy:
@h-kataria I think this PR looks very similar to a previous migration, but I can't find it.  If so, did you copy/paste it exactly? (if so, curious how we missed @gtanzillo's duplication question in the original PR)

db/migrate/20200910163614_adjust_control_explorer_features.rb
Fryguy: I think this does create 2 control entries...you should be able to verify by changing the `it 'renames the features' do` spec to count the number of "control" product features and ensure there is only one.

db/migrate/20200910163614_adjust_control_explorer_features.rb
```diff
@@ -0,0 +1,42 @@
...
+  def down
+    return if MiqRolesFeature.none?
+
+    say_with_time 'Adjusting split explorer features back to control explorer' do
+      admin_feature = MiqProductFeature.find_or_create_by!(:identifier => 'control_explorer')
+      MiqRolesFeature.where(:miq_product_feature_id => admin_feature.id).each do |feature|
+        old_feature_ids.each do |id|
+          MiqRolesFeature.create!(:miq_product_feature_id => id, :miq_user_role_id => feature.miq_user_role_id)
+        end
+      end
```
Fryguy, line 29: Feels like this piece is now no longer needed either

db/migrate/20200910163614_adjust_control_explorer_features.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      FEATURE_MAPPING.each do |to, from|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+    end
+  end
+
+  private
+
+  def old_feature_ids
```
Fryguy, line 39: And if that goes away then so does this.

db/migrate/20200910163614_adjust_control_explorer_features.rb
```diff
@@ -0,0 +1,29 @@
+class AdjustControlExplorerFeatures < ActiveRecord::Migration[5.2]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
```
Fryguy, line 3: This can be dropped now too.

gtanzillo (approved):
LGTM 👍

PR #511, opened 2020-09-14
Add target platform to the configuration profiles

epostea:
As part of the [IBM Terraform provider](https://github.com/ManageIQ/manageiq-providers-ibm_terraform) implementation, we discover the terraform templates from CAM and currently we represent them as configuration profiles. 

Template contains metadata (name, description, cloud provider) and they can have multiple template versions. 
Each version has a unique name (i.e. "v1.0.0") and it either has the template source reference such as git repo, branch, token, etc or contains the template code source if the template version is stored locally. A template can have only one default version.

This PR is to enhance the configuration_profiles table and add the cloud provider information.

### DIFF
--- a/db/migrate/20200914005000_add_target_platform_to_configuration_profiles.rb
+++ b/db/migrate/20200914005000_add_target_platform_to_configuration_profiles.rb
@@ -1,0 +1,5 @@
+class AddTargetPlatformToConfigurationProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :configuration_profiles, :target_platform, :string
+  end
+end


db/migrate/20200914005000_add_version_and_cloud_provider_to_configuration_profiles.rb
```diff
@@ -0,0 +1,6 @@
+class AddVersionAndCloudProviderToConfigurationProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :configuration_profiles, :default_version, :string
```
agrare, line 3: I wonder if it'd be better to collect all versions and have a "default" boolean

db/migrate/20200914005000_add_version_and_cloud_provider_to_configuration_profiles.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_column :configuration_profiles, :cloud_provider, :string
```
agrare, line 4: I think `cloud_provider` is going to be confusing because we have Cloud Providers in MIQ along with Infra Providers and others

I think something like "target_platform" would be better

spec/migrations/20200914005000_add_version_and_cloud_provider_to_configuration_profiles.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe AddVersionAndCloudProviderToConfigurationProfiles do
+  let(:config_profile_stub) { migration_stub(:ConfigurationProfile) }
+
+  migration_context :up do
+    it 'creates the new columns' do
+      config_profile = config_profile_stub.create!(:name => "test")
+
+      expect(config_profile).to_not respond_to(:default_version)
+      expect(config_profile).to_not respond_to(:cloud_provider)
```
agrare, line 11: We don't commonly say you don't need tests :laughing:, but if you're just adding or removing columns not doing any data migrations you don't really need specs since you're just testing that ActiveRecord `add_column` works

epostea: Good to know, thx :-)

db/migrate/20200914005000_add_version_and_cloud_provider_to_configuration_profiles.rb
epostea: OK, it makes sense.

db/migrate/20200914005000_add_version_and_cloud_provider_to_configuration_profiles.rb
epostea: Removed the versions column from the existing PR. Will create a separate one when we have a better solution for representing the template versions in manageig

agrare:
@epostea is there anything else that you would like to add to the schema with this change?

If not can you update the title and description to reflect that we are adding the target platform, and the commits can probably be squashed down to just one here.

agrare (approved):

PR #510, opened 2020-09-12
[wip] convert resource pool relationship data to ancestry

d-m-u:
we're working on converting resource pool genealogy to use ancestry. In order to do this, we need to use the existing relationship resource information on all resource pool  with genealogy rels of type ResourcePool and convert those resource ids into resource pool  ids. 

This PR, just like the equivalent https://github.com/ManageIQ/manageiq-schema/pull/492 , 1) adds the necessary index and 2) gets the ancestry information from the relationships, splitting on slashes, then maps those back to the resource pool  ancestry records. 

https://github.com/ManageIQ/manageiq-schema/pull/513 was opened merely for testing and just adds the two lines in this code necessary for ancestry. I can take those two lines out if needed, or this can be merged and 513 closed. 

## needs concurrent merge with:
the core pr https://github.com/ManageIQ/manageiq/pull/20392

PR #513, opened 2020-09-18
[wip] add ancestry to folders and resource pools

d-m-u:
we're working on converting resource pool genealogy and folder genealogy to use ancestry and it's better to have the change that adds ancestry be separate from the migration to sort out the relationship side. 

split out from https://github.com/ManageIQ/manageiq-schema/pull/497 and https://github.com/ManageIQ/manageiq-schema/pull/510 and if those are merged as is, this can be closed. or if separation of concerns matters, i can take out the relevant lines from those PRs and leave this intact. 

## related to 
 https://github.com/ManageIQ/manageiq/pull/20392


d-m-u:
closing because the right place for this is in https://github.com/ManageIQ/manageiq-schema/pull/510 and https://github.com/ManageIQ/manageiq-schema/pull/497 which are also going to be consolidated into one pr in the future.

PR #514, opened 2020-09-21
Rename the ContainerLabelTagMapping table to ProviderTagMapping

agrare:
This mapping table is used by more than just container providers now so it should be renamed accordingly

Core issue: https://github.com/ManageIQ/manageiq/pull/20577


Fryguy:
@agrare Let's get this merged first, with a simple change in the model to point to the new table.  Then we can follow up with https://github.com/ManageIQ/manageiq/pull/20567 which is much bigger.

agrare:
:+1: https://github.com/ManageIQ/manageiq/pull/20577

PR #515, opened 2020-09-24
Add vendor to configured_systems

epostea:
As part of the [IBM Terraform provider](https://github.com/ManageIQ/manageiq-providers-ibm_terraform) implementation, we discover the virtual machines from CAM and represent them as configured systems. 

The terraform resources contain a resource type (i.e. aws_instance) and we would like to capture this information in the configured systems.

Parent user story: https://github.com/ManageIQ/manageiq-providers-ibm_terraform/issues/28

### DIFF
--- a/db/migrate/20200924151045_add_vendor_to_configured_systems.rb
+++ b/db/migrate/20200924151045_add_vendor_to_configured_systems.rb
@@ -1,0 +1,5 @@
+class AddVendorToConfiguredSystems < ActiveRecord::Migration[5.2]
+  def change
+    add_column :configured_systems, :vendor, :string
+  end
+end


Fryguy:
:+1: on the field...not sure on the name... (my concern is that system_type could be confused for a polymorphic relationship).  @agrare Thoughts?  I'm wondering if maybe real STI would make sense here, but the downside to that might be that the subtype is truly dynamic.

epostea:
> 👍 on the field...not sure on the name... (my concern is that system_type could be confused for a polymorphic relationship). @agrare Thoughts? I'm wondering if maybe real STI would make sense here, but the downside to that might be that the subtype is truly dynamic.

Thanks Jason, I was not sure about the column name either. What about `system_category`?

Fryguy:
What about `vendor`? This would match what we have in the `vms` table, which, in turn may allow us to reuse things like the vendor -> icon decoration

epostea:
> What about `vendor`? This would match what we have in the `vms` table, which, in turn may allow us to reuse things like the vendor -> icon decoration

Nice, let's use `vendor` then. Will update the PR shortly.

agrare:
:+1: yeah I like vendor

chessbyte:
If we need to expand in the future, we can have another column called `vendor_service`.  That way, `vendor` can be `AWS` and `vendor_service` can be `ec2` or `rds`, for example.

agrare (approved):

PR #519, opened 2020-10-05
Schema dumper fixes

Fryguy:
@NickLaMuro Please review.  Each commit is a separate fix

- Fix issue where table number was dumped as a string
- Fix issue where indent of the trigger did not match original indent
- Fix sequences on metrics subtables 

These were determined by using the Schema Diff tool that is a part of pgadmin4.  After these changes there are no differences found (aside from the schema_migrations_ran_id_seq start value, which I'm not sure we can do anything about, but I will look into tackling in a follow-up PR).

### DIFF
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -5,6 +5,7 @@ module ManageIQ
 
       def tables(stream)
         super
+        miq_metric_table_sequences(stream)
         miq_metric_table_constraints(stream)
         miq_metric_views(stream)
         triggers(stream)
@@ -24,6 +25,15 @@ module ManageIQ
         return unless comment
 
         stream.puts "  change_column_comment #{remove_prefix_and_suffix(table).inspect}, #{pk.inspect}, #{comment.inspect}"
+        stream.puts
+      end
+
+      def miq_metric_table_sequences(stream)
+        inherited_metrics_tables.each do |(table, inherit_from)|
+          stream.puts "  change_miq_metric_sequence #{table.inspect}, " \
+                          "#{inherit_from.inspect}"
+        end
+
         stream.puts
       end
 
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -66,14 +66,12 @@ module ManageIQ
           # - Removes the "BEGIN" and "END;" stanzas
           # - Chomps the new lines from the beginning (keeps existing indent)
           # - Strips whitespace from the end (no need to worry about indent)
-          # - Indents by four spaces
           #
           # For formatting, we want to ensure at least some bit of an indent,
           # regardless of how the SQL string was originally added.
           #
           formatted_body = body.gsub(/(BEGIN$|^END;$)/, "")
                                .reverse.chomp.reverse.rstrip
-                               .indent(4)
 
           stream.puts "  add_trigger #{direction_key.inspect}, "       \
                                     "#{table.inspect}, "               \
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -5,7 +5,7 @@ module ManageIQ
 
       def tables(stream)
         super
-        miq_metric_table_contraints(stream)
+        miq_metric_table_constraints(stream)
         miq_metric_views(stream)
         triggers(stream)
       end
@@ -28,18 +28,18 @@ module ManageIQ
       end
 
       # Must be done after all of the table definitions since `metrics_01` is
-      # dumpped prior to `metrics_base`, etc.
+      # dumped prior to `metrics_base`, etc.
       #
-      def miq_metric_table_contraints(stream)
+      def miq_metric_table_constraints(stream)
         inherited_metrics_tables.each do |(table, inherit_from)|
-          match             = table.match(METRIC_ROLLUP_TABLE_REGEXP)
+          child_table_num   = table.match(METRIC_ROLLUP_TABLE_REGEXP)[:CHILD_TABLE_NUM].to_i
           child_table       = remove_prefix_and_suffix(table).inspect
           primary_condition = if inherit_from.include?("rollup")
                                 "capture_interval_name != ? AND EXTRACT(MONTH FROM timestamp) = ?"
                               else
                                 "capture_interval_name = ? AND EXTRACT(HOUR FROM timestamp) = ?"
                               end
-          conditions        = [primary_condition, "realtime", match[:CHILD_TABLE_NUM]]
+          conditions        = [primary_condition, "realtime", child_table_num]
 
           stream.puts "  add_miq_metric_table_inheritance #{child_table}, " \
                           "#{inherit_from.inspect}, "                       \
--- a/lib/manageiq/schema/schema_statements.rb
+++ b/lib/manageiq/schema/schema_statements.rb
@@ -14,6 +14,10 @@ module ManageIQ
         execute("DROP FUNCTION IF EXISTS #{quoted_name}();", 'Drop trigger function')
       end
 
+      def drop_sequence(table)
+        execute("DROP SEQUENCE #{table}_id_seq CASCADE", 'Drop sequence')
+      end
+
       def create_miq_metric_view(name)
         execute("CREATE VIEW #{name} AS SELECT * FROM #{name}_base")
         execute("ALTER VIEW #{name} ALTER COLUMN id SET DEFAULT nextval('#{name}_base_id_seq')")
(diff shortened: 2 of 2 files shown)


lib/manageiq/schema/schema_dumper.rb
```diff
@@ -27,19 +28,28 @@ def add_id_column_comment(table, stream)
...
+          stream.puts "  change_miq_metric_sequence #{table.inspect}, " \
+                          "#{inherit_from.inspect}"
+        end
+
+        stream.puts
+      end
+
       # Must be done after all of the table definitions since `metrics_01` is
-      # dumpped prior to `metrics_base`, etc.
+      # dumped prior to `metrics_base`, etc.
```
NickLaMuro, line 41: 🤦 

Thanks for fixing my derp.

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -27,19 +28,28 @@ def add_id_column_comment(table, stream)
...
       #
-      def miq_metric_table_contraints(stream)
+      def miq_metric_table_constraints(stream)
```
NickLaMuro, line 43: 🤦 🤦 

I hope this isn't a trend...

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -28,18 +28,18 @@ def add_id_column_comment(table, stream)
...
 
       # Must be done after all of the table definitions since `metrics_01` is
-      # dumpped prior to `metrics_base`, etc.
+      # dumped prior to `metrics_base`, etc.
       #
-      def miq_metric_table_contraints(stream)
+      def miq_metric_table_constraints(stream)
         inherited_metrics_tables.each do |(table, inherit_from)|
-          match             = table.match(METRIC_ROLLUP_TABLE_REGEXP)
+          child_table_num   = table.match(METRIC_ROLLUP_TABLE_REGEXP)[:CHILD_TABLE_NUM].to_i
```
NickLaMuro, line 45: Okay, for anyone else looking over this and at a first glance are confused at what this change was, the only difference here is the addition of the `.to_i`.  The `[:CHILD_TABLE_NUM]` portion was previously below, and the variable name is just being refactored at bit.

I think the reason I split those up is I might have had this as a single method while developing this and the `.match` was also an existence check at the same time (to avoid a `nil` `NoMethodError`), which I think is still being done else where in the dumper.  I think the refactoring is fine, so I have no problem with the change.

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -66,14 +66,12 @@ def triggers(stream)
...
           # - Chomps the new lines from the beginning (keeps existing indent)
           # - Strips whitespace from the end (no need to worry about indent)
-          # - Indents by four spaces
           #
           # For formatting, we want to ensure at least some bit of an indent,
           # regardless of how the SQL string was originally added.
           #
           formatted_body = body.gsub(/(BEGIN$|^END;$)/, "")
                                .reverse.chomp.reverse.rstrip
-                               .indent(4)
```
NickLaMuro, line 76: So I am not going to "put my foot down" on this change, but the reason this existed was so that this had a minimum amount of indent when dumped to the `db/schema.rb`.

Alternatively to this, you could "unindent" on the way out when executing `add_trigger` if you wanted the pgadmin diff to match.  I didn't think that postgres stored the whitespace, just the function itself (though, in retrospect, I realize I was a fool for thinking that), so I figured it didn't matter since functionally it was equivalent.

lib/manageiq/schema/schema_statements.rb
```diff
@@ -42,6 +46,11 @@ def drop_miq_metric_table_inheritance(table, inherit_from)
         execute("ALTER TABLE #{quoted_table} NO INHERIT #{quoted_inherit}", 'Drop table inheritance')
       end
 
+      def change_miq_metric_sequence(table, inherit_from)
+        drop_sequence(table)
+        change_column_default(table, :id, -> { "nextval('#{inherit_from}_id_seq')" })
```
NickLaMuro, line 51: Have you tested this change with `rake db:schema:load`?  This was part of what I was making sure didn't cause an error with this commit from my [original PR](https://github.com/ManageIQ/manageiq-schema/pull/504):

- https://github.com/ManageIQ/manageiq-schema/commit/ca2a5dbac4d1f33461a81bc97bf94f2f6096c24a

If this does work, does that mean we can remove `column_spec_for_primary_key`, or is this part to augment it?

NickLaMuro (changes requested):
I am actually mostly good with this PR.  Just one major question, and a few comments/suggestions, but overall I think this is good.

Thanks for doing this!

lib/manageiq/schema/schema_dumper.rb
Fryguy: But it does have indenting already without this extra indenting.  Removing this extra indenting is also the only value that actually makes it correct (correct in the sense that it matched what it previously in the database, which itself it also indented a little bit).  It seems the schema dumper automatically indents a little bit for presentation purposes.  This is the before/after (notice that the extra indent actually makes it extra-indented)

keeping `.indent(4)`

```ruby
  add_trigger "insteadof", "metrics", "metrics_partition", <<-SQL
          CASE EXTRACT(HOUR FROM NEW.timestamp)
            WHEN 0 THEN
              INSERT INTO metrics_00 VALUES (NEW.*);
            ...
            WHEN 23 THEN
              INSERT INTO metrics_23 VALUES (NEW.*);
          END CASE;
          RETURN NEW;
  SQL
```

removing `.indent(4)`:

```ruby
  add_trigger "insteadof", "metrics", "metrics_partition", <<-SQL
      CASE EXTRACT(HOUR FROM NEW.timestamp)
        WHEN 0 THEN
          INSERT INTO metrics_00 VALUES (NEW.*);
        ...
        WHEN 23 THEN
          INSERT INTO metrics_23 VALUES (NEW.*);
      END CASE;
      RETURN NEW;
  SQL
```

lib/manageiq/schema/schema_statements.rb
Fryguy: I did not...I only tried `rake db:reset`...I'll give that a shot.

Fryguy: Just tried with `db:schema:load` and it works :+1:

---

I tried removing/adjusting column_spec_for_primary_key, but there are multiple problems.  If I straight remove it, the problem is that Rails tries to create the id column with

```sql
"id" bigserial DEFAULT nextval('metric_rollups_base_id_seq'::regclass) NOT NULL PRIMARY KEY
```

The problem there is that there is a sequence declared, but `bigserial` _also_ implies adding a default sequence, so there are 2 sequences, this yielding this error: `ERROR:  multiple default values specified for column "id" of table "metric_rollups_01"`.  The fix is to instead use `bigint` when you declare a sequence, and I hacked that in, but then I get a different error: `ERROR:  relation "metric_rollups_base_id_seq" does not exist`, which is because the tables are in alphabetical order and not in topological order.

I'll investigate changing the ordering, but I still am not sure how to change the bigserial to bigint without a manual hack.

Fryguy: Yeah, the bigger problem I'm having is forcing rails to use `bigint` over `bigserial`.  Under the covers, the Rails type is specified, and `:bigint` auto-maps to `bigserial`.

I think we have to leave the changes as is.  That is, keep `column_spec_for_primary_key` to remove the default, and let it create the sequence via `bigserial`, then we run this method which removes the sequence and puts the base sequence in.

NickLaMuro: Okay, cool.  I just was a bit concerned with that because it looked very similar to what was in the commit message I linked, so if we are good here with `db:schema:load`, no worries then.

Kinda sucks it requires such a round about way of handling it, but what we are dong is a bit different from a standard Rails deployment.

Anyway, I am good with this change, just wanted to make sure.  Thanks!

lib/manageiq/schema/schema_dumper.rb
NickLaMuro: @Fryguy it only does the indenting because there was extra indenting in the first place since we didn't use the `<~SQL` when doing the original migrations:

https://github.com/ManageIQ/manageiq-schema/blob/47979dabfb7632cd47d9ba4e7f87760af65eb0b9/db/migrate/20190122213042_use_views_for_metrics.rb#L59-L63

https://github.com/ManageIQ/manageiq-schema/blob/47979dabfb7632cd47d9ba4e7f87760af65eb0b9/db/migrate/20190122213042_use_views_for_metrics.rb#L115-L119

So this was making up for that.

Fryguy: I guess what I'm thinking is I'd expect the schema dumper to preserve the original indenting, otherwise it creates a delta between what's dumped and what's in the DB.

NickLaMuro (approved):
I am good with this.  While having the `.indent` would be nice, I think Jason is correct that it is better to keep in line with what is in the DB, so that can be dropped.

👍

jrafanie (approved):

PR #520, opened 2020-10-08
Add orchestration stacks reference to configured systems

epostea:
As discussed in https://github.com/ManageIQ/manageiq-providers-ibm_terraform/issues/36, we need to add a reference to `orchestration_stacks` in the `configured_systems`

### DIFF
--- a/db/migrate/20201008185329_add_orchestration_stacks_to_configured_systems.rb
+++ b/db/migrate/20201008185329_add_orchestration_stacks_to_configured_systems.rb
@@ -1,0 +1,5 @@
+class AddOrchestrationStacksToConfiguredSystems < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :configured_systems, :orchestration_stack
+  end
+end


epostea:
@miq-bot add-review @agrare

db/migrate/20201008185329_add_orchestration_stacks_to_configured_systems.rb
```diff
@@ -0,0 +1,5 @@
+class AddOrchestrationStacksToConfiguredSystems < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :configured_systems, :orchestration_stacks, :type => :bigint, :index => true
```
agrare, line 3: This should be singular, `:orchestration_stack` since this will create a `:orchestration_stack_id` column which will work with the `belongs_to :orchestration_stack` association

Also `:type => :bigint, :index => true` are the defaults for `add_reference` on rails 5.2 [ref](https://apidock.com/rails/v5.2.3/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference) so you don't have to pass them

agrare (changes requested):

db/migrate/20201008185329_add_orchestration_stacks_to_configured_systems.rb
epostea: thanks! i have fixed the migration script

agrare (approved):
:+1: LGTM thanks @epostea !

PR #521, opened 2020-10-12
Update vms table 'format' column comment

jaywcarman:
The vms table 'format' column is being used by manageiq-providers-ibm_cloud:
ManageIQ/manageiq-providers-ibm_cloud#82

In this PR I'm trying update the comment to remove the deprecation note. I've marked
it WIP because I'm unsure of the DB migration mechanics.

### DIFF
--- a/db/migrate/20201012120000_update_comment_to_vms_table_format.rb
+++ b/db/migrate/20201012120000_update_comment_to_vms_table_format.rb
@@ -1,0 +1,5 @@
+class UpdateCommentToVmsTableFormat < ActiveRecord::Migration[5.2]
+  def up
+    change_column_comment :vms, :format, "The format of the VM's disk, such as vmdk, qcow2, and tier1."
+  end
+end


db/migrate/20201012120000_update_vms_table_comment.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateCommentToVmsTableFormat < ActiveRecord::Migration[5.0]
```
Fryguy, line 1: The class name must match the underscored file name (so one of them needs to adjust).  Additionally, this should be `[5.2]`

db/migrate/20201012120000_update_vms_table_comment.rb
```diff
@@ -0,0 +1,5 @@
...
+  def up
+    change_column_comment :vms, :format, "The format of the VM's disk, such as tier1."
```
Fryguy, line 3: `tier1` isn't clear to me...perhaps these are more/better examples.  @agrare ?

jaywcarman: Let me give some context to explain my approach (and maybe you'll tell I'm trying to push a square peg through a round hole).

In IBM Cloud PowerVS, VMs have an attribute that determines which type of volume can be attached.  We're writing this volume type in this `vms` table `format` column. Also, this type corresponds to `CloudVolumeType` stored in the `cloud_volume_types` DB table. We haven't implemented any of it yet, but in the future we'll want to use this data for things like filtering the list of VMs shown in the attach volume interface.

Since it's been marked deprecated I'm assuming no other providers make use of it, so I changed the example name to one of the types we use.

agrare: Some of the other options would be `vmdk, qcow2, raw`

db/migrate/20201012120000_update_vms_table_comment.rb
jaywcarman: Updated

db/migrate/20201012120000_update_vms_table_comment.rb
Fryguy: > but in the future we'll want to use this data for things like filtering the list of VMs shown in the attach volume interface.

Typically we do this with an "allowed_cloud_volume_types" method (as a part of the provisioning workflow), but yeah, I could see under the covers that method using this column.

db/migrate/20201012120000_update_comment_to_vms_table_format.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateCommentToVmsTableFormat < ActiveRecord::Migration[5.2]
+  def up
+    change_column_comment :vms, :format, "The format of the VM's disk, such as tier1."
```
agrare, line 3: @jaywcarman can you change the comment to something like `"The format of the VM's disk, such as vmdk, qcow2, and tier1.`

I don't think parsing a column comment is going to work better than other options so the list doesn't have to be exhaustive but it should be descriptive.

agrare (changes requested):

db/migrate/20201012120000_update_comment_to_vms_table_format.rb
agrare: Really we just need to indicate that this column is no longer "deprecated" (even though it never really was)

jaywcarman: I've updated the comment to include the other example values, along with deprecating the deprecation :laughing:

agrare: :rofl: perfect!

PR #523, opened 2020-10-21
Introduce Openstack cloud manager child manager migration

kbrock:
Depends on https://github.com/ManageIQ/manageiq/pull/20701

Child storage managers were introduced in E release.
When creating a cloud manager, these child managers are auto created.

For existing ems, the cloud volumes were linked to the cloud manager but
they now will point to the storage managers.

to avoid a migration, we had introduced code in the model to do this migration.
This PR moves that code to migrations where it belongs.

This is only necessary for customers transitioning from darga and have
not run a refresh (in the past 5 years...)

Since we are deleting this from the model, we need this in here for safe measure.

### DIFF
--- a/db/migrate/20201019211215_create_legacy_managers.rb
+++ b/db/migrate/20201019211215_create_legacy_managers.rb
@@ -1,0 +1,88 @@
+class CreateLegacyManagers < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    include MigrationStubHelper
+
+    has_one  :swift_manager,
+             -> { swift_managers },
+             :foreign_key => :parent_ems_id,
+             :class_name  => "CreateLegacyManagers::ExtManagementSystem",
+             :autosave    => true
+
+    has_one  :cinder_manager,
+             -> { cinder_managers },
+             :foreign_key => :parent_ems_id,
+             :class_name  => "CreateLegacyManagers::ExtManagementSystem",
+             :autosave    => true
+
+    def self.openstack
+      where(:type => "ManageIQ::Providers::Openstack::CloudManager")
+    end
+
+    def self.swift_managers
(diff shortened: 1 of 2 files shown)


Fryguy:
LGTM... please also do the rubocops.

@agrare Can you review also?

agrare:
:+1: LGTM

Fryguy:
:+1:  This needs to be merged along with ManageIQ/manageiq#20701

kbrock:
actually - we can merge at any time.

not worried about someone introducing this problem before I merge the model code because this is only for older databases (ones older than 5 years and never having run the refresh in that time)

a recent database or a recent refresh will clear this up

db/migrate/20201019211215_create_legacy_managers.rb
```diff
@@ -0,0 +1,76 @@
...
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudObjectStoreObject < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Adding missing Swift and Cinder Managers for OpenStack Cloud Managers") do
+      ExtManagementSystem.openstack.each do |parent|
```
agrare, line 49: Not sure how much this matters since most people won't have 1000s of openstack cloud_managers but we could
`needs_child_cinder_manager = ExtManagementSystem.openstack - ExtManagementSystem.where(:type => "ManageIQ::Providers::Openstack::StorageManager::CinderManager").pluck(:parent_ems_id)`

So that we're not looping over every openstack cloud manager and doing a child query for each.

Might be over-optimization

kbrock: totally don't think this is necessary, but added it

But since this will be in all practical purposes a no-op, good to have minimal data come back

kbrock:
ok. fixed rubocop for the trailing commas

I'm good with this and would love a merge any time. no need to synchronize multiple repos

agrare (approved):

agrare:
Certainly looks like those method calls all have parens to me

PR #524, opened 2020-10-21
Drop unused / typo timestamp column (fix for Rails 6)

jrafanie:
It was meant to be t.timestamps :null => false to magically
add created_at and updated_at columns but without the 's', it
expects a column name.

In rails 4.2, 5.0, 5.1, and 5.2, this line is ignored without a column name.
In rails 6.0, it raises: Missing column name(s) for timestamp.

It's been 5 years, we didn't need timestamps up until now, so we'll just drop it.

### DIFF
--- a/db/migrate/20151104115400_create_chargeback_tiers.rb
+++ b/db/migrate/20151104115400_create_chargeback_tiers.rb
@@ -6,8 +6,6 @@ class CreateChargebackTiers < ActiveRecord::Migration[4.2]
       t.float :finish
       t.float :fixed_rate
       t.float :variable_rate
-
-      t.timestamp :null => false
     end
   end
 end


jrafanie:
I used this script to validate that versions 4.2-5.2 silently discard this typo, while 6.0 complains.  Since 4.2-6.0 don't create the timestamp columns at all, it's safe to just delete the line.

jrafanie:
```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end


VERSION = 5.2

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "activerecord", "~>#{VERSION}.0"
  # gem "sqlite3", "~> 1.3.6"
  gem 'pg', '~>0.19.0'
end

require "active_record"
require "minitest/autorun"
require "logger"

# Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "rails_test")
ActiveRecord::Base.logger = Logger.new(STDOUT)


class Payment < ActiveRecord::Base
end

class User < ActiveRecord::Base
end

class CreatePayments < ActiveRecord::Migration[VERSION]
  def change
    create_table :payments, force: true do |t|
      t.decimal :amount, precision: 10, scale: 0, default: 0, null: false
      t.timestamp :null => false
    end
  end
end

class CreateUsers < ActiveRecord::Migration[VERSION]
  def change
    create_table :users, force: true do |t|
      t.string :name
      t.timestamps :null => false
    end
  end
end

class BugTest < Minitest::Test
  def test_migrate_payment
    CreatePayments.migrate(:up)
    Payment.reset_column_information

    assert_equal %w{amount id}, Payment.column_names.sort
  end

  def test_migrate_user
    CreateUsers.migrate(:up)
    User.reset_column_information

    assert_equal %w{created_at id name updated_at}, User.column_names.sort
  end
end
```

Fryguy:
> I used this script to validate that versions 4.2-5.2 silently discard this typo, while 6.0 complains. Since 4.2-6.0 doesn't create the timestamp columns at all, it's safe to just delete the line.

WOW

Fryguy:
I also verified via the collapsed migration that the timestamp column does not exist.  Merging.

PR #525, opened 2020-10-22
Rename TagMapping resource_type for Amazon Vms and Images

agrare:
The Amazon provider was the first cloud provider to participate in the (originally) ContainerLabelTagMapping class.  It has since been expanded to include many providers including Openstack, Azure, and VMware vSphere.

This means that the original labeled_resource_type of "Vm" and "Image" meaning Amazon VMs and Templates is quite confusing.

Follow-up to: https://github.com/ManageIQ/manageiq-providers-vmware/pull/658#issuecomment-714732041
Depends on: https://github.com/ManageIQ/manageiq-ui-classic/pull/7438 https://github.com/ManageIQ/manageiq-providers-amazon/pull/671

### DIFF
--- a/db/migrate/20201022201901_update_amazon_tag_mapper_resource_type.rb
+++ b/db/migrate/20201022201901_update_amazon_tag_mapper_resource_type.rb
@@ -1,0 +1,22 @@
+class UpdateAmazonTagMapperResourceType < ActiveRecord::Migration[5.2]
+  class ProviderTagMapping < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Update ProviderTagMapping Vm label") do
+      ProviderTagMapping.where(:labeled_resource_type => "Vm").update_all(:labeled_resource_type => "VmAmazon")
+    end
+    say_with_time("Update ProviderTagMapping Image label") do
+      ProviderTagMapping.where(:labeled_resource_type => "Image").update_all(:labeled_resource_type => "ImageAmazon")
+    end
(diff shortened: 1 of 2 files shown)


agrare:
Now that kasparov is branched off I'm going to un-wip this

lpichler (approved):

db/migrate/20201022201901_update_amazon_tag_mapper_resource_type.rb
```diff
@@ -0,0 +1,14 @@
+class UpdateAmazonTagMapperResourceType < ActiveRecord::Migration[5.2]
+  class ProviderTagMapping < ActiveRecord::Base
+  end
+
+  def up
+    ProviderTagMapping.where(:labeled_resource_type => "Vm").update_all(:labeled_resource_type => "VmAmazon")
```
Fryguy, line 6: Can you add a `say_with_time` on all of these?

Fryguy:
> Now that kasparov is branched off I'm going to un-wip this

Is this master only?

db/migrate/20201022201901_update_amazon_tag_mapper_resource_type.rb
agrare: :+1:

agrare:
@Fryguy correct we don't need this on kasparov

Fryguy:
LGTM...just needs the dependent PR to be ready.

PR #530, opened 2020-11-03
Update the GCE CloudVolume STI class

agrare:
Google Cloud never had a CloudVolume subclass and these volumes are of the base class type.

Provider PR: https://github.com/ManageIQ/manageiq-providers-google/pull/167
Noticed from: https://github.com/ManageIQ/manageiq/pull/20761

### DIFF
--- a/db/migrate/20201103143405_fix_google_cloud_volume_sti.rb
+++ b/db/migrate/20201103143405_fix_google_cloud_volume_sti.rb
@@ -1,0 +1,30 @@
+class FixGoogleCloudVolumeSti < ActiveRecord::Migration[5.2]
+  class ExtManagementSystem < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudVolume < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system,
+               :foreign_key => :ems_id,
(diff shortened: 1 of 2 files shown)


Fryguy:
@agrare Does this and ManageIQ/manageiq-providers-google#167 need to be kasparov/yes?

db/migrate/20201103143405_fix_google_cloud_volume_sti.rb
```diff
@@ -0,0 +1,30 @@
...
+    belongs_to :ext_management_system,
+               :foreign_key => :ems_id,
+               :class_name  => "FixGoogleCloudVolumeSti::ExtManagementSystem"
+  end
+
+  GOOGLE_CLOUD_KLASS        = "ManageIQ::Providers::Google::CloudManager".freeze
+  GOOGLE_CLOUD_VOLUME_KLASS = "ManageIQ::Providers::Google::CloudManager::CloudVolume".freeze
+
+  def up
+    gce_relation = ExtManagementSystem.in_my_region.where(:type => GOOGLE_CLOUD_KLASS)
```
Fryguy, line 20: I totally forgot before merging, but can you do a follow up PR with a `say_with_time` around all these?

agrare:
@Fryguy no this is fine just being in L, we really just need this for the DDF cloud_volume forms

db/migrate/20201103143405_fix_google_cloud_volume_sti.rb
agrare: :+1: https://github.com/ManageIQ/manageiq-schema/pull/534

PR #531, opened 2020-11-04
Rails 6.0

NickLaMuro:
**Requires new release of `activerecord-id_regions`**:  https://github.com/ManageIQ/activerecord-id_regions/pull/18

Links
-----

* Epic:  https://github.com/ManageIQ/manageiq/issues/19977
* Current Cross Repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/179

### DIFF
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "more_core_extensions",    ">= 3.5", "< 5"
   spec.add_dependency "pg"
   spec.add_dependency "pg-pglogical",            "~> 2.1.1"
-  spec.add_dependency "rails",                   "~>5.2.4", ">=5.2.4.3"
+  spec.add_dependency "rails",                   "~>6.0.0"
 
   spec.add_development_dependency "manageiq-style"
   spec.add_development_dependency "rspec"
--- a/spec/support/migration_helper.rb
+++ b/spec/support/migration_helper.rb
@@ -103,7 +103,11 @@ module Spec
 
       def migrate_to(version)
         suppress_migration_messages do
-          if ActiveRecord::VERSION::STRING >= "5.2.0"
+          if ActiveRecord::VERSION::STRING >= "6.0.0"
+            migration_dir  = Rails.application.config.paths["db/migrate"]
+            migration_conn = ::ActiveRecord::Base.connection.schema_migration
+            ActiveRecord::MigrationContext.new(migration_dir, migration_conn).migrate(version)
+          elsif ActiveRecord::VERSION::STRING >= "5.2.0"
             ActiveRecord::MigrationContext.new('db/migrate').migrate(version)
           else
             ActiveRecord::Migrator.migrate('db/migrate', version)
@@ -123,7 +127,13 @@ module Spec
       end
 
       def run_migrate
-        if ActiveRecord::VERSION::STRING >= "5.2.0"
+        if ActiveRecord::VERSION::STRING >= "6.0.0"
+          migration_dir  = Rails.application.config.paths["db/migrate"]
+          migration_conn = ::ActiveRecord::Base.connection.schema_migration
+          context        = ActiveRecord::MigrationContext.new(migration_dir, migration_conn)
+
+          context.run(migration_direction, this_migration_version)
+        elsif ActiveRecord::VERSION::STRING >= "5.2.0"
           ActiveRecord::MigrationContext.new('db/migrate').run(migration_direction, this_migration_version)
         else
           described_class.migrate(migration_direction)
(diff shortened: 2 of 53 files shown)


manageiq-schema.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "more_core_extensions",    ">= 3.5", "< 5"
   spec.add_dependency "pg"
   spec.add_dependency "pg-pglogical",            "~> 2.1.1"
-  spec.add_dependency "rails",                   "~>5.2.4", ">=5.2.4.3"
```
bdunne, line 28: Can we apply https://github.com/ManageIQ/activerecord-id_regions/pull/18/files#r537735807 here as well?

NickLaMuro: I don't see why we would, as this isn't a gem that ever needs to be released, and is only used by `ManageIQ/manageiq`, which now requires Rails 6.  The only place where I think this does make sense is in gems that are released to rubygems.org or rubygems.manageiq.org which this is not.

(This is going to be a carbon copy answer that I will be pasting in all provider repos where this comment exists)

manageiq-schema.gemspec
```diff
@@ -19,13 +19,13 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "ancestry"
-  spec.add_dependency "activerecord-id_regions", "~> 0.3.0"
+  spec.add_dependency "activerecord-id_regions", "~> 0.3.0", ">= 0.3.1"
```
bdunne, line 22:
```suggestion
  spec.add_dependency "activerecord-id_regions", "~> 0.3.1"
```

NickLaMuro: @bdunne I already changed this prior to even seeing this, but this isn't a pattern that I pulled out of thin air, as it is something we commonly do:

https://github.com/ManageIQ/manageiq/blob/a94268f6aab6d41b691fdb0ea8c07712f357aa61/Gemfile#L36

I didn't want to leave it to chance that it would work though, so I tried a slightly more restrictive approach by "raising the floor".  However, it seems like `bundler` wasn't having it:

https://travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/builds/207873090

So I reverted back.  Though, I don't think it was wrong in at least considering it.

bdunne: I think it makes sense to specify `"~> x.y", ">= x.y.1"` when we want our dependency to be `~> x.y` (letting y & z float), but require at least `x.y.1` because there was a security bug or something in `x.y.0`.  But in a case like this where we're already locked down to z, I think it adds extra complexity.

i.e. If activerecord-id_regions was 1.0, I agree that we would probably want `"~> 1.0", ">= 1.0.1"`

bdunne: But since it doesn't fix a security issue, we probably don't need to change anything since bundler should resolve properly as is.

NickLaMuro: In this case was more being explicit, and not necessarily requiring it.  I think this was more residual habit from days of 1.4 bundler where `molinillo` would be pretty good, but have a few edge cases where it couldn't properly resolve things and got confused.

But because of the other bump in the `rails` dependency in this gem, `manageiq` will always resolve to `< 0.3.1` for Rails 5.x and `>= 0.3.1` for Rails 6.0ish, so you are correct and going with just the `~> 0.3.0` we had is fine (again, what I already changed it back to).

spec/support/migration_helper.rb
```diff
@@ -103,7 +103,11 @@ def suppress_migration_messages
 
       def migrate_to(version)
         suppress_migration_messages do
-          if ActiveRecord::VERSION::STRING >= "5.2.0"
+          if ActiveRecord::VERSION::STRING >= "6.0.0"
+            migration_dir  = Rails.application.config.paths["db/migrate"]
+            migration_conn = ::ActiveRecord::Base.connection.schema_migration
+            ActiveRecord::MigrationContext.new(migration_dir, migration_conn).migrate(version)
```
bdunne, line 109: Is this also a problem on Rails 5.2?  https://github.com/ManageIQ/manageiq/pull/20784

NickLaMuro: I puled this code from here:

https://github.com/ManageIQ/manageiq/pull/20778/commits/4d95a2ada9bac2368815669bee075a9e204e33d3

Which did have the `Rails.application.config.paths["db/migrate"]` fix from ManageIQ/manageiq#20784 already in it, and didn't bother to change it.

That said, since this is a `spec/*` file, unless we plan on running specs in 5.2 in containers, I don't see it being a problem, and the new code here was just some copy pasta.

PR #535, opened 2020-11-11
update i18n config fallbacks line

d-m-u:
I think this needs to be updated:

Please check your Rails app for 'config.i18n.fallbacks = true'.
If you're using I18n (>= 1.1.0) and Rails (< 5.2.2), this should be
'config.i18n.fallbacks = [I18n.default_locale]'.
If not, fallbacks will be broken in your app by I18n 1.1.x.

### DIFF
--- a/spec/dummy/config/environments/production.rb
+++ b/spec/dummy/config/environments/production.rb
@@ -63,7 +63,7 @@ Rails.application.configure do
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
+  config.i18n.fallbacks = [I18n.default_locale]
 
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify


chessbyte:
@d-m-u in the core [Gemfile](https://github.com/ManageIQ/manageiq/blob/master/Gemfile), `gem "rails", "~>5.2.4", ">=5.2.4.4"` is specified.  The message `If you're using I18n (>= 1.1.0) and Rails (< 5.2.2)` seems to imply that this is not applicable to ManageIQ because we use a Rails version greater than 5.2.2.  What am I missing?

d-m-u:
You're right! Sorry. It's odd that we're using this line in the core repo then.

chessbyte:
> You're right! Sorry. It's odd that we're using this line in the core repo then.

Which line?  If something is odd, let's try to get to the bottom of it.  Maybe something will need to be fixed at the end of the investigation.

d-m-u:
| config.i18n.fallbacks = [I18n.default_locale]

That's the line for the rails version we're past, and it's what's in core.

d-m-u:
```
Please check your Rails app for 'config.i18n.fallbacks = true'.
If you're using I18n 1.1.x and Rails (< 6.0), this should be
'config.i18n.fallbacks = [I18n.default_locale]'.
If not, fallbacks will be broken in your app by I18n 1.1.x.
```

that's the commit message from the line in core

chessbyte:
@Fryguy @mzazrivec can you guys weigh in on this?

Fryguy:
@d-m-u How did you make this error show up?

Fryguy:
@chessbyte The core Gemfile is irrelevant here, because we use a dummy app, however the [.gemspec](https://github.com/ManageIQ/manageiq-schema/blob/master/manageiq-schema.gemspec#L28) here hits on almost the same thing you stated.

d-m-u:
It's on our travis output: https://travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/jobs/433221782#L1080

Fryguy:
ohhh I see ... that's just a post-install message from i18n ... it will happen any time you install i18n and doesn't mean anything.

d-m-u:
I'm just sayin', you all have both lines in various places, and this was something someone tried to address at some point.

Fryguy:
Fair point... So i checked what a vanilla Rails 5.2.4.4 has and it's:

```
  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
  # the I18n.default_locale when a translation cannot be found).
  config.i18n.fallbacks = true
```

The change was made in core in https://github.com/ManageIQ/manageiq/pull/18281 in order to deal with the upgrade of i18n at the time (and at that time the message stated "rails < 6.0".  That seems to have been relaxed, and since we are on a valid version, we can probably revert that change in core now.

d-m-u:
Thanks!

PR #536, opened 2020-11-12
[wip] replace relationships with miq_set_memberships table

d-m-u:
the work to refactor miq_set to just be a normal habtm rather than using the relationships and ancestry hierarchy needs a new table, miq_set_memberships

related to https://github.com/ManageIQ/manageiq/pull/20665


db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  class MiqSetMembership < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      sets = Relationship.where(:relationship => "membership").roots
+
+      sets.each do |set|
+        members = set.children
```
d-m-u, line 16: pretty sure we want `descendants` not `children` here

spec/migrations/20190118204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  migration_context :up do
+    context "parent/child/grandchild rel" do
+      it 'updates ancestry' do
+        ns1 = ns_stub.create!
+        ns1_child = ns_stub.create!
+        ns1_grandchild = ns_stub.create!
+        ns1_rel = rel_stub.create!(:relationship => "membership", :ancestry => nil, :resource_id => ns1.id, :resource_type => "MiqAeNamespace")
+        ns1_child_rel = rel_stub.create!(:relationship => "membership", :ancestry => ancestry_for(ns1_rel), :resource_id => ns1_child.id, :resource_type => "MiqAeNamespace")
+        rel_stub.create!(:relationship => "membership", :ancestry => ancestry_for(ns1_rel, ns1_child_rel), :resource_id => ns1_grandchild.id, :resource_type => "MiqAeNamespace")
```
kbrock, line 17: this is pretty similar to the code in a couple of PRs.
and this is a simple linear 3 level relationship.

going line by line it looks good.
but I'm having trouble parsing it.

Also, on the note of contract tests, I see how you are creating these, but I'm not sure that this matches the way that it is coded elsewhere.
I do seem to remember a linear only 1 parent, 1 child model.

this seems to be different from the way it was stated above (`we only create these for the first level`).
sorry, being vague here - just a little confused and the create rel create ns isn't making it any easier for my brain.

spec/migrations/20190118204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        migrate
+
+        membership = find_membership(ns1_child)
+        expect(membership.miq_set_id).to eq(ns1.id)
+        expect(membership.member_type).to eq("MiqAeNamespace")
+        expect(membership.member_id).to eq(ns1_child.id)
```
kbrock, line 24: would be nice to do a single test per membership.
not sure if it is `attributes includes(:miq_set_id => ns1.id, :member_type => ...)` that may read better?

Also, would be nice if more memberships are tested for here - the parent, child, and grand are all setup correctly.

 guess that is another reason why I'm trying to reduce the per object length - so you can test all 3 and it isn't too verbose - and the actual test will be lost in the boiler plate

kbrock (commented):
I really like a single PR and 2 migrations in it.
Having the schema change in one PR and the changes in another makes running the specs more clear.

d-m-u:
```
I really like a single PR and 2 migrations in it.
Having the schema change in one PR and the changes in another makes running the specs more clear.
```

... so should I leave this?

spec/migrations/20190118204247_move_miq_set_memberships_spec.rb
d-m-u: it was intending as a first pass test to show that that my statement about how `we only create these for the first level` was in fact veridical.

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,39 @@
...
+        members.each do |member|
+          MiqSetMembership.create!(
```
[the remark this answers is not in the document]
d-m-u: @Fryguy, you have an opinion here possibly that you'd like to share?

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,39 @@
...
+  class Relationship < ActiveRecord::Base
+    has_ancestry
+  end
+
+  class MiqSetMembership < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      sets = Relationship.where(:relationship => "membership").roots
```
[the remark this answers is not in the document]
kbrock: ok, so my views on this have changed.

I'm guessing there are a bunch of these.

|record|ancestry|
|---|---|
|root|nil|
|first|root|
|second|root/first|

we want to create:
```ruby
Relationship.where(:relationship => "membership").not(:ancestry => nil).each do |node|
  set_id = node.root.resource_id
  MiqSetmembership.create(
            :miq_set_id  => set_id,
            :member_type => member.resource_type,
            :member_id   => member.resource_id,
            :position    => member.ancestry.count("/") +1
  )
end
```


```ruby
Relationship.where(:relationship => "membership").not(:ancestry => nil).each do |member|
  root_id = member.ancestry.split("/").first.to_i
  set_id = Relationship.select(:resource_id).where(:relationship => "membership, :ancestry => nil, :id => root_id).resource_id
  MiqSetmembership.create(
            :miq_set_id  => set_id,
            :member_type => member.resource_type,
            :member_id   => member.resource_id
            :position    => member.ancestry.count("/") +1
  )
end
```

and good news:

|ruby|sql|
|---|---|
|ancestry.count("/")|regexp_count(ancestry,'/')|
|ancestry.split("/").first.to_i|split_part(ancestry, '/', 1)::bigint|
|m1.all.each { |m| m2.create()}|insert m2 (col1, col2) select c1,c2 from m1|

... think this will be much easier sql for everyone to read without too  much mumbo jumbo

kbrock (commented):
really need to dive in here and figure out if depth is the length of the children or it is 2
also wondering if this needs STI

And better yet, is there a reason why we are writing this rather than using acts_as_list -- I remember he did a lot of work on that set to remove deadlocks -- would be nice to get that work and not do it ourselves. (I don't think we have the use cases that will cause deadlocks but the problem may be harder than it seems)

Think the sql will be simpler than the ruby here.

db/migrate/20190118200613_add_miq_set_memberships.rb
```diff
@@ -0,0 +1,8 @@
+class AddMiqSetMemberships < ActiveRecord::Migration[5.0]
+  def change
+    create_table :miq_set_memberships do |t|
```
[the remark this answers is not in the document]
Fryguy: There is no order in sets.

db/migrate/20190118204247_move_miq_set_memberships.rb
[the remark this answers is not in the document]
Fryguy: This seems fine...there aren't a lot of sets typically.

db/migrate/20190118204247_move_miq_set_memberships.rb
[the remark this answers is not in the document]
Fryguy: I think this can be queried in one shot with - 

```ruby
sets = Relationship.where(:relationship => "membership")
```

and then I think there are ancestry methods to arrange records that are already in memory, so in the end it's a single query for all of the data.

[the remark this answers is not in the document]
Fryguy: Actually, you don't even need any built in arrangement methods, since these are trivial trees:

```ruby
sets = Relationship.where(:relationship => "membership")
set_groups = sets.group_by(&:parent_id)
roots_by_id = set_groups.delete(nil).index_by(&:id)
set_groups.transform_keys! { |k| roots_by_id[k] }
```

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,35 @@
...
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      # ancestry holds the id for the relationship record that references the set
+      # the parent record contains the set object
+      # because these are trivial trees, we can look up the parent and get the resource_id
+      parent_miq_set_id = "(select rel2.resource_id from relationships rel2 where rel2.id = rels.ancestry::bigint)"
+
+      connection.execute(<<-SQL)
+        INSERT INTO miq_set_memberships (member_type, miq_set_id, member_id, created_at, updated_at)
+        SELECT rels.resource_type, #{parent_miq_set_id}, rels.resource_id, rels.created_at, rels.updated_at from relationships rels
+        WHERE rels.relationship = 'membership' AND rels.ancestry IS NOT NULL
```
Fryguy, line 21: minor, for readability...

```suggestion
        INSERT INTO miq_set_memberships (member_type, miq_set_id, member_id, created_at, updated_at)
          SELECT rels.resource_type, #{parent_miq_set_id}, rels.resource_id, rels.created_at, rels.updated_at
          FROM relationships rels
          WHERE rels.relationship = 'membership' AND rels.ancestry IS NOT NULL
```

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  class MiqSetMembership < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      # ancestry holds the id for the relationship record that references the set
+      # the parent record contains the set object
+      # because these are trivial trees, we can look up the parent and get the resource_id
+      parent_miq_set_id = "(select rel2.resource_id from relationships rel2 where rel2.id = rels.ancestry::bigint)"
```
Fryguy, line 16: Minor, for readability:

```suggestion
      parent_miq_set_id = "(SELECT rel2.resource_id FROM relationships rel2 WHERE rel2.id = rels.ancestry::bigint)"
```

spec/migrations/20190118204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_migration
+
+describe MoveMiqSetMemberships do
+  let(:rel_stub) { migration_stub(:Relationship) }
+  let(:membership_stub) { migration_stub(:MiqSetMembership) }
+  let(:ns_stub) { Class.new(ActiveRecord::Base) { self.table_name = "miq_ae_namespaces" } }
+
+  migration_context :up do
+    context "two trivial trees" do
+      it 'creates set memberships' do
```
kbrock, line 10: business statement (not a question about the code)

What is this code doing?

It seems you should have a create_set and create_set_member helper.

As far as people are concerned, they have a set and members - not children and relationships and stuff.
Not saying the abstraction won't be a little leaky, but this is a tad bit confusing

spec/migrations/20190118204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_migration
+
+describe MoveMiqSetMemberships do
+  let(:rel_stub) { migration_stub(:Relationship) }
+  let(:membership_stub) { migration_stub(:MiqSetMembership) }
+  let(:ns_stub) { Class.new(ActiveRecord::Base) { self.table_name = "miq_ae_namespaces" } }
+
+  migration_context :up do
+    context "two trivial trees" do
```
kbrock, line 9:
```suggestion
    context "two sets with one member each" do
```

kbrock (commented):
the code looks nice.

Your code is fixing a problem.
We have a set with members - a simple concept, but before they were stored in a parent/child type relationship. Now, they are stored with a simple rails belongs_to. (er, habtm - but anyway)

Reading this code doesn't tell me that.

Would be nice if I were able to easily see the great change you are making and get a relief that you are bringing us from a complex past to a simple straightforward future.

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  class MiqSetMembership < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      # ancestry holds the id for the relationship record that references the set
+      # the parent record contains the set object
+      # because these are trivial trees, we can look up the parent and get the resource_id
```
kbrock, line 15: sorry - I probably told you to write that last sentence and here I am changing again.

```suggestion
      # because these are trivial trees, we can get the set_id from the parent's resource_id
```

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,38 @@
...
+      parent_miq_set_id = "(SELECT rel2.resource_id FROM relationships rel2 WHERE rel2.id = rels.ancestry::bigint)"
```
kbrock, line 16: would this still read fine if you use `miq_set_id` instead of `parent_miq_set_id`?

(again see above disclaimer/apology)

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,38 @@
...
+          WHERE rels.relationship = 'membership' AND rels.ancestry IS NOT NULL
+      SQL
+
+      Relationship.where(:relationship => "membership").delete_all
+    end
+  end
+
+  def down
+    say_with_time("Converting miq_set_memberships to set relationships") do
+      MiqSetMembership.all.each do |m|
```
kbrock, line 31: ok. So according to above, we need relationship records for the sets and the membership records.

Think we need one loop for each of the sets. (probably store the set_id to relationship record id in a hash for the next step)

Then we can run the membership loop and the new relationship will have the set as the parent / ancestry value.

would be cool if the tests mimic each other in this way

kbrock (commented):
this looks good. Think the down needs a little love but all in all, this looks great

db/migrate/20190118204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,38 @@
+require 'ancestry'
+
+class MoveMiqSetMemberships < ActiveRecord::Migration[5.0]
+  class Relationship < ActiveRecord::Base
+    has_ancestry
```
Fryguy, line 5: Can this be removed?  (along with the ancestry gem reference?)

Fryguy: Also I think this needs the disabling of STI.

PR #537, opened 2020-12-02
Implement Host initiators

llivne:
Host Initiators (AKA physical storage hosts) are virtual entities that only exists as a set of definitions within the `PhysicalStorage' and hold the information about consumers/hosts that use (defined) the physical storage.

This PR includes the work for retrieving and presenting Host Initiators for a specific physical storage in manageiq.

g
h

Example data
---------
Below are diagrams that shows how we'd represent rather standard scenario. 

All items in white are already part of miq:
We have a single IRL storage-system Called SVC Storage 1.
It has 2 storage pools: Pool1 and Pool2.
Pool1 has two volumes: Volume1 and Volume2.

All items in blue are changes done in this PR:
The Physical Storage has a single host-initiator defined in it. It has been defined with two addresses: iqn-addr-1, and wwpn-addr-1. For the sake of brevity I won't use real IQN and WWPN addresses as they are too long. 

All items in peach are future drafts pending changes (not part of this PR):
The host-initiator in the storage-system has been mapped to Volume1.
A DB server consumes Volume1.
One of the FC addresses of the DB server is wwpn-addr-2. Remeber that this address was defined on the host-initiator above. This, together with the fact that the host-initiator was mapped to Volume1 allows the DB server to consume it.

========== using STI for SAN-Addresses ==========
In this diagram, the addresses are implemented as an STI.
There's a general SAN (Storage Area Network) Address table that holds all possible info for all address. It's expected to be a sparse table as concrete addresses will only populate their specific fields. 
So FcAddresses will have nil for their iscsi values such as IQN and CHAP. And the reverse holds true for IscsiAddresses.

links
----------
https://github.com/ManageIQ/manageiq-ui-classic/pull/7529
https://github.com/ManageIQ/manageiq/pull/20858
https://github.com/ManageIQ/manageiq-providers-autosde/pull/49

### DIFF
--- a/db/migrate/20201111154314_create_physical_storage_consumers.rb
+++ b/db/migrate/20201111154314_create_physical_storage_consumers.rb
@@ -6,6 +6,7 @@ class CreatePhysicalStorageConsumers < ActiveRecord::Migration[5.2]
       t.string :ems_ref
       t.string :uid_ems
       t.references :physical_storage, :type => :bigint, :index => true
+      # t.references :addresses, :type => :bigint, :index => true
 
       t.timestamps
     end
--- a/db/migrate/20201111154314_create_physical_storage_consumers.rb
+++ b/db/migrate/20201111154314_create_physical_storage_consumers.rb
@@ -6,7 +6,6 @@ class CreatePhysicalStorageConsumers < ActiveRecord::Migration[5.2]
       t.string :ems_ref
       t.string :uid_ems
       t.references :physical_storage, :type => :bigint, :index => true
-      # t.references :addresses, :type => :bigint, :index => true
 
       t.timestamps
     end
--- a/db/migrate/20201111154314_create_physical_storage_consumers.rb
+++ b/db/migrate/20201111154314_create_physical_storage_consumers.rb
@@ -1,12 +1,12 @@
-class CreateStorageConsumers < ActiveRecord::Migration[5.2]
+class CreatePhysicalStorageConsumers < ActiveRecord::Migration[5.2]
   def change
-    create_table :storage_consumers do |t|
+    create_table :physical_storage_consumers do |t|
       t.string :name
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
       t.string :uid_ems
       t.references :physical_storage, :type => :bigint, :index => true
-      t.references :addresses, :type => :bigint, :index => true
+      # t.references :addresses, :type => :bigint, :index => true
 
       t.timestamps
     end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -1,9 +1,12 @@
+
 class CreateAddresses < ActiveRecord::Migration[5.2]
   def change
     create_table :addresses do |t|
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
-      t.references :owner, :polymorphic => true, :index => true
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.references :physical_storage_consumer, :type => :bigint, :index => true
+      t.references :port, polymorphic: true, :index => true
     end
   end
 end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -1,15 +1,12 @@
+
 class CreateAddresses < ActiveRecord::Migration[5.2]
   def change
     create_table :addresses do |t|
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
       t.references :physical_storage, :type => :bigint, :index => true
-      t.string :chap_name
-      t.string :chap_secret
-      t.string :iqn
-      t.string :wwpn
-      t.string :port_type
       t.references :physical_storage_consumer, :type => :bigint, :index => true
+      t.references :port, polymorphic: true, :index => true
     end
   end
 end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -1,12 +1,9 @@
-
 class CreateAddresses < ActiveRecord::Migration[5.2]
   def change
     create_table :addresses do |t|
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
-      t.references :physical_storage, :type => :bigint, :index => true
-      t.references :physical_storage_consumer, :type => :bigint, :index => true
-      t.references :port, polymorphic: true, :index => true
+      t.references :owner, :polymorphic => true, :index => true
     end
   end
 end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -1,0 +1,15 @@
+class CreateAddresses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :addresses do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.string :chap_name
+      t.string :chap_secret
+      t.string :iqn
+      t.string :wwpn
+      t.string :port_type
+      t.references :storage_consumer, :type => :bigint, :index => true
+    end
+  end
+end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -9,7 +9,7 @@ class CreateAddresses < ActiveRecord::Migration[5.2]
       t.string :iqn
       t.string :wwpn
       t.string :port_type
-      t.references :storage_consumer, :type => :bigint, :index => true
+      t.references :physical_storage_consumer, :type => :bigint, :index => true
     end
   end
 end
--- a/db/migrate/20201124063738_create_addresses.rb
+++ b/db/migrate/20201124063738_create_addresses.rb
@@ -1,11 +1,11 @@
-
 class CreateAddresses < ActiveRecord::Migration[5.2]
   def change
     create_table :addresses do |t|
       t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
       t.string :ems_ref
-      t.references :physical_storage, :type => :bigint, :index => true
-      t.references :physical_storage_consumer, :type => :bigint, :index => true
+      # t.references :physical_storage, :type => :bigint, :index => true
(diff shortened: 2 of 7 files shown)


db/migrate/20201111154314_create_physical_storage_consumers.rb
```diff
@@ -0,0 +1,14 @@
+class CreatePhysicalStorageConsumers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :physical_storage_consumers do |t|
+      t.string :name
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :uid_ems
+      t.references :physical_storage, :type => :bigint, :index => true
+      # t.references :addresses, :type => :bigint, :index => true
```
gregoryb-at-ibm, line 9: Remove commented-out code.

db/migrate/20201124063738_create_addresses.rb
```diff
@@ -0,0 +1,12 @@
+
+class CreateAddresses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :addresses do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.references :physical_storage, :type => :bigint, :index => true
```
gregoryb-at-ibm, line 7: An address belongs to a `physical_storage_consumer`, not a `physical_storage`, correct? If that's the case, creating a reference to `physical_storage` is duplicate date, because the storage is determined by the consumer.

db/migrate/20201124063738_create_addresses.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.references :physical_storage_consumer, :type => :bigint, :index => true
+      t.references :port, polymorphic: true, :index => true
```
gregoryb-at-ibm, line 9: What is this `port` reference going to do? What do we expect to link it to?

db/migrate/20201124063738_create_addresses.rb
```diff
@@ -0,0 +1,12 @@
+
+class CreateAddresses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :addresses do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.references :physical_storage_consumer, :type => :bigint, :index => true
```
gregoryb-at-ibm, line 8: If we plan this to be a general address model that can be used to hold various addresses addresses, then the container of the address should be generic. Not hard-coded to `physical_storage_consumer`. Kinda like in `Endpoint`

`t.belongs_to :resource, :polymorphic => true, :type => :bigint`

db/migrate/20201201064001_create_fiber_channel_addresses.rb
```diff
@@ -0,0 +1,9 @@
+class CreateFiberChannelAddresses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :fiber_channel_addresses do |t|
+      t.string :chap_name
+      t.string :chap_secret
```
gregoryb-at-ibm, line 5: chap is for iscsi, not fc.

db/migrate/20201201064001_create_fiber_channel_addresses.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.string :wwpn
```
gregoryb-at-ibm, line 6: Maybe also add `t.string: wwnn`

gregoryb-at-ibm (changes requested):

agrare:
> Physical storage consumers (AKA physical hosts)

This is a little confusing because we have a physical hosts model but that's not what you're talking about right?  As I understood it the storage consumers are any entity which mounts physical storage is that accurate?

> are physical storage entities that hold the information about consumers/hosts that use (defined) the physical storage.

Then this seems to go against that, the physical storage consumers are physical storage entities (aka instances of physical storage)?  And then "use (defined) the physical storage" sound more like the storage system not the storage consumers.

I think a lot of this can be done with existing tables as we map the storage topology for virtualization hosts with the guest_devices, miq_scsi_luns, and miq_scsi_targets tables.  Despite the name I don't think `miq_scsi_*` isn't also used for e.g. FC.  The VMware parser currently parses Block, Fibre, iSCSI, SCSI, and HBA controller_types for vmware hosts.

We can absolutely extend these tables I just don't want to end up with two sets of tables that represent very similar concepts.

llivne:
> > Physical storage consumers (AKA physical hosts)
> 
> This is a little confusing because we have a physical hosts model but that's not what you're talking about right? As I understood it the storage consumers are any entity which mounts physical storage is that accurate?
> 
> > are physical storage entities that hold the information about consumers/hosts that use (defined) the physical storage.
> 
> Then this seems to go against that, the physical storage consumers are physical storage entities (aka instances of physical storage)? And then "use (defined) the physical storage" sound more like the storage system not the storage consumers.
> 
> I think a lot of this can be done with existing tables as we map the storage topology for virtualization hosts with the guest_devices, miq_scsi_luns, and miq_scsi_targets tables. Despite the name I don't think `miq_scsi_*` isn't also used for e.g. FC. The VMware parser currently parses Block, Fibre, iSCSI, SCSI, and HBA controller_types for vmware hosts.
> 
> We can absolutely extend these tables I just don't want to end up with two sets of tables that represent very similar concepts.

ok, I add some typos in the description. Storages holds information about the entities/objects that are connected/mounted to them. in SVC storage system for example it's called "storage hosts", but in miq we think the name may be confusing so we think physical storage consumer may be a good generic name for it. each consumer has information about how it is connected (ISCSI, fc ...) and we think that this information should be kept in an address table, but obviously if we can use an existing table for it we will do it, I didnt find one that fit but I will look on your suggestions (we also looked on the endpoints table in miq but I dont think its the same).

llivne:
@agrare Looks like `MiqScsiTarget` can only belong to `GuestDevice`? We were thinking about an address that could belong to anything polymorphically. A StorageConsumer can have addresses. But so can a PhysicalStorage. And a Host.
I suppose we could link a `GuestDevice` to our `StorageConsumer`s, `PhsysicalStorage`s, etc..
It would a bit odd to link a `GuestDevice` to a `StorageConsumer`, though. As a `StorageConsumer` is a virtual entity that only exists as a set of definitions within the `PhysicalStorage.

agrare:
@llivne can you update the description with example data so that we get an idea what we're dealing with?

llivne:
> @llivne can you update the description with example data so that we get an idea what we're dealing with?

@agrare I added a detailed example data in the description. The implementation in the PR is with polymorphic association (the second diagram in the example data) but we are not sure this is the best choice,  and after brainstorming on this issue we think STI may be better but we waiting for your/miq feedback before continuing with the changes.

agrare:
Thanks @llivne that diagrams helps a lot.  I'd like to try to keep the naming (and architecture for that matter) consistent with standard storage concepts.  It looks like the `physical_storage_consumer` is the "initiator" in this context.  The initiator would then connect to a target port for a LUN.

I think the miq_scsi_targets and miq_scsi_luns does capture this to a certain degree, the "address" for miq_scsi_target is just the simple concatenated string but we could definitely expand on this.

And we don't have to have the miq_scsi_target only connect to guest_devices, I think if we introduced an "initiator" concept (which is what the guest_device is representing in this case) it would help.

Do you need to have the address pulled out into its own model?  Seems like it should belong with the target port.

agrare:
Okay so @llivne schooled me on the storage terminology, I think an STI `storage_address` model would work, and possibly `storage_initiator_groups` to represent the Storage Hosts?

llivne:
> Okay so @llivne schooled me on the storage terminology, I think an STI `storage_address` model would work since it can represent initiator addresses or target addresses, and possibly `storage_initiator_groups` to represent the Storage Hosts?

@agrare, regarding the name 'storage_address',  I think it doesn't fit because we want to use those addresses for more things in the future, not just storage (e.g. real hosts). I think a more generic name is preferred. As for the name suggestion `storage_initiator_groups`, the word initiator may be confusing. I know you dont like the name 'storage_consumers', what about 'storage_connections' ? @gregoryb-at-ibm @ereztheodorou do you have any other suggestions ?

agrare:
> regarding the name 'storage_address', I think it doesn't fit because we want to use those addresses for more things in the future, not just storage (e.g. real hosts)

It will be a SCSI/iSCSI/FC address for that real host correct?  It isn't an address for only storages but rather a storage-type address.  Just "address" is way too generic.

> what about storage_connections ?

I think it should reflect the fact that these are "configured" but not necessarily active.  Storage consumers and Storage connections sound like they are "active" or "current"

gregoryb-at-ibm:
Hi @agrare 
> It will be a SCSI/iSCSI/FC address for that real host correct? It isn't an address for only storages but rather a storage-type address. Just "address" is way too generic.

We thought that eventually we could also add ETH IP and MAC to this model to have a general "address" model that can be used throughout MIQ for representing any kind of networking (maybe merge it with the Endpoint model). Even FC is used for more application than just storage. And sometimes it's beneficial to also keep track of the iSCSI IP addresses, not only their IQNs.

If you at the MIQ team feel strongly about limiting it to just "storage addresses", I think an appropriate name could be "san_address". SAN stands for Storage Area Network, and it is a common industry name for networking that has to do with attaching to a storage array.

@ereztheodorou, What do you think?

gregoryb-at-ibm:
Hi @agrare 

> I think it should reflect the fact that these are "configured" but not necessarily active. Storage consumers and Storage connections sound like they are "active" or "current"

I see what you mean with the name "consumer" sounding too active. 
Storage_initiator seems like a good name since in cases of ISCSI, FC, and NVME the storage-host actually represents an initiator.
Since storage-hosts tend to be grouped into host-groups inside the storage, I think we might want to save the term "storage_initiator_group" for future use in case we should want to reflect that. And "storage_initiator" is shorter, and conveys the idea clearly enough.

Maybe we could consider the name "host_initiator". It's close in meaning but it has the benefit of having the term "host" in it, which is reminiscent of the industry standard term. So this could make the UI more friendly for people from a storage background.

ereztheodorou:
@gregoryb-at-ibm san_addresses is a good name. 
I think storage_initiator* may be confused with initiator ports existing in the strage for storage machine-to-storage machine communication (i.e. replication) 
host_initiator sounds good too.

agrare:
:+1: on san_addresses and host_initiators

llivne:
I also support san_addresses and host_initiators. I will work on it next week and commit the updated code (name refactoring and changing san_addresses to work with STI instead of polymorphic).

gregoryb-at-ibm (approved):

agrare (approved):
LGTM pending the answer to https://github.com/ManageIQ/manageiq/pull/20858/files#r547277590
Take this out of WIP when you're ready for this to be merged

PR #538, opened 2020-12-02
Fix CloudVolume* STI types on older CinderManagers

agrare:
Older `Openstack` `CinderManagers` didn't use the subclassed `ManageIQ::Providers::Openstack::StorageManager::CinderManager` but rather the base `ManageIQ::Providers::StorageManager::CinderManager`.  This caused the previous data migration to miss these records because it was querying on active `Openstack` `CinderManagers`.

This led to:
`[ActiveRecord::SubclassNotFound] The single-table inheritance mechanism failed to locate the subclass: 'ManageIQ::Providers::Openstack::CloudManager::CloudVolumeSnapshot'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't intend it to be used for storing the inheritance class or overwrite CloudVolumeSnapshot.inheritance_column to use another column for that information.`

Follow-up to https://github.com/ManageIQ/manageiq-schema/pull/532

### DIFF
--- a/db/migrate/20201202172818_fix_cinder_manager_cloud_volumes.rb
+++ b/db/migrate/20201202172818_fix_cinder_manager_cloud_volumes.rb
@@ -1,0 +1,57 @@
+class FixCinderManagerCloudVolumes < ActiveRecord::Migration[5.2]
+  class CloudVolume < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudVolumeBackup < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudVolumeSnapshot < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudVolumeType < ActiveRecord::Base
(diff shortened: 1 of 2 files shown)


himdel:
```
== 20201202172818 FixCinderManagerCloudVolumes: migrating =====================
-- Fixing OpenStack CloudVolume STI class
   -> 0.0200s
   -> 26 rows
-- Fixing OpenStack CloudVolumeBackup STI class
   -> 0.0071s
   -> 0 rows
-- Fixing OpenStack CloudVolumeSnapshot STI class
   -> 0.0084s
   -> 7 rows
-- Fixing OpenStack CloudVolumeType STI class
   -> 0.0070s
   -> 0 rows
== 20201202172818 FixCinderManagerCloudVolumes: migrated (0.0428s) ============
```

worked for me, thanks! :+1:

agrare:
https://github.com/ManageIQ/manageiq-providers-openstack/pull/242 changed the class that was created to use the subclassed openstack cinder_manager, however it didn't change the association's class_name https://github.com/mansam/manageiq-providers-openstack/blob/c25ab93cc33ba768f8b6d895cec17593bdf0ed50/app/models/manageiq/providers/openstack/cinder_manager_mixin.rb#L10 and I don't see a data migration associated with it which is why newly created Cinder managers are of the openstack subclass but what @himdel had in his database was the older type.

PR #539, opened 2020-12-07
Data migration to cover edge conditions after control explorer split

h-kataria:
There could be some edge conditions where user's might have old hidden features explicitly mentioned in their user role record, more details/discussions are in https://github.com/ManageIQ/manageiq-api/pull/966#pullrequestreview-542927153

@NickLaMuro @skateman @gtanzillo @Fryguy We only needed to rename the old features in this migration. After looking at changes made in original split PR, it appears that `control_explorer_view` was not directly being used at all and did not have any children under that node so even if user had an access to that feature somehow, it did not do anything. 
Few of features that were deleted in original PR, that were not in use are listed below.
`control_explorer_view`,  `event_admin`, `event_edit`, `policy_profile_assign`

Condition related features remain the same as before.

Original [PR](https://github.com/ManageIQ/manageiq/pull/20438) with control feature split changes

### DIFF
--- a/db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
+++ b/db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
@@ -1,0 +1,64 @@
+class ControlExplorerSplitFeatureMapping < ActiveRecord::Migration[5.2]
+  class MiqProductFeature < ActiveRecord::Base; end
+  class MiqRolesFeature < ActiveRecord::Base; end
+
+  FEATURE_MAPPING = {
+    "control_explorer"       => "control",
+    "policy_profile"         => "miq_policy_set",
+    "policy_profile_view"    => "miq_policy_set_view",
+    "policy_profile_admin"   => "miq_policy_set_admin",
+    "profile_new"            => "miq_policy_set_new",
+    "profile_delete"         => "miq_policy_set_delete",
+    "profile_edit"           => "miq_policy_set_edit",
+    "policy"                 => "miq_policy",
+    "policy_view"            => "miq_policy_view",
+    "policy_admin"           => "miq_policy_admin",
+    "policy_new"             => "miq_policy_new",
+    "policy_copy"            => "miq_policy_copy",
+    "policy_edit"            => "miq_policy_edit",
+    "policy_edit_conditions" => "miq_policy_conditions_assignment",
+    "policy_edit_events"     => "miq_policy_events_assignment",
+    "policy_delete"          => "miq_policy_delete",
+    "event"                  => "miq_event",
+    "action"                 => "miq_action",
+    "action_admin"           => "miq_action_admin",
+    "action_show_list"       => "miq_action_show_list",
+    "action_show"            => "miq_action_show",
+    "action_new"             => "miq_action_new",
+    "action_edit"            => "miq_action_edit",
+    "action_delete"          => "miq_action_delete",
+    "alert_profile"          => "miq_alert_set",
+    "alert_profile_admin"    => "miq_alert_set_admin",
+    "alert_profile_new"      => "miq_alert_set_new",
+    "alert_profile_edit"     => "miq_alert_set_edit",
+    "alert_profile_assign"   => "miq_alert_set_assign",
+    "alert_profile_delete"   => "miq_alert_set_delete",
+    "alert"                  => "miq_alert",
+    "alert_admin"            => "miq_alert_admin",
+    "alert_new"              => "miq_alert_new",
+    "alert_copy"             => "miq_alert_copy",
+    "alert_edit"             => "miq_alert_edit",
+    "alert_delete"           => "miq_alert_delete",
+  }.freeze
+
+  def up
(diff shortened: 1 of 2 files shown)


db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
```diff
@@ -0,0 +1,84 @@
...
+      FEATURE_MAPPING.each do |to, from|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+    end
+  end
+
+  private
+
+  def old_feature_ids
+    FEATURE_MAPPING.keys.map { |identifier| MiqProductFeature.find_by(:identifier => identifier) }
```
gtanzillo, line 82: This looks like it's returning full `MiqProductFeature` instances and not just the ids.

I think this would be better, and it wouldn't hurt to cache them -

`@old_feature_ids ||= MiqProductFeature.where(:identifier => FEATURE_MAPPING.keys).pluck(:id)`

h-kataria: > This looks like it's returning full `MiqProductFeature` instances and not just the ids.
> 
> I think this would be better, and it wouldn't hurt to cache them -
> 
> `@old_feature_ids ||= MiqProductFeature.where(:identifier => FEATURE_MAPPING.keys).pluck(:id)`

@gtanzillo made suggested change

Fryguy:
@h-kataria Can you add specs?

db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def down
+    return if MiqProductFeature.none?
+
+    say_with_time 'Naming features back to match old features before the explorer split' do
+      admin_feature = MiqProductFeature.find_or_create_by!(:identifier => 'control_explorer')
+
+      MiqRolesFeature.where(:miq_product_feature_id => admin_feature.id).each do |feature|
+        old_feature_ids.each do |id|
+          MiqRolesFeature.create!(:miq_product_feature_id => id, :miq_user_role_id => feature.miq_user_role_id)
```
Fryguy, line 69: If we are creating in the up I'd expect a delete in the down, but I don't see any

db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
```diff
@@ -0,0 +1,84 @@
...
+    say_with_time 'Renaming old hidden features under control explorer after the split' do
+      new_feature = MiqProductFeature.find_or_create_by!(:identifier => 'control')
+      user_role_ids_with_old_features = MiqRolesFeature.where(:miq_product_feature_id => old_feature_ids).pluck(:miq_user_role_id).uniq if old_feature_ids
+      user_role_ids_with_old_features.each do |user_role_id|
+        MiqRolesFeature.create!(:miq_product_feature_id => new_feature.id, :miq_user_role_id => user_role_id)
+      end
+
+      # Direct renaming of features
+      FEATURE_MAPPING.each do |from, to|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
```
Fryguy, line 56: I think this will accidentally create 2 `control` features, because it will be created once here and also once in [line 48](https://github.com/ManageIQ/manageiq-schema/pull/539/files#diff-db54a24bab2d11439ff03641a3a985ceced1154461c66785abd258f0cdb1141eR48)

db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
```diff
@@ -0,0 +1,84 @@
...
+  }.freeze
+
+  def up
+    return if MiqProductFeature.none?
+
+    say_with_time 'Renaming old hidden features under control explorer after the split' do
+      new_feature = MiqProductFeature.find_or_create_by!(:identifier => 'control')
+      user_role_ids_with_old_features = MiqRolesFeature.where(:miq_product_feature_id => old_feature_ids).pluck(:miq_user_role_id).uniq if old_feature_ids
+      user_role_ids_with_old_features.each do |user_role_id|
+        MiqRolesFeature.create!(:miq_product_feature_id => new_feature.id, :miq_user_role_id => user_role_id)
```
Fryguy, line 51: I don't understand why this is creating new records, if the purpose is to rename existing records.

db/migrate/20201202164337_control_explorer_split_feature_mapping.rb
```diff
@@ -0,0 +1,70 @@
...
+    say_with_time 'Naming features back to match old features before the explorer split' do
+      FEATURE_MAPPING.each do |to, from|
+        MiqProductFeature.find_by(:identifier => from)&.update!(:identifier => to)
+      end
+    end
+  end
+
+  private
+
+  def old_feature_ids
```
gtanzillo, line 67: Looks like this method is no longer nedded

gtanzillo (approved):

PR #541, opened 2021-01-04
Data Migration for Policy Profile Startpage url update

h-kataria:
If any of the users had their startpage set to Policy Profile explorer screen, this migration sets it to non-explorer version screen of Policy Profile list view.

Core PR https://github.com/ManageIQ/manageiq/pull/20820
UI PR https://github.com/ManageIQ/manageiq-ui-classic/pull/7348

### DIFF
--- a/spec/migrations/20201227173629_update_startup_shortcut_after_policy_profile_deexplorization_spec.rb
+++ b/spec/migrations/20201227173629_update_startup_shortcut_after_policy_profile_deexplorization_spec.rb
@@ -1,0 +1,65 @@
+require_migration
+
+describe UpdateStartupShortcutAfterPolicyProfileDeexplorization do
+  let(:user_stub) { migration_stub :User }
+
+  migration_context :up do
+    describe 'starting page update' do
+      it 'update user start page if miq_policy_set/explorer' do
+        user = user_stub.create!(:settings => {:display => {:startpage => 'miq_policy_set/explorer'}})
+
+        migrate
+        user.reload
+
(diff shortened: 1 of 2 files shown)


Fryguy (approved):
LGTM...should be merge with the dependent PRs.

Fryguy:
Needs a data migration spec.

spec/migrations/20201227173629_update_startup_shortcut_after_policy_profile_deexplorization_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        expect(user.settings[:display][:startpage]).to eq('miq_policy_set/show_list')
+      end
+    end
+
+    it 'does not affect users without settings' do
+      user = user_stub.create!
+
+      migrate
+
+      expect(user_stub.find(user.id)).to eq(user)
```
gtanzillo, line 23: I'm not sure what this is testing? Should it check that a user that has a different start page set would still have the same page set after the migration?

h-kataria: @gtanzillo It just tests if user did not ave any settings set at all remain the unchanged, but i can add another test to test the case you mentioned in your comment

h-kataria: @gtanzillo please re-review

gtanzillo (approved):

PR #552, opened 2021-01-14
[rails6][schema_dumper.rb] Use table_name properly

NickLaMuro:
A change that was introduced in Rails 6:

https://github.com/rails/rails/pull/35890/commits/f185e0ebc9cc18c596eb9cab1a9cd99e26a20500

Caused `column.table_name` to be an invalid call, and thus the schema was not printed out properly.  This wasn't noticed originally because it was silently passing thanks to this rescue statement:

https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/activerecord/lib/active_record/schema_dumper.rb#L165-L169

The error that ended up being displayed in the schema was something like:

```ruby
ActiveRecord::Schema.define(version: 2021_01_06_200226) do

  # These are extensions that must be enabled in order to support this database
  enable_extension "plpgsql"

  # Could not dump table "accounts" because of following NoMethodError #   undefined method `table_name' for
  #<ActiveRecord::ConnectionAdapters::PostgreSQL::Column:0x000055ae1dbe7d70>
  
  # ...
```

In the same Rails commit that removed the `column.table_name`, a `table_name` attribute was added as an attr_accessor to `SchemaDumper` to track that current `table_name` being worked on, so that was used in this fix instead of the previous `table_name` method on `Column`.

* * *

This commit also adds schema_dumper_spec.rb, which is meant to catch issues like this in the future in CI instead of by another user.

Links
-----

* Reported in `gitter` here: https://gitter.im/ManageIQ/manageiq/core?at=5fff178081c55b09c7065620
* Fallout after merging:  https://github.com/ManageIQ/manageiq/issues/19977


NickLaMuro:
@miq-bot assign @Fryguy 
@miq-bot add_reviewer @agrare 

@agrare This should address the issue you brought up in gitter yesterday.

spec/schema/schema_dumper_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    expect(miq_metric_inheritences.count).to eq(metric_list.count + metric_rollup_list.count)
+
+    combined_metric_list.each do |metric_subdivision|
+      has_definition = miq_metric_inheritences.detect { |l| l.include?(metric_subdivision) }
+      expect(metric_subdivision).to be_truthy
+    end
+  end
+
+  it "includes all views" do
+    relavent_schema_lines = schema.lines.grep(/create_miq_metric_view/).join
```
d-m-u, line 57: nit, can you fix this for me please? it's `relevant`

spec/schema/schema_dumper_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  it "includes all id column comments" do
+    id_comment_list  = %w[conversion_hosts miq_tasks vms]
+    id_comment_lines = schema.lines.grep(/^\s*change_column_comment/)
+
+    expect(id_comment_lines.count).to eq(id_comment_list.count)
+
+    id_comment_list.each do |table_name|
+      has_table =  id_comment_lines.detect { |l| l.include?(table_name) }
```
d-m-u, line 29: also nit, i think you've got some extra space here

spec/schema/schema_dumper_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    expect(miq_metric_sequences.count).to eq(metric_list.count + metric_rollup_list.count)
+
+    combined_metric_list.each do |metric_subdivision|
+      has_definition = miq_metric_sequences.detect { |l| l.include?(metric_subdivision) }
+      expect(metric_subdivision).to be_truthy
+    end
+  end
+
+  it "includes all metrics table inheritence" do
+    miq_metric_inheritences = schema.lines.grep(/^\s*add_miq_metric_table_inheritance/)
```
d-m-u, line 46: this is the last one, can this be `inheritances` please?

lib/tasks_private/spec.rake
```diff
@@ -18,8 +18,31 @@ namespace :spec do
   desc "Prepare all specs"
   task :setup => [:initialize, :setup_db]
 
+  desc <<~DESC
+    Generate table list for schema specs
+
+    Generates a list of tables from the `spec/dummy/db/schema.rb` to be used in
+    the `schema_dumper_spec.rb`.
+  DESC
+  task :table_list => "spec:migrations:complete_up" do
```
Fryguy, line 27: Is this task meant to be a "public" task?  Adding a `desc` adds it to rake -T making it public.  If it's expected to be private, prefer a regular comment over `desc`

lib/tasks_private/spec.rake
```diff
@@ -18,8 +18,31 @@ namespace :spec do
...
+    table_list = File.open("spec/support/table_list.txt", "w")
+    File.open("spec/dummy/db/schema.rb").each_line do |line|
+      next unless line =~ /^\s*create_table\s+"([^"]+)"/
+
+      table_list.puts $1
+    end
+  ensure
+    table_list.close
+  end
```
Fryguy, line 36:
```suggestion
  task :table_list => "spec:migrations:complete_up" do
    File.open("spec/support/table_list.txt", "w") do |table_list|
      File.open("spec/dummy/db/schema.rb").each_line do |line|
        next unless line =~ /^\s*create_table\s+"([^"]+)"/

        table_list.puts $1
      end
    end
  end
```

lib/tasks_private/spec.rake
```diff
@@ -18,8 +18,31 @@ namespace :spec do
...
+
   desc "Run all non-migration specs"
   RSpec::Core::RakeTask.new(:non_migration => :initialize) do |t|
     t.pattern = FileList["spec/**/*_spec.rb"].exclude("spec/migrations/**/*_spec.rb")
+                                             .exclude("spec/schema/**/*_spec.rb")
+  end
+
+  desc "Run schema specs"
+  RSpec::Core::RakeTask.new(:schema => [:initialize, "spec:migrations:complete_up"]) do |t|
```
Fryguy, line 45: I don't think we need to depend on `spec:migrations:complete_up`.  It is ok to assume that the database will already be fully upgraded, similar to how we do for `spec:non-migration` and even `spec:migration`.  All we need to ensure is that the database is left in the up state when this spec completes.

lib/tasks_private/spec.rake
NickLaMuro: @Fryguy yeah, it is.  The intent is that it is used to re-generate the `table_list.txt` for when a new migration adds a table.  It is designed so that a developer doesn't have to add it to the file manually, and just re-runs that task.

That said, I was going to add a comment into the spec (which I forgot to do), that explained that if this spec fails, please run this task to re-generate the file and try again.

spec/support/table_list.txt
```diff
@@ -0,0 +1,335 @@
+accounts
```
Fryguy, line 1: Not sure I love authors having to keep this list up to date, but I can't think of a better way to do this.  @agrare Thoughts?

lib/tasks_private/spec.rake
Fryguy: > That said, I was going to add a comment into the spec (which I forgot to do), that explained that if this spec fails, please run this task to re-generate the file and try again.

Perfect.  That was going to be my next suggestion.

Fryguy:
Overall LGTM.  The only one I think I really care about is https://github.com/ManageIQ/manageiq-schema/pull/552#pullrequestreview-569404510 , as introducing it will only add a performance penalty (it takes a long-ish time to run even an empty complete_up).

spec/schema/schema_dumper_spec.rb
```diff
@@ -0,0 +1,69 @@
+describe "SchemaDumper" do
+  SCHEMA_RB_FILE  = File.expand_path("../dummy/db/schema.rb", __dir__)
+  TABLE_LIST_FILE = File.expand_path("../support/table_list.txt", __dir__)
+
+  let(:combined_metric_list) { metric_rollup_list + metric_list }
+  let(:metric_list)          { (0..23).to_a.map { |num| "metrics_#{'%<num>02d' % { :num => num }}" } }
```
Fryguy, line 6: Might be cleaner / easier to read?

```suggestion
  let(:metric_list)          { (0..23).to_a.map { |num| "metrics_#{num.to_s.rjust(2, '0')}" } }
```

NickLaMuro: Well... it was smaller than that prior to `rubocop` requiring the formatted names...

```
'%02d' % num

# v.s.

.to_s.rjust(2, '0')
```

Anyway, sure, I will change

spec/schema/schema_dumper_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  let(:schema)               { File.read(SCHEMA_RB_FILE) }
+
+  it "includes all of the manageiq tables" do
+    create_table_lines = schema.lines.grep(/^\s*create_table/)
+
+    expect(create_table_lines.count).to eq(table_list.count)
+
+    table_list.each do |table_name|
+      has_table = create_table_lines.detect { |l| l.include?(table_name) }
+      expect(has_table).to be_truthy
```
Fryguy, line 18: Super minor, but in my experience, expects in a loop become hard to debug because when it blows up all this will say is "expected false to be true" and you have no idea _which_ table_name was the one responsible.  There are 2 general ways to fix this

1) Add an extra message like `expect(has_table).to be_truthy, "table not found for #{table_name}"
2) Put the loop *outside* of the `it`, and have a separate `it` per table_name, with the table name as part of the `it` description.

NickLaMuro:
> ... as introducing it will only add a performance penalty (it takes a long-ish time to run even an empty complete_up).

@Fryguy Not really...

```console
$ time bundle exec rake spec:schema
** Migrating all the way up
ruby -Irspec-core-3.10.1/lib:rspec-support-3.10.1/lib rspec spec/schema/schema_dumper_spec.rb
......

Finished in 0.16648 seconds (files took 1.63 seconds to load)
6 examples, 0 failures


real    0m10.032s
user    0m4.136s
sys     0m2.290s
```

spec/schema/schema_dumper_spec.rb
Fryguy: For someone that small, I'd be fine without the formatted names.  I think rubocop is wrong 😆

Fryguy:
@NickLaMuro What is is the timing without the migrating up?

Fryguy:
Even so, we don't require ensuring migrated up for any other spec suite across the product ... It is always assumed that when a spec suite runs the database has been migrated (either hear or in any plugin or core), usually though `spec:setup`

NickLaMuro:
$ time bundle exec rspec spec/schema/schema_dumper_spec.rb
......

Finished in 0.16372 seconds (files took 1.63 seconds to load)
6 examples, 0 failures


real    0m2.236s
user    0m1.385s
sys     0m0.834s
```

A bit longer if I actually make the change to remove the prerequisite from rake task and have it run that way:

```console
$ time bundle exec rake spec:schema
ruby -Igems/rspec-core-3.10.1/lib:rspec-support-3.10.1/lib rspec spec/schema/schema_dumper_spec.rb
......

Finished in 0.16921 seconds (files took 1.62 seconds to load)
6 examples, 0 failures


real    0m3.958s
user    0m2.462s
sys     0m1.469s
```

@Fryguy for the record, this penalty is only observed when running the spec in isolation, but when we run this with `rake spec` with the change I have in this PR:

```diff
-task :spec => ["spec:non_migration", "spec:migrations"]
+task :spec => ["spec:non_migration", "spec:migrations", "spec:schema"]
```

`spec:migrations:complete_up` will not run again because it has already run once via `spec:migrations`.  This just ensures if a developer is running this locally, it is run against a fully migrated DB (if say, `spec:migrations:complete_down` was run before it).  If the developer wants to run the individual file separately via `bundle exec rspec`, then they can still do that without the performance penalty, but the requirement to ensure a DB is migrated fully is on them.

* * *

Side note (and I probably should add this to the description), the intent with this is to add a new category to the existing test suite categories:

- `spec:non_migration`: No DB necessary.  Effectively unit tests
- `spec:migrations:up`: Requires a DB.  Runs up migrations, and assumes a starting point of 0 migrations ran.
- `spec:migrations:down`: Requires a DB.  Runs down migrations, and assumes a starting point of all migrations ran.
- **NEW** `spec:migrations:schema`: Requires a DB.  Assumes a starting point of all migrations ran, and tests the final state of all of the migrations.

spec/schema/schema_dumper_spec.rb
NickLaMuro: Yeah, but I am just not going to bother, so I will go with your suggestion.  I just:

a) Forgot how to do it with `rjust`, and didn't look it up
b) Forgot `rubocop` had this rule...

spec/schema/schema_dumper_spec.rb
NickLaMuro: Will probably go with option 1, since it is less "meta-program-y", and at least the error output will be more obvious.

* * *

Sidenote:  The reason I am doing the `.detect` prior is that when trying to do a lookup prior to this against the entire schema string, the `diff` from `rspec` is basically useless, so this is a hopefully slightly more legible solution.

NickLaMuro:
@Fryguy I updated based on most of your changes, but am fine waiting on some feedback from Adam as well.  Let me know what you think.

lib/tasks_private/spec.rake
NickLaMuro: I went with a slightly different solution, by adding the comment as the error message.  Figured it allowed the action items o be seen in travis as well.

Might want to consider some updates to the `README.md` to inform developers in other ways that this will be a new pattern that is needed, but that probably can happen in a followup PR.

NickLaMuro:
@agrare @Fryguy Re-ping

PR #553, opened 2021-01-18
Subclass CinderManagers under Openstack

agrare:
Follow-up to https://github.com/ManageIQ/manageiq-schema/pull/543 but for the CinderManager.

https://github.com/ManageIQ/manageiq/pull/20959

### DIFF
--- a/db/migrate/20210118185402_subclass_cinder_manager_openstack.rb
+++ b/db/migrate/20210118185402_subclass_cinder_manager_openstack.rb
@@ -1,0 +1,19 @@
+class SubclassCinderManagerOpenstack < ActiveRecord::Migration[6.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("Subclass CinderManager under Openstack") do
+      ExtManagementSystem.where(:type => "ManageIQ::Providers::StorageManager::CinderManager")
+                         .update_all(:type => "ManageIQ::Providers::Openstack::StorageManager::CinderManager")
+    end
+  end
+
+  def down
+    say_with_time("Move CinderManager from Openstack") do
+      ExtManagementSystem.where(:type => "ManageIQ::Providers::Openstack::StorageManager::CinderManager")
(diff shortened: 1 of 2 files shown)


db/migrate/20210118185402_subclass_cinder_manager_openstack.rb
```diff
@@ -0,0 +1,19 @@
...
+  def up
+    say_with_time("Subclass CinderManager under Openstack") do
+      ExtManagementSystem.where(:type => "ManageIQ::Providers::StorageManager::CinderManager")
+                         .update_all(:type => "ManageIQ::Providers::Openstack::StorageManager::CinderManager")
+    end
+  end
+
+  def down
+    say_with_time("Move CinderManager from Openstack") do
+      ExtManagementSystem.where(:type => "ManageIQ::Providers::Openstack::StorageManager::CinderManager")
```
kbrock, line 15: lets not change these to generics.
I think the openstack version is the correct version.
-
If something was openstack cinder manager before the up and down, it will end up as generic which isn't quite valid

kbrock (commented):
This looks like it will work as designed.

This is addressing an invalid use case. So there shouldn't be any.
I don't think this is necessary but understand a little why it was done.

kbrock:
ok, was clarified - it was a generic CinderManager before, now it is one specific to openstack.

db/migrate/20210118185402_subclass_cinder_manager_openstack.rb
kbrock: corrected - thanks agrare

agrare: Hm this isn't a bad point, we know for sure going up that any "old" types need to be Openstack subclassed but that isn't true in reverse.

Before https://github.com/ManageIQ/manageiq-providers-openstack/pull/194 they were all `ManageIQ::Providers::StorageManager::CinderManager` but between then and now it is theoretically possible to have a mixture.  Running the down migration today would basically reset CinderManagers back to 2018.

@Fryguy thoughts?  Maybe drop the down migration?

PR #555, opened 2021-01-28
Flatten worker settings

Fryguy:
Depends on https://github.com/ManageIQ/manageiq/pull/20975


Fryguy:
WIP because I still have to do the down migration

db/migrate/20210128194322_flatten_worker_settings.rb
```diff
@@ -0,0 +1,36 @@
+class FlattenWorkerSettings < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base
+    serialize :value
```
Fryguy, line 3: Also, I want to remove this, but it was helpful for making the tests.

PR #556, opened 2021-02-03
don't require thread

d-m-u:
I'm pretty sure we don't need this one either


Fryguy:
I can't seem to get the tests to run for some reason.  Can you please rebase / amend commit and force push a new commit to see if it triggers the tests?

PR #557, opened 2021-02-03
Introducing VolumeMapping

gregoryb-at-ibm:
Part of https://github.com/ManageIQ/manageiq/issues/21018

### DIFF
--- a/db/migrate/20201220133400_create_volume_mappings.rb
+++ b/db/migrate/20201220133400_create_volume_mappings.rb
@@ -1,0 +1,16 @@
+class CreateVolumeMappings < ActiveRecord::Migration[5.2]
+  def change
+    create_table :volume_mappings do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :cloud_volume, :type => :bigint, :index => true
+      t.references :host_initiator, :type => :bigint, :index => true
+
+      t.string :ems_ref
+      t.string :uid_ems
+
+      t.integer :lun
+
+      t.timestamps
+    end
+  end
+end
--- a/spec/support/table_list.txt
+++ b/spec/support/table_list.txt
@@ -123,6 +123,7 @@ hardwares
 host_aggregate_hosts
 host_aggregates
 host_initiators
+volume_mappings
 host_service_groups
 host_storages
 host_switches


agrare:
@gregoryb-at-ibm you state this is internal to block storage mapping, I wonder if we actually have to model this or is it sufficient to have the operation create this mapping if it doesn't already exist.

Also are these items that are exposed through the provider API and have their own unique ID/Ref or do they just join two other items?

gregoryb-at-ibm:
Hi @agrare 

First to answer your question about how we represent the mappings in our provider. Our provider exposes the defined mappings through its API. Each mapping has its own UUID in the provider. We also hold various other metadata in the mapping records.

Regarding the suggestion of not representing mappings in MIQ.
There currently isn't a mechanism in MIQ that would perform volume mapping on the fly as you suggest.  
In the future, when we introduce the feature of host-side volume mounting to MIQ, then we might add an auto-map functionality that would create mapping on the fly. 
But for now we are trying to provide administrators with the ability of mapping volumes to their consumers (and their consumers aren't even necesarily in MIQ).

Also, seeing what mappings exist and being able to perform unmapping is needed in most sites for security compliance.

agrare:
> Our provider exposes the defined mappings through its API. Each mapping has its own UUID in the provider. 

Okay I saw you building the ems_ref but missed the comment [here](https://github.com/ManageIQ/manageiq-providers-autosde/pull/56/files#diff-2e719385b77fb2003bef6d2cccfc9eb4c044886e4895d1b2aa92f527f7794140R33), with that and what you say here this makes more sense

db/migrate/20201220133400_create_volume_mappings.rb
```diff
@@ -0,0 +1,15 @@
+class CreateVolumeMappings < ActiveRecord::Migration[5.2]
+  def change
+    create_table :volume_mappings do |t|
+      t.integer :lun
```
agrare, line 4: It is minor but can you move this below the references and unique values (so after uid_ems before timestamps) and bump the `ems` reference above the `cloud_volume` one?

This keeps the "important" stuff easy to see early in the table, it is impractical to change the layout after

gregoryb-at-ibm: Could you see if I did it like you meant?
Also, maybe it'll help to break the fields into sections using comments?
Maybe something like below?
```
def change
# references needed by all children of EMS
t. reference :ems ...

# custom references
t.string :name...

# Fields needed by all EMS inventory items
...

# Custom fields
...

```

agrare (approved):

db/migrate/20201220133400_create_volume_mappings.rb
agrare: Looks great @gregoryb-at-ibm, not needed for this PR but better than code comments is something that @djberg96 started doing for some tables which is adding column comments which you can actually access from psql and ActiveRecord.  This would help us when we use the schema to auto-build an openapi spec.

PR #558, opened 2021-02-11
Added :type to volume_mapping and host_initiator

gregoryb-at-ibm:
As explained in the PR below, type is required for these records since the providers are expected to subclass them using STI.
https://github.com/ManageIQ/manageiq-providers-autosde/pull/57


agrare (changes requested):
Hey @gregoryb-at-ibm since this migration has already ran it won't actually re-run and add the column.

You have to create a new migration that adds the column to the table.  Probably want to add `:type` to host_initiators while you're at it.

gregoryb-at-ibm:
> 
> 
> Hey @gregoryb-at-ibm since this migration has already ran it won't actually re-run and add the column.
> 
> You have to create a new migration that adds the column to the table. Probably want to add `:type` to host_initiators while you're at it.

Ah, right! Sorry.

agrare (approved):

PR #563, opened 2021-02-25
Add new vmware disk read/write latency metrics

agrare:
Required for: https://github.com/ManageIQ/manageiq-providers-vmware/pull/705

https://github.com/ManageIQ/manageiq-providers-vmware/issues/688


lib/migration_helper.rb
```diff
@@ -51,6 +51,25 @@ def estimate_batch_complete(start_time, progress)
   end
   private :estimate_batch_complete
 
+  #
+  # Views
+  #
+
+  def create_view(name, base = "#{name}_base")
```
Fryguy, line 58: I feel like we should rename this to `create_metrics_view`, only because it's so specific.

db/migrate/20210225190611_add_metrics_disk_read_write_latency.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    remove_disk_latency_metric_columns("metrics_base")
+    remove_disk_latency_metric_columns("metric_rollups_base")
+
+    create_view("metrics")
+    create_view("metric_rollups")
+  end
+
+  def add_disk_latency_metric_columns(table_name)
+    say_with_time("Adding disk read/write latency to #{table_name}") do
```
Fryguy, line 38: No need for `say_with_time` with schema changes as they automatically do that under the covers.

db/migrate/20210225190611_add_metrics_disk_read_write_latency.rb
```diff
@@ -0,0 +1,52 @@
...
+    add_disk_latency_metric_columns("metrics_base")
+    add_disk_latency_metric_columns("metric_rollups_base")
+
+    create_view("metrics")
+    create_view("metric_rollups")
+  end
+
+  def down
+    drop_view("metrics")
+    drop_view("metric_rollups")
```
Fryguy, line 28: wonder if this would be cleaner as `recreate_metrics_views`...something like...

```ruby
recreate_metrics_views do
  remove_disk_latency_metric_columns("metrics_base")
  remove_disk_latency_metric_columns("metric_rollups_base")
end
```

and then under the covers recreate is just

```
# In migration_helper.rb
def recreate_metrics_views
  drop_view("metrics")
  drop_view("metric_rollups")

  yield

  create_metrics_view("metrics")
  create_metrics_view("metric_rollups")
end
```

lib/migration_helper.rb
```diff
@@ -51,6 +51,25 @@ def estimate_batch_complete(start_time, progress)
...
+    execute(<<-SQL)
+      CREATE VIEW #{name} AS SELECT * FROM #{base}
+    SQL
+    execute(<<-SQL)
+      ALTER VIEW #{name} ALTER COLUMN id SET DEFAULT nextval('#{base}_id_seq')
+    SQL
+  end
+
+  def drop_view(name)
```
Fryguy, line 67: This one is generic though, so it can probably stay.

lib/migration_helper.rb
Fryguy: Or maybe split this into generic and non-generic

```ruby
def create_view(name, query, sequence)
    execute("CREATE VIEW #{name} AS #{query}")
    execute("ALTER VIEW #{name} ALTER COLUMN id SET DEFAULT nextval('#{sequence}')"
end

def create_metrics_view(name)
  base = "#{name}_base"
  create_view(name, "SELECT * FROM #{base}", "#{base}_id_seq")
end
```

lib/migration_helper.rb
```diff
@@ -51,6 +51,42 @@ def estimate_batch_complete(start_time, progress)
...
+    execute(<<-SQL)
+      DROP VIEW #{name}
+    SQL
+  end
+
+  def recreate_metrics_views
+    drop_view("metrics")
+    drop_view("metric_rollups")
```
Fryguy, line 79: when we drop the view does it automatically drop the triggers?

Fryguy:
Although I have a comment, merging anyway, as it can always be fixed in a follow up if it's actually an issue.

lib/migration_helper.rb
agrare: Correct:
Before the migration:
```
vmdb_development=# select trigger_name from information_schema.triggers;
       trigger_name       
--------------------------
 metrics_partition
 metric_rollups_partition
(2 rows)
```

After dropping and recreating the view but without creating the triggers:
```
-- execute("      CREATE VIEW metrics AS SELECT * FROM metrics_base\n")
   -> 0.0023s
-- execute("      ALTER VIEW metrics ALTER COLUMN id SET DEFAULT nextval('metrics_base_id_seq')\n")
   -> 0.0006s
-- execute("      CREATE VIEW metric_rollups AS SELECT * FROM metric_rollups_base\n")
   -> 0.0024s
-- execute("      ALTER VIEW metric_rollups ALTER COLUMN id SET DEFAULT nextval('metric_rollups_base_id_seq')\n")
   -> 0.0006s
== 20210225190611 AddMetricsDiskReadWriteLatency: migrated (0.0150s) ==========
$ rails db
vmdb_development=# select trigger_name from information_schema.triggers;
 trigger_name 
--------------
(0 rows)
```

And then after recreating the views and adding the triggers:
```
-- execute("      CREATE VIEW metrics AS SELECT * FROM metrics_base\n")
   -> 0.0026s
-- execute("      ALTER VIEW metrics ALTER COLUMN id SET DEFAULT nextval('metrics_base_id_seq')\n")
   -> 0.0008s
-- execute("      CREATE VIEW metric_rollups AS SELECT * FROM metric_rollups_base\n")
   -> 0.0024s
-- execute("      ALTER VIEW metric_rollups ALTER COLUMN id SET DEFAULT nextval('metric_rollups_base_id_seq')\n")
   -> 0.0005s
-- add_trigger(:insteadof, :metrics, :metrics_partition)
   -> 0.0017s
-- add_trigger(:insteadof, :metric_rollups, :metric_rollups_partition)
   -> 0.0009s
== 20210225190611 AddMetricsDiskReadWriteLatency: migrated (0.0182s) ==========
$ rails db
vmdb_development=# select trigger_name from information_schema.triggers;
       trigger_name       
--------------------------
 metrics_partition
 metric_rollups_partition
(2 rows)
```

PR #568, opened 2021-03-26
Add ems_ref to authentications table.

jaredm-ibm:
Used for storing UUID of SSH keys in IBM Cloud VPC.


agrare (approved):

agrare:
Currently the name is being used as the "unique reference" for ssh_key brought in from inventory refresh [[ref]](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/cloud_manager.rb#L25)

This is far from ideal as generally the name makes a poor unique ref and was likely only done to work within the existing columns on the authentications table.

A far better strategy is to add an ems_ref attribute which can be used to store the native unique ID, and if there are any providers for which the name truly is the unique ref this can be copied to the ems_ref attribute

PR #571, opened 2021-03-30
remove regular expressions from settings

kbrock:
This is part of https://github.com/ManageIQ/manageiq/pull/20907

We are no longer using ruby regular expressions in our settings since these are not compatible with json.

Instead we are using strings with the format of regular expressions and converting into regular expressions where necessary.

There should not be any settings overrides of regular expressions but this is here just in case

### DIFF
--- a/db/migrate/20210330012001_remove_regex_from_settings.rb
+++ b/db/migrate/20210330012001_remove_regex_from_settings.rb
@@ -1,0 +1,12 @@
+class RemoveRegexFromSettings < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base
+    serialize :value
+  end
+
+  def up
+    say_with_time("Changing regexps in configurations") do
+      SettingsChange.where("value LIKE ?", "%!ruby/regexp%")
+                    .update_all("value = REGEXP_REPLACE(value, '!ruby/regexp ','','g')")
+    end
+  end
+end
--- a/spec/migrations/20210330012001_remove_regex_from_settings_spec.rb
+++ b/spec/migrations/20210330012001_remove_regex_from_settings_spec.rb
@@ -1,0 +1,20 @@
+require_migration
+
+describe RemoveRegexFromSettings do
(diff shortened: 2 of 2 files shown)


kbrock:
@miq-bot assign @agrare

db/migrate/20210330012001_remove_regex_from_settings.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveRegexFromSettings < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base
+    serialize :value
+  end
+
+  def change
```
agrare, line 6: @kbrock should this be `up` ?  Don't think this is reversible

spec/migrations/20210330012001_remove_regex_from_settings_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+describe RemoveRegexFromSettings do
+  let(:settings_change_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    it "converts regular expression object to a string" do
+      record = settings_change_stub.create!(
+        :resource_type => "MiqServer",
+        :key           => "/ems/ems_nuage/event_handling/event_groups/addition/critical",
+        :value         => ["/hello/","beautiful",/world/]
```
agrare, line 15: @kbrock is this actually a Regexp object?

kbrock: the first is a string w/ a regex (no change), the second is a string (no change) and the third is a regex (change)

agrare: 🤦 sorry didn't see the last one didn't have `"` around it like the others

kbrock: @agrare do you think I should document this - are the 3 use cases too sneaky?

Fryguy (approved):
LGTM - needs to be merged with the base PR

agrare (approved):

PR #576, opened 2021-05-18
replace miq_set relationships with miq_set_memberships table

NickLaMuro:
**Requires https://github.com/ManageIQ/manageiq/pull/21241**

This is a alternative to https://github.com/ManageIQ/manageiq-schema/pull/536 with some fixes.  Details below:

To mirror things properly in the `:down` of the second migration, it is required to execute 2 queries to add in the records back into the `relationships` table:

- First:   add in the root nodes for the relationships (parents)
- Second:  use the ids from those records to build the ancestry column for the child nodes

This was not done in the previous iteration of this ~~commit~~ PR, but has been updated to do so here.

Extra test cases have been added to ensure we are rebuilding the records properly, and that it only migrates the records we want moved (since I don't trust my SQL).

### DIFF
--- a/db/migrate/20210519204247_move_miq_set_memberships.rb
+++ b/db/migrate/20210519204247_move_miq_set_memberships.rb
@@ -1,0 +1,69 @@
+class MoveMiqSetMemberships < ActiveRecord::Migration[5.0]
+  class Relationship < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled # disable STI
+  end
+
+  class MiqSetMembership < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("Converting set relationships to miq_set_memberships") do
+      # The `ancestry` column holds the id for the `relationship` record that
+      # references the miq_set the parent record contains the miq_set object
+      #
+      # Because these are trivial trees (1 parent with 1 level of children),
+      # the `miq_set_id` can be derived from the `resource_id` of the parent
+      miq_set_id = "(SELECT rel2.resource_id FROM relationships rel2 WHERE rel2.id = rels.ancestry::bigint)"
+
+      connection.execute(<<-SQL.squish)
+        INSERT INTO miq_set_memberships (member_type, member_id, miq_set_id, created_at, updated_at)
+          SELECT rels.resource_type, rels.resource_id, #{miq_set_id}, rels.created_at, rels.updated_at
+          FROM relationships rels
+          WHERE rels.relationship = 'membership' AND rels.ancestry IS NOT NULL
+      SQL
+
+      Relationship.where(:relationship => "membership").delete_all
+    end
+  end
+
+  def down
+    say_with_time("Converting miq_set_memberships to set relationships") do
+      # Find the find the id for the `relationships` record that is from the
+      # 'membership' relationship, `miq_set_id` that matches the `resource_id`,
+      # and has a `ancestry` column value of NULL, all of which should be
+      # created in the first query.
+      ancestry_id = <<~ANCESTRY_ID.squish
+        (
+          SELECT rel2.id::varchar
+          FROM relationships rel2
+          WHERE rel2.relationship = 'membership'
+            AND rel2.resource_type = set_mem.member_type || 'Set'
+            AND rel2.resource_id = set_mem.miq_set_id
+            AND rel2.ancestry IS NULL
+          LIMIT 1
+        )
+      ANCESTRY_ID
+
+      # First Query:   Add root nodes (only if children exist)
+      connection.execute(<<~SQL.squish)
+        INSERT INTO relationships (relationship, resource_type, resource_id, ancestry, created_at, updated_at)
+          SELECT 'membership', set_mem.member_type || 'Set', set_mem.miq_set_id, NULL, set_mem.created_at, set_mem.updated_at
+          FROM miq_set_memberships set_mem
+          WHERE set_mem.miq_set_id IN (
+            SELECT set_mem_2.miq_set_id
+            FROM miq_set_memberships set_mem_2
+            GROUP BY set_mem_2.miq_set_id
+          );
+      SQL
+
+      # Second Query:  Add child nodes (based on root notes for ancestry column)
+      connection.execute(<<~SQL.squish)
+        INSERT INTO relationships (relationship, resource_type, resource_id, ancestry, created_at, updated_at)
+          SELECT 'membership', set_mem.member_type, set_mem.member_id, #{ancestry_id}, set_mem.created_at, set_mem.updated_at
+          FROM miq_set_memberships set_mem;
+      SQL
+
+      MiqSetMembership.delete_all
+    end
+  end
+end
--- a/spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
+++ b/spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
@@ -1,0 +1,133 @@
+require_migration
+
+# rubocop:disable Style/HashSyntax
+describe MoveMiqSetMemberships do
+  let(:rel_stub)        { migration_stub(:Relationship) }
+  let(:membership_stub) { migration_stub(:MiqSetMembership) }
+  let(:set_stub)        { Class.new(ActiveRecord::Base) { self.table_name = "miq_sets" } }
+  let(:item_stub)       { Class.new(ActiveRecord::Base) { self.table_name = "scan_items" } }
+  let(:rp_stub)         { Class.new(ActiveRecord::Base) { self.table_name = "resource_pools" } }
+
+  migration_context :up do
+    it 'converts relationship records to set membership table records' do
+      # Empty Set
+      set1 = set_stub.create!
+      create_rel(set1, resource_type: "ScanItemSet")
+
+      # Single Child
+      set2  = set_stub.create!
+      item1 = item_stub.create!
+
+      set2_rel = create_rel(set2, resource_type: "ScanItemSet")
+      create_rel(item1, ancestry: set2_rel.id.to_s)
+
+      # Multi Child (with shared child)
+      set3  = set_stub.create!
+      item2 = item_stub.create!
+      item3 = item_stub.create!
+      item4 = item_stub.create!
+
+      set3_rel = create_rel(set3, resource_type: "ScanItemSet")
+      create_rel(item1, ancestry: set3_rel.id.to_s)
+      create_rel(item2, ancestry: set3_rel.id.to_s)
+      create_rel(item3, ancestry: set3_rel.id.to_s)
+      create_rel(item4, ancestry: set3_rel.id.to_s)
+
+      # Extra non-MiqSet relations
+      create_resource_pool
+
+      migrate
+
+      membership1 = membership_stub.find_by(:member_id => item1.id, :miq_set_id => set2.id)
+      membership2 = membership_stub.find_by(:member_id => item1.id, :miq_set_id => set3.id)
+      membership3 = membership_stub.find_by(:member_id => item2.id, :miq_set_id => set3.id)
+      membership4 = membership_stub.find_by(:member_id => item3.id, :miq_set_id => set3.id)
+      membership5 = membership_stub.find_by(:member_id => item4.id, :miq_set_id => set3.id)
+
+      expect(membership1).to have_attributes(:miq_set_id => set2.id, :member_type => "ScanItem")
+      expect(membership2).to have_attributes(:miq_set_id => set3.id, :member_type => "ScanItem")
+      expect(membership3).to have_attributes(:miq_set_id => set3.id, :member_type => "ScanItem")
+      expect(membership4).to have_attributes(:miq_set_id => set3.id, :member_type => "ScanItem")
+      expect(membership5).to have_attributes(:miq_set_id => set3.id, :member_type => "ScanItem")
+
+      # Total MiqSetMembership records after migration
+      expect(membership_stub.count).to eq(5)
+      # Remaining Membership relationships after migration
+      expect(rel_stub.where(:relationship => 'membership').count).to eq(0)
+      # Remaining ResourcePool relationships after migration
+      expect(rel_stub.count).to eq(2)
+    end
+  end
+
+  migration_context :down do
+    it 'converts set membership table records to relationship records' do
+      # Empty Set
+      set1 = set_stub.create!
+
+      # Single Child
+      set2  = set_stub.create!
+      item1 = item_stub.create!
+
+      membership_stub.create!(:miq_set_id => set2.id, :member_type => "ScanItem", :member_id => item1.id)
+
+      # Multi Child (with shared child)
+      set3  = set_stub.create!
+      item2 = item_stub.create!
+      item3 = item_stub.create!
+      item4 = item_stub.create!
+
+      membership_stub.create!(:miq_set_id => set3.id, :member_type => "ScanItem", :member_id => item1.id)
+      membership_stub.create!(:miq_set_id => set3.id, :member_type => "ScanItem", :member_id => item2.id)
+      membership_stub.create!(:miq_set_id => set3.id, :member_type => "ScanItem", :member_id => item3.id)
+      membership_stub.create!(:miq_set_id => set3.id, :member_type => "ScanItem", :member_id => item4.id)
+
+      # Extra non-MiqSet relations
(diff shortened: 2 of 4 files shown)


NickLaMuro:
@miq-bot assign @kbrock

@kbrock you have a lot of experience reviewing this from https://github.com/ManageIQ/manageiq-schema/pull/536 so I think you should also be the one to review this.  If there is anything that needed to be updated in the original that I didn't address here, let me know.

cc @Fryguy

kbrock (approved):
This was very close - I think close enough to rip off the bandaid - If @Fryguy doesn't see anything glaring, then I'm happy to merge. (it was based upon his original hypothesis)

kbrock:
the sql is solid. Are we all set to also merge https://github.com/ManageIQ/manageiq/pull/20665 ?

NickLaMuro:
FYI:  Just updated the migration names to be in the 2021 range in stead of dated at 2019.  Makes my life a bit easier when testing this out locally.

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+          WHERE miq_set.miq_set_id IN (
+            SELECT miq_set_id
+            FROM miq_set_memberships miq_set_2
+            GROUP BY miq_set_2.miq_set_id
+          );
+
+        INSERT INTO relationships (relationship, resource_type, resource_id, ancestry, created_at, updated_at)
+          SELECT 'membership', miq_set.member_type, miq_set.member_id, #{ancestry_id}, miq_set.created_at, miq_set.updated_at
+          FROM miq_set_memberships miq_set;
+      SQL
```
Fryguy, line 60: Can we have 2 separate execute statements for these?  Not sure why we need to mash them into a single execute.

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,77 @@
+require_migration
+
+describe MoveMiqSetMemberships do
+  let(:rel_stub)        { migration_stub(:Relationship) }
+  let(:membership_stub) { migration_stub(:MiqSetMembership) }
+  let(:ns_stub)         { Class.new(ActiveRecord::Base) { self.table_name = "miq_ae_namespaces" } }
+  let(:rp_stub)         { Class.new(ActiveRecord::Base) { self.table_name = "resource_pools" } }
+
+  migration_context :up do
+    context "two sets with one member each" do
```
Fryguy, line 10: In the previous PR I had asked Drew if we could have specs for more complicated examples, such as empty sets and sets with multiple children.  Can we add them here as well?

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      it 'converts relationship records to set membership table records' do
+        ns1 = ns_stub.create!
+        ns2 = ns_stub.create!
+        ns1_child = ns_stub.create!
+        ns2_child = ns_stub.create!
+        ns1_rel = create_rel(ns1, nil)
+        ns2_rel = create_rel(ns2, nil)
+        create_rel(ns1_child, ancestry_for(ns1_rel))
+        create_rel(ns2_child, ancestry_for(ns2_rel))
+        create_resource_pool
```
Fryguy, line 20: Why do we have this?  Is this just to prove we _arent_ touching non-membership relationships?

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+      # The `ancestry` column holds the id for the `relationship` record that
+      # references the miq_set the parent record contains the miq_set object
+      #
+      # Because these are trivial trees (1 parent with 1 level of children),
+      # the `miq_set_id` can be derived from the `resource_id` of the parent
+      miq_set_id = "(SELECT rel2.resource_id FROM relationships rel2 WHERE rel2.id = rels.ancestry::bigint)"
+
+      connection.execute(<<-SQL.squish)
+        INSERT INTO miq_set_memberships (member_type, miq_set_id, member_id, created_at, updated_at)
+          SELECT rels.resource_type, #{miq_set_id}, rels.resource_id, rels.created_at, rels.updated_at
```
Fryguy, line 20: Super minor, but readability wise, I think member_type and member_id should be next to each other since they are a pair.

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+          FROM relationships rels
+          WHERE rels.relationship = 'membership' AND rels.ancestry IS NOT NULL
+      SQL
```
Fryguy, line 23: Maybe it's just me, but I find something like the following much more readable:

```ruby
miq_set_ids = Relationship.where(:relationship => "membership", :ancestry => nil).pluck(:id, :resource_id).to_h

Relationship.where(:relationship => "membership").where.not(:ancestry => nil).each do |r|
  MiqSetMembership.create!(
    :miq_set_id  => miq_set_ids[r.ancestry.to_i],
    :member_type => r.resource_type,
    :member_id   => r.resource_id,
    :created_at  => r.created_at,
    :updated_at  => r.updated_at
  )
end
```

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
NickLaMuro: Yes, I wasn't fully confident in the SQL I was writing, so I wanted something that was going to be a sanity check to make sure we didn't delete records we shouldn't be.

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
NickLaMuro: I will look for those comments and see if you have any other suggestions as well.  Since I was re-writing the `:down`, the first pass was to make sure what existed still worked, but I can add stuff now since we have a working copy.

db/migrate/20210519204247_move_miq_set_memberships.rb
NickLaMuro: oh, sure.  Not sure why I did it that way, but I agree.  👍

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+        migrate
+
+        # Ensure we now are back to 4 records.  2 parents and 2 children.
+        parent_rel1 = rel_stub.find_by(:resource_id => ns1.id)
+        parent_rel2 = rel_stub.find_by(:resource_id => ns2.id)
+        child_rel1  = rel_stub.find_by(:resource_id => ns1_child.id)
+        child_rel2  = rel_stub.find_by(:resource_id => ns2_child.id)
+        expect(parent_rel1).to have_attributes(:relationship => "membership", :resource_type => "MiqAeNamespace", :ancestry => nil)
+        expect(parent_rel2).to have_attributes(:relationship => "membership", :resource_type => "MiqAeNamespace", :ancestry => nil)
```
Fryguy, line 51: I don't think this is right?  The root node's resource_type should be the set type, not the type of the child records.  It's confusing in this particular spec because MiqAeNamespace is not an existing set type.  If we change this to, say, ScanItemSet and ScanItem, I'd expect the root nodes to be ScanItemSet and the child nodes to be ScanItem.

db/migrate/20210519204247_move_miq_set_memberships.rb
NickLaMuro: Yes, but it also is a major N+1, and on large datasets I would prefer to make this a single statement where possible.

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+            AND rel2.ancestry IS NULL
+        )
+      ANCESTRY_ID
+
+      # First Query:   Add root node
+      # Second Query:  Add child nodes
+      connection.execute(<<-SQL.squish)
+        INSERT INTO relationships (relationship, resource_type, resource_id, ancestry, created_at, updated_at)
+          SELECT 'membership', miq_set.member_type, miq_set.miq_set_id, NULL, miq_set.created_at, miq_set.updated_at
+          FROM miq_set_memberships miq_set
```
Fryguy, line 50: This is really confusing because we have a table named `miq_sets`, but this is renaming miq_set_memberships to `miq_set`

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
NickLaMuro: That is probably true.  I was going off what was here, but I am completely fine changing the underlying models to something that makes a bit more sense (and is easier to grok.  Will change.

db/migrate/20210519204247_move_miq_set_memberships.rb
NickLaMuro: Whatever, sure, I will change.

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+            AND rel2.resource_id = miq_set.miq_set_id
+            AND rel2.ancestry IS NULL
+        )
+      ANCESTRY_ID
+
+      # First Query:   Add root node
+      # Second Query:  Add child nodes
+      connection.execute(<<-SQL.squish)
+        INSERT INTO relationships (relationship, resource_type, resource_id, ancestry, created_at, updated_at)
+          SELECT 'membership', miq_set.member_type, miq_set.miq_set_id, NULL, miq_set.created_at, miq_set.updated_at
```
Fryguy, line 49: While playing around with making this just be a regular AR method, I think the resource_type is wrong. (see my comment also on the spec).  One convenient thing is I think the Set and their members are named similarly via convention (e.g. ScanItemSet and ScanItem), so we can just tack on "Set" to the member_type and it should work.

db/migrate/20210519204247_move_miq_set_memberships.rb
NickLaMuro: Oh, that is why I didn't remember doing that... that was how this was when I got here...

db/migrate/20210519204247_move_miq_set_memberships.rb
```diff
@@ -0,0 +1,65 @@
...
+          FROM miq_set_memberships miq_set
+          WHERE miq_set.miq_set_id IN (
+            SELECT miq_set_id
+            FROM miq_set_memberships miq_set_2
+            GROUP BY miq_set_2.miq_set_id
+          );
```
Fryguy, line 55: Similarly, I find the following much more readable

```ruby
# Create root nodes
root_nodes = MiqSetMembership.select("DISTINCT ON(miq_set_id) *").order(:miq_set_id, :created_at).map do |m|
  Relationship.create!(
    :relationship  => "membership",
    :ancestry      => nil,
    :resource_type => "#{m.member_type}Set",
    :resource_id   => m.miq_set_id,
    :created_at    => m.created_at,
    :updated_at    => m.updated_at
  )
end.index_by(&:resource_id)

# Create child nodes
MiqSetMembership.each do |m|
  Relationship.create!(
    :relationship  => "membership",
    :ancestry      => root_nodes[m.miq_set_id].id,
    :resource_type => m.member_type,
    :resource_id   => m.member_id,
    :created_at    => m.created_at,
    :updated_at    => m.updated_at
  )
end
    
```

db/migrate/20210519204247_move_miq_set_memberships.rb
Fryguy: There are unlikely no large datasets, so while it is technically an N+1, it will have no practical difference.  See also my comment here (though I admittedly hedged there on us looking at some "real" databases): https://github.com/ManageIQ/manageiq-schema/pull/536#discussion_r524354668

Fryguy: Interesting, I just noticed the original code in #536 was AR based, and was switched _to_ pure SQL...not sure why.

Fryguy:
@NickLaMuro Note that I do realize you didn't write much of this code and are just updating it to get it into a mergable state, so I understand you may not know the original decisions made and why things are the way they are.

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
Fryguy: Ok that's good...I think it's a good idea to keep it in.

db/migrate/20210519204247_move_miq_set_memberships.rb
NickLaMuro: Two things with this:

- I was responding to this comment https://github.com/ManageIQ/manageiq-schema/pull/536#discussion_r530650239 and trying to make it work like the `:up`
- In that form **wrong** because it only re-added the children, and not the parent nodes.  The switch to `miq_set_members` flattens the relationships, so the `:down` needs to put it back in a tree, which it wasn't doing previously.

I wasn't in the business of making this a complete rewrite, just getting it to where it worked properly.

(P.S.  This is also code you originally wrote)

Fryguy: Yeah I'm ok with merging as is if the tests all pass, I just find pure SQL really hard to read

spec/migrations/20210519204247_move_miq_set_memberships_spec.rb
Fryguy: I can't find the comments, but I do recall saying them...it may have happened in gitter or gchat

NickLaMuro: Yeah, I am just doing the following currently (haven't pushed the changes yet):

- 1 empty set
- 1 set with 1 child
- 1 set with 4 children, sharing one with the second

Fryguy: oh good idea on the sharing...i forgot about that.

NickLaMuro:
Going to mark this as `[WIP]` until I can get the related core PR in a working state.

db/migrate/20210519204247_move_miq_set_memberships.rb
kbrock: for this one, they read the same to me.
But for my test database, it took tens of seconds (possibly in the minutes?) vs 100ms

db/migrate/20210519204247_move_miq_set_memberships.rb
kbrock: I remember the down being a little more complicated than I originally thought it would be.
Probably it was the realization that we have relationship records for all sets and relationships.
Since we already know that when approaching this, it ends up being easier. (joys of having the benefit of seeing someone else tackle the problem 3 times)

the fact that these only have 1 layer deep of ancestry makes it much easier for the creation though

kbrock (approved):
so we rebuilt the sql 3 different times to try and make it readable.
And we did a ruby version and built the sql which was like it.

I think this looks great. Thanks for fixing up the remaining issues and rewriting what you needed to do.

NickLaMuro:
@kbrock might make a quick change to fix the `rubocop` errors, but I think it is mostly where it should be.

db/migrate/20210519204247_move_miq_set_memberships.rb
kbrock: I originally wrote this as ruby. then wrote the identical method using sql instead of ruby.

The goal was to actually write the SQL so it was readable to a ruby user rather than write sql that was fast.
Granted the ruby I wrote was intentionally written in a way that would translate to sql (and vice versa)

Not sure where all those ruby methods went.

NickLaMuro: > Not sure where all those ruby methods went.

Uh... I probably rebased over them...

kbrock:
@jrafanie I like the sql here. I like the intent. I'm a definite 👍🏽  here.

I have been too close to this PR over the past 8 months. Writing most of the SQL with d-m-u.
Jason wrote the original PR.
Nick picked up the slack and got it over the finish line.

that is why unfortunately, we have called you (the big guns) in to help us close this one

jrafanie (approved):
LGTM.

PR #578, opened 2021-05-20
[WIP] add ancestry to resource pools

d-m-u:
KB said I could open this as long as I stated clearly that this work relies on me also getting the core work for this done. 

It's the same as the VM one that was opened and merged last year. It requires some work on relationships (doesn't everything, though?)

### DIFF
--- a/db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
+++ b/db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
@@ -1,0 +1,130 @@
+class AddAncestryToResourcePool < ActiveRecord::Migration[5.2]
+  class ResourcePool < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+    has_many :all_relationships, :class_name => "AddAncestryToResourcePool::Relationship", :dependent => :destroy, :as => :resource
+    include ActiveRecord::IdRegions
+  end
+
+  class Relationship < ActiveRecord::Base
+    belongs_to :resource_pool, :class_name => 'AddAncestryToResourcePool::ResourcePool'
+  end
+
+  def up
+    add_column :resource_pools, :ancestry, :string
+    add_index :resource_pools, :ancestry
+
+    say_with_time("set ResourcePool ancestry from existing genealogy relationship resource information") do
+      transfer_relationships_to_ancestry(ResourcePool, 'ResourcePool', 'genealogy')
+    end
+
+    Relationship.where(:relationship => 'genealogy', :resource_type => 'ResourcePool', :resource_id => ResourcePool.all.select(:id)).delete_all
+  end
+
+  def down
+    say_with_time("create relationship records from ResourcePool ancestry") do
+      create_relationships_from_ancestry(ResourcePool.all, Relationship, 'ResourcePool', 'genealogy')
+    end
+
+    remove_column :resource_pools, :ancestry
+  end
+
+  # src is relationship, dest is ResourcePool
+  def transfer_relationships_to_ancestry(dest_model, resource_type, relationship)
+    ancestry_resources = ancestry_resource_ids(relationship, resource_type, dest_model.rails_sequence_range(dest_model.my_region_number))
+    ancestry_sources = ancestry_src_ids(ancestry_resources)
+    new_ancestors = ancestry_of_src_ids_for_src(ancestry_sources)
+    connection.execute(update_src(new_ancestors, dest_model))
+  end
+
+  # src is ResourcePool, dest is relationship
+  def create_relationships_from_ancestry(src_model, dest_model, resource_type, relationship)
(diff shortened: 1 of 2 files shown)


db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def down
+    say_with_time("create relationship records from vm ancestry") do
+      create_relationships_from_ancestry(ResourcePool.all, Relationship, 'ResourcePool', 'genealogy')
+    end
+
+    remove_column :resource_pools, :ancestry
+  end
+
+  # src is relationship, dest is vm
```
kbrock, line 31: in this case, dest is resource pool (copy and paste error)

maybe dest is something to the effect that it is the `model with ancestry column that will get changed`

db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
```diff
@@ -0,0 +1,130 @@
...
+  def transfer_relationships_to_ancestry(dest_model, resource_type, relationship)
+    ancestry_resources = ancestry_resource_ids(relationship, resource_type, dest_model.rails_sequence_range(dest_model.my_region_number))
+    ancestry_sources = ancestry_src_ids(ancestry_resources)
+    new_ancestors = ancestry_of_src_ids_for_src(ancestry_sources)
+    connection.execute(update_src(new_ancestors, dest_model))
+  end
+
+  # src is vm, dest is relationship
```
kbrock, line 39: src is the model with the ancestry column. dest is the relationship that will get created

db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def up
+    add_column :resource_pools, :ancestry, :string
+    add_index :resource_pools, :ancestry
+
+    say_with_time("set vm ancestry from existing genealogy relationship resource information") do
+      transfer_relationships_to_ancestry(ResourcePool, 'ResourcePool', 'genealogy')
+    end
+
+    Relationship.where(:relationship => 'genealogy', :resource_type => 'ResourcePool', :resource_id => ResourcePool.all.select(:id)).delete_all
```
kbrock, line 20: too bad we duplicated this query. once in sql (with that lateral join stuff) and once in active record here.

db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
```diff
@@ -0,0 +1,130 @@
...
+  def transfer_relationships_to_ancestry(dest_model, resource_type, relationship)
+    ancestry_resources = ancestry_resource_ids(relationship, resource_type, dest_model.rails_sequence_range(dest_model.my_region_number))
+    ancestry_sources = ancestry_src_ids(ancestry_resources)
+    new_ancestors = ancestry_of_src_ids_for_src(ancestry_sources)
+    connection.execute(update_src(new_ancestors, dest_model))
+  end
+
+  # src is vm, dest is relationship
+  def create_relationships_from_ancestry(src_model, dest_model, resource_type, relationship)
+    children = src_model.select("ancestry::bigint").where("ancestry NOT LIKE '%/%'")
```
kbrock, line 41: can you remind me/put in a comment what exactly we are doing here?

is this saying all top level nodes?

kbrock (commented):
Could you verify my logic?
feel like every time we revisit, I have to come up with the logic again.

and our algorithm is:

from a model, follow polymorphic relationship `resource` over to the relationship model.
The relationship has the hierarchy in `ancestry`, just with relationship `id` values instead of the model `id`s.
So we want to go from the model to the relationships, get the hierarchy, and convert those ids to the model ids (go to the relationship in the path and convert to the resource id)

Since we are doing a split on the ancestry to get the resource ids, and joining each of those to a relationship, we end up needing to convert the array into table rows, and then reverse that mapping back into an array. that is what all the `UNNEST` (convert array into rows) and `ARRAY_AGG` (aggregate rows into an array) come into play. And since a table does not necessarily have an order, and we want to ensure that we are keeping the ancestry in order, the `ORDINALITY` was added.

Sound about right drew?

db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
d-m-u: | is this saying all top level nodes?
yes

d-m-u:
| Sound about right drew?
also yes

db/migrate/20200912032913_add_ancestry_to_resource_pool.rb
d-m-u: well, in a perfect world we have one more iteration of this code for folders, so if you're invested in me redoing for that, that's certainly your prerogative.

PR #579, opened 2021-05-25
Loosen manageiq-password to < 2 

Fryguy:
Must be merged before https://github.com/ManageIQ/manageiq/pull/21178

### DIFF
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ancestry"
   spec.add_dependency "activerecord-id_regions", "~> 0.3.0"
   spec.add_dependency "linux_admin",             "~> 2.0"
-  spec.add_dependency "manageiq-password",       "~> 0.3"
+  spec.add_dependency "manageiq-password",       "< 2"
   spec.add_dependency "more_core_extensions",    ">= 3.5", "< 5"
   spec.add_dependency "pg"
   spec.add_dependency "pg-pglogical",            "~> 2.1.1"
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,7 +9,7 @@ require File.expand_path("../dummy/config/environment", __FILE__)
 require 'rspec/rails'
 
 require "manageiq/password/rspec_matchers"
-ManageIQ::Password.v2_key = ManageIQ::Password::Key.new
+ManageIQ::Password.key_root = File.expand_path("dummy/certs", __dir__)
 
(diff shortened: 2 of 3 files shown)


Fryguy:
@bdunne Updated.

manageiq-schema.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ancestry"
   spec.add_dependency "activerecord-id_regions", "~> 0.3.0"
   spec.add_dependency "linux_admin",             "~> 2.0"
-  spec.add_dependency "manageiq-password",       "~> 0.3"
+  spec.add_dependency "manageiq-password",       "< 2"
```
bdunne, line 24: This one probably needs to be `~> 1.0` since it depends on the `key=` method which was introduced in v1.0.0

Fryguy: oh good eyes

spec/spec_helper.rb
```diff
@@ -9,7 +9,7 @@
 require 'rspec/rails'
 
 require "manageiq/password/rspec_matchers"
-ManageIQ::Password.v2_key = ManageIQ::Password::Key.new
+ManageIQ::Password.key = ManageIQ::Password::Key.new
```
kbrock, line 12: Is this here because we want to run without a `manageiq/certs/v2_key?
I guess this would run without a manageiq checked out at all.

Wonder if we could setup the v2 key the same way we set it up for all other processes so we would not rely upon this api at all (and we could be loose with the dependency)

Fryguy: This is because schema runs in the context of a dummy app, so something needs to set the key.

> Wonder if we could setup the v2 key the same way we set it up for all other processes so we would not rely upon this api at all 

Not sure I follow - this API _is_ the way we set up the v2 key in other processes (unless I'm missing something)

Fryguy: Or are you suggesting committing a fake v2_key and then use that for test by setting the key root?

kbrock: This is the only spot that uses this code.
So there is probably another way for us to have a dummy v2_key.

Fryguy: Oh that's very interesting - didn't realize `.key=` was literally only used here...I'm good with switching to `.key_root=`

Fryguy: Though to be honest I actually _prefer_ this interface 😉

Fryguy:
Ok, I've changed this to have a loose dependency similar to the other PRs.  This can now be merged independently.

jrafanie (approved):

PR #564, opened 2021-03-16
Data migrations for Ansible Tower De-explorization

h-kataria:
- Data migration to support feature renaming
- Data migration to support updated startpage url
- Data migration to support OOTB user roles updates to add all 3 new de-explorized screens instead of unified Ansible Tower Explorer feature
Core [PR](https://github.com/ManageIQ/manageiq/pull/21108)

PR #581, opened 2021-06-01
Enable ems_storage* feature if ems_block_storage* is enabled

agrare:
As we transition to combined storage managers we need to ensure anyone that had ems_block_storage* features enabled also has the equivalent ems_storage* feature on.

https://github.com/ManageIQ/manageiq-api/pull/1041
https://github.com/ManageIQ/manageiq-ui-classic/issues/7675

### DIFF
--- a/db/migrate/20210601125525_ensure_ems_storage_features.rb
+++ b/db/migrate/20210601125525_ensure_ems_storage_features.rb
@@ -1,0 +1,40 @@
+class EnsureEmsStorageFeatures < ActiveRecord::Migration[6.0]
+  class MiqUserRole < ActiveRecord::Base
+    has_and_belongs_to_many :miq_product_features, :join_table => :miq_roles_features, :class_name => "EnsureEmsStorageFeatures::MiqProductFeature"
+  end
+
+  class MiqProductFeature < ActiveRecord::Base
+    has_and_belongs_to_many :miq_user_roles, :join_table => :miq_roles_features, :class_name => "EnsureEmsStorageFeatures::MiqUserRole"
+  end
+
+  BLOCK_STORAGE_FEATURES = %w[ems_block_storage ems_block_storage_view ems_block_storage_show_list
+                              ems_block_storage_show ems_block_storage_timeline ems_block_storage_control
+                              ems_block_storage_tag ems_block_storage_protect ems_block_storage_refresh
+                              ems_block_storage_admin ems_block_storage_new ems_block_storage_edit
+                              ems_block_storage_delete].freeze
+  OBJECT_STORAGE_FEATURES = %w[ems_object_storage ems_object_storage_view ems_object_storage_show_list
+                               ems_object_storage_show ems_object_storage_timeline ems_object_storage_control
+                               ems_object_storage_tag ems_object_storage_protect ems_object_storage_refresh
+                               ems_object_storage_admin ems_object_storage_delete].freeze
+  def up
+    return if MiqProductFeature.none?
+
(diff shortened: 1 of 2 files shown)


db/migrate/20210601125525_ensure_ems_storage_features.rb
```diff
@@ -0,0 +1,35 @@
+class EnsureEmsStorageFeatures < ActiveRecord::Migration[6.0]
+  class MiqUserRole < ActiveRecord::Base
+    has_and_belongs_to_many :miq_product_features, :join_table => :miq_roles_features, :class_name => "EnsureEmsStorageFeatures::MiqProductFeature"
```
Fryguy, line 3: If I recall, using a habtm ends up creating the intermediate join table class on the fly, but I can't remember if it can affect "real" updates. Do we use habtm elsewhere in migrations?

db/migrate/20210601125525_ensure_ems_storage_features.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  class MiqProductFeature < ActiveRecord::Base
+    has_and_belongs_to_many :miq_user_roles, :join_table => :miq_roles_features, :class_name => "EnsureEmsStorageFeatures::MiqUserRole"
+  end
+
+  def up
```
Fryguy, line 10: We've been quick-gating these kinds of data migration prs with a `return if MiqProductFeautre.none?` to avoid the empty database going through all of this unnecessarily

Fryguy:
One thing we should avoid is giving the user the individual child feature as well as the parent feature.  If you have the parent feature it is implied you get _all_ of the child features so those should _not_ be set.  Otherwise what happens is that if we ever add a new feature under that parent feature, then users with the parent feature suddenly won't have the new feature, which is unexpected.

Fryguy:
Looking at the features in https://github.com/ManageIQ/manageiq-api/pull/1041/, are these just renames?

db/migrate/20210601125525_ensure_ems_storage_features.rb
agrare: I pulled this from https://github.com/ManageIQ/manageiq-schema/pull/564/files#diff-0572029e3d6c232fd7472ce434f3577bd44d06628727cbce2b250f3cf80b6f84R2-R4 which does similar updates to this

agrare:
> Looking at the features in ManageIQ/manageiq-api#1041, are these just renames?

Not really, these features exist already so this just ensures that if someone has the block storage features that they also have the base ems_storage feature.

> One thing we should avoid is giving the user the individual child feature as well as the parent feature. 

This is a one-for-one replace so if they have the parent ems_block_storage feature they'll get the parent ems_storage feature as well.

Fryguy:
> This is a one-for-one replace so if they have the parent ems_block_storage feature they'll get the parent ems_storage feature as well.

Ah ok now i understand - this is effectively mirroring the existing features into a sister set of features, and you're making sure the user now has both

agrare:
Yes exactly, I think this sums it up https://github.com/ManageIQ/manageiq-ui-classic/pull/7746#issuecomment-853875672

Fryguy (dismissed):

Fryguy:
I just noticed this doesn't have a down feature - should it? At the very least, we probably shouldn't leave the new features in the table I'd think.

agrare:
Yeah I should have mentioned this, I considered that but the issue is that these aren't new features they're existing features.  We can't tell if a user had the base storage features enabled before or not.

PR #506, opened 2020-09-02
[WIP] Collapse initial migration up to Ivanchuk release

Fryguy:
To aid in review, the meat of the changes are here: https://github.com/ManageIQ/manageiq-schema/pull/506/commits/7651e200dd7c387a752f517a39981283989ba83f#diff-61a170ce033e5153604509eea2066607

Performance-wise
- Travis spec timings go from `~5m23s` to `~1m36s` (per Ruby version, `~71%` faster)
- `time bundle exec rake db:migrate` from an empty DB goes from `19.75s` to `11.38s` (`~43%` but may be significantly more on systems with slower IO)

---

This PR moves the collapsed initial migration to 20190726204302 which is the last common migration from the Ivanchuk release.  See https://github.com/ManageIQ/manageiq-schema/tree/ivanchuk/db/migrate.

This migration was chosen as it is the common point across ivanchuk, jansa, and master.

schema_ivanchuk txt, schema_master txt, schema_jansa txt - SourceGear DiffMerge 2020-09-02 15-40-29

Additionally, this aligns well to both upstream and downstream (specifically Red Hat CloudForms) releases.  That is, downstrem CloudForms 5.0 == ivanchuk and is a major version upgrade due to the bump from RHEL 7->8 and Postgres 9->10 upgrades.  This means that those customers must upgrade to 5.0 first before upgrading beyond that.  Therefore their database is guaranteed to be at the ivanchuk level.  In upstream, ivanchuk is partly a major upgrade due to the postgres bump, but jansa is effectively the major version upgrade point due to the appliance bump from CentOS 7 to 8.  Therefore, all users must go through at least jansa (and therefore are guanranteed to have gone through ivanchuk as well).

To summarize, with those considerations in mind, I've chosen this common point in ivanchuk for the collapse, which will land in the kasparov release. Upstream and downstream must go through a prior version (cf 5.0/ivanchuk or jansa), and are guaranteed to have this migration in their database.

---

WIP because I still want to:
- [x] Verify assumptions about upgrade paths for upstream and downstream, particularly if converting to podified
- [x] Super mega triple check all the database objects
- [x] Add a guard to ensure that if someone _tries_ to upgrade without first having gone through jansa, we catch that (this can be checked by just ensuring the database is empty at the start of the collapsed migration)
- [ ] See if there are any other spec/support files or methods that can be removed.
- [ ] Run a giant cross-repo test
- [ ] Need to verify that the seeding of the replication tables is working correctly both from scratch and on upgrade.
[...]

PR #584, opened 2021-07-06
Collapse initial migration up to Gaprindashvili release

### LINKED ISSUE #489: Update collapsed initial migration
With the release of Jansa there are many appliance-level changes which makes it effectively a major upgrade.  What this means is users will be forced to upgrade first to ivanchuk latest before continuing to upgrade to Jansa.  As such, we should take advantage of this to collapse the initial migration to improve migration times.  If we don't do it at this Jansa inflection point, it will be more complicated to do it later.

cc @chessbyte

Fryguy:
Replaces #506
Closes #489

@jrafanie Please review.  This is much easier to review by only looking at the last 4 commits.

---

This PR moves the collapsed initial migration to 20171030131403 which is the last common migration from the gaprindashvili release.  See https://github.com/ManageIQ/manageiq-schema/tree/gaprindashvili/db/migrate.

This migration was chosen as it is the common point across gaprindashvili, hammer, and master.  It also conveniently comes right before the addition of the schema_migrations_ran table and so should not affect replication scenarios.

gaprindashvili_txt__master_txt__hammer_txt_-_SourceGear_DiffMerge

---

- [x] Super mega triple check all the database objects - This was verified with [pgAdmin4's schema diff tool](https://www.pgadmin.org/docs/pgadmin4/latest/schema_diff.html) and no changes were detected in any of these scenarios
  pgAdmin_4
  - [x] migrate to 2017 with master == migrate to 2017 with collapsed
  - [x] full migrate from 0 with master == full migrate from 0 with collapsed
  - [x] migrate to 2017 with master, then full migrate with collapsed == full migrate from 0 with master
- [x] Add a guard to ensure that if someone _tries_ to upgrade without first having gone through gaprindashvili, we catch that (this can be checked by just ensuring the database is empty at the start of the collapsed migration)
- [x] Run a giant cross-repo test

The following do not apply, since this collapse occurs before the schema_migrations_ran is created:
- Need to verify that the seeding of the replication tables is working correctly both from scratch and on upgrade.
- Need to verify that schema_migrations_ran works properly.

Followup PRs:
- See if there are any other spec/support files or methods that can be removed.


Fryguy:
@miq-bot cross-repo-test /all

Fryguy:
@jrafanie I think this is ready to go.

jrafanie:
✂️ 🔥 ✂️ 🔥 ✂️ 🔥 

🍰  🍪 👏 🙇 😍 🎉

PR #587, opened 2021-07-18
Update IBM Cloud 'vendor' attribute to 'ibm_cloud'

jaywcarman:
Since there are multiple IBM providers in ManageIQ (IBM Cloud, IBM
PowerVC) more descriptive 'vendor' values are needed to differentiate.

### DIFF
--- a/db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
+++ b/db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
@@ -1,0 +1,24 @@
+class UpdateIbmVendorToIbmCloud < ActiveRecord::Migration[6.0]
+  class Vm < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time('Updating vendor to "IBM Cloud" for VPC managers') do
+      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
+    end
+    say_with_time('Updating vendor to "IBM Cloud" for PowerVirtualServer managers') do
+      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
+    end
+  end
+
+  def down
+    say_with_time('Reverting vendor to "IBM" for VPC managers') do
+      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm').update_all(:vendor => 'ibm')
+    end
+    say_with_time('Reverting vendor to "IBM" for PowerVirtualServer managers') do
+      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm').update_all(:vendor => 'ibm')
+    end
+  end
+end
--- a/db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
+++ b/db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
@@ -1,24 +1,25 @@
 class UpdateIbmVendorToIbmCloud < ActiveRecord::Migration[6.0]
-  class Vm < ActiveRecord::Base
+  class VmOrTemplate < ActiveRecord::Base
     include ActiveRecord::IdRegions
     self.inheritance_column = :_type_disabled
+    self.table_name = 'vms'
   end
 
   def up
     say_with_time('Updating vendor to "IBM Cloud" for VPC managers') do
-      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
+      VmOrTemplate.in_my_region.where(:type => %w[ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm ManageIQ::Providers::IbmCloud::VPC::CloudManager::Template]).update_all(:vendor => 'ibm_cloud')
     end
     say_with_time('Updating vendor to "IBM Cloud" for PowerVirtualServer managers') do
-      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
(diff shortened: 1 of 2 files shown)


db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
```diff
@@ -0,0 +1,24 @@
+class UpdateIbmVendorToIbmCloud < ActiveRecord::Migration[6.0]
+  class Vm < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time('Updating vendor to "IBM Cloud" for VPC managers') do
+      Vm.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
```
jaywcarman, line 9: I think I also need to update the template (image) records, but `VPC::CloudManager` and `PowerVirtualServers::CloudManager` don't subclass `VmOrTemplate` (https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/vm_or_template.rb#L1). Should I update `Template` records separately?

Fryguy: I have honestly never seen that particular VmOrTemplate class before and it's blowing my mind 🤯 

Even so, in migrations we ignore the STI type (via line 4), so you will need to add both types to the :type clause in the query.
```suggestion
      Vm.in_my_region.where(:type => ['ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm', 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm']).update_all(:vendor => 'ibm_cloud')
```

spec/migrations/20210718074605_update_ibm_vendor_to_ibm_cloud_spec.rb
```diff
@@ -0,0 +1,29 @@
+require_migration
+
+RSpec.describe UpdateIbmVendorToIbmCloud do
+  let(:vm_stub) { migration_stub(:Vm) }
+
+  migration_context :up do
+    it "updates vendor to 'ibm_cloud' for IBM Cloud managers" do
+      vpc_vm = vm_stub.create!(:type => "ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm", :vendor => "ibm")
+      powervs_vm = vm_stub.create!(:type => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm", :vendor => "ibm")
```
Fryguy, line 9: Can you also add a _non_ VPC Vm, and also verify that it _doesn't_ get changed post-migration?

db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
jaywcarman: > Even so, in migrations we ignore the STI type (via line 4), so you will need to add both types to the :type clause in the query.

I found an existing migration and tried to do something similar - will this work?
https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20200607025146_add_ancestry_to_vm.rb#L2

spec/migrations/20210718074605_update_ibm_vendor_to_ibm_cloud_spec.rb
jaywcarman: I added checks for both OpenStack and PowerVC type to ensure they don't get changed (seemed like the most likely type to have something IBM related and thus be inadvertently changed).

db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
agrare: What you have here https://github.com/ManageIQ/manageiq-schema/pull/587/files#diff-36fa40d8b3ea49df31e7c2b95ed2a9b80bd6fc326a6d673850146a45693925cbR2-R6 looks good

db/migrate/20210718074605_update_ibm_vendor_to_ibm_cloud.rb
```diff
@@ -0,0 +1,25 @@
+class UpdateIbmVendorToIbmCloud < ActiveRecord::Migration[6.0]
+  class VmOrTemplate < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+    self.table_name = 'vms'
+  end
+
+  def up
+    say_with_time('Updating vendor to "IBM Cloud" for VPC managers') do
+      VmOrTemplate.in_my_region.where(:type => 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm').update_all(:vendor => 'ibm_cloud')
```
agrare, line 10: Think @Fryguy mentioned this but you'll need to update templates as well:
```
VmOrTemplate.in_my_region.where(:type => ['ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm', 'ManageIQ::Providers::IbmCloud::VPC::CloudManager::Template']).update_all(:vendor => 'ibm_cloud')
```

jaywcarman: I _think_ I've got it right now. I also added `::Template` checks to the specs.

spec/migrations/20210718074605_update_ibm_vendor_to_ibm_cloud_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      vpc_template = vm_stub.create!(:type => "ManageIQ::Providers::IbmCloud::VPC::CloudManager::Template", :vendor => "ibm")
+
+      powervs_vm = vm_stub.create!(:type => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm", :vendor => "ibm")
+      powervs_template = vm_stub.create!(:type => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Template", :vendor => "ibm")
+
+      powervc_vm = vm_stub.create!(:type => "ManageIQ::Providers::IbmPowerVc::CloudManager::Vm", :vendor => "ibm_power_vc")
+      powervc_template = vm_stub.create!(:type => "ManageIQ::Providers::IbmPowerVc::CloudManager::Template", :vendor => "ibm_power_vc")
+
+      openstack_vm = vm_stub.create!(:type => "ManageIQ::Providers::Openstack::CloudManager::Vm", :vendor => "ibm")
+      openstack_template = vm_stub.create!(:type => "ManageIQ::Providers::Openstack::CloudManager::Template", :vendor => "ibm")
```
agrare, line 18: Minor (doesn't impact the test but looks strange) the the `:vendor` for these two would be `"openstack"`

jaywcarman: Fixed!

agrare (approved):

PR #588, opened 2021-07-22
Remove rhn and rhms/satellite columns

kbrock:
part of https://github.com/ManageIQ/manageiq/issues/5547

rhn went away in 2017 and we are no longer supporting satellite:

- Removing the columns that were used by the update manager and the subscription manager
- remove rhn and proxy authentication credentials
- remove setting for `update_repo_name`

### DIFF
--- a/db/migrate/20210810191933_remove_update_repo_names_from_settings.rb
+++ b/db/migrate/20210810191933_remove_update_repo_names_from_settings.rb
@@ -1,0 +1,11 @@
+class RemoveUpdateRepoNamesFromSettings < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base
+  end
+
+  def up
+    say_with_time("deleting update_repo_names that are no longer used") do
+      SettingsChange.where(:resource_type => "MiqRegion", :key => "/product/update_repo_names")
+                    .destroy_all
+    end
+  end
+end
--- a/db/migrate/20210720180318_remove_rh_subscription_columns.rb
+++ b/db/migrate/20210720180318_remove_rh_subscription_columns.rb
@@ -1,0 +1,43 @@
+class RemoveRhSubscriptionColumns < ActiveRecord::Migration[6.0]
+  def up
+    change_table "miq_servers" do |t|
+      t.remove "rh_registered"
+      t.remove "rh_subscribed"
+      t.remove "last_update_check"
+      t.remove "updates_available"
+      t.remove "upgrade_status"
+      t.remove "upgrade_message"
+    end
+
+    change_table "miq_databases" do |t|
+      t.remove "cfme_version_available"
+      t.remove "postgres_update_available"
+      t.remove "registration_type"
+      t.remove "registration_organization"
+      t.remove "registration_server"
+      t.remove "registration_http_proxy_server"
+      t.remove "registration_organization_display_name"
+    end
+  end
+
+  def down
+    change_table "miq_databases" do |t|
+      t.string "cfme_version_available"
+      t.boolean "postgres_update_available"
+      t.string "registration_type"
+      t.string "registration_organization"
+      t.string "registration_server"
+      t.string "registration_http_proxy_server"
+      t.string "registration_organization_display_name"
+    end
+
+    change_table "miq_servers" do |t|
+      t.boolean "rh_registered"
+      t.boolean "rh_subscribed"
+      t.string "last_update_check"
+      t.boolean "updates_available"
+      t.string "upgrade_status"
+      t.string "upgrade_message"
+    end
+  end
+end
--- a/db/migrate/20210726194608_delete_rhn_authentications.rb
+++ b/db/migrate/20210726194608_delete_rhn_authentications.rb
@@ -1,0 +1,12 @@
+class DeleteRhnAuthentications < ActiveRecord::Migration[6.0]
+  class Authentication < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    say_with_time("removing authentication records for Red Hat Network") do
+      Authentication.where(:authtype => %w[registration registration_http_proxy])
+                    .destroy_all
+    end
+  end
+end
--- a/spec/migrations/20210726194608_delete_rhn_authentications_spec.rb
+++ b/spec/migrations/20210726194608_delete_rhn_authentications_spec.rb
@@ -1,0 +1,25 @@
+require_migration
+
+describe DeleteRhnAuthentications do
+  let(:authentication_stub) { migration_stub(:Authentication) }
+
+  migration_context :up do
+    it "deletes" do
+      authentication_stub.create!(:authtype => "registration")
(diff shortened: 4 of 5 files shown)


kbrock:
WIP: need to properly remove the rows associated with the subscription manager

db/migrate/20210722111142_remove_rhsm_columns.rb
```diff
@@ -0,0 +1,41 @@
...
+  def up
+    change_table "miq_databases" do |t|
+      t.remove "registration_type"
+      t.remove "registration_organization"
+      t.remove "registration_server"
+      t.remove "registration_http_proxy_server"
+    end
+
+    Authentication.where(:authtype => %w[registration registration_http_proxy])
+                  .destroy_all
```
Fryguy, line 15: This should be wrapped in a `say_with_time`.  Also, I'm thinking this should be moved into a separate migration so that it can be tested independently.

db/migrate/20210722111142_remove_rhsm_columns.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    change_table "authentications" do |t|
+      t.remove "rhsm_sku"
+      t.remove "rhsm_pool_id"
+      t.remove "rhsm_server"
+      #t.remove "manager_ref"
```
Fryguy, line 21: This column should stay because it is for provider private key pairs. cc @agrare

db/migrate/20210722111142_remove_rhsm_columns.rb
Fryguy: Also this needs a spec.

Fryguy:
@agrare @bdunne please also review.

db/migrate/20210722111142_remove_rhsm_columns.rb
```diff
@@ -0,0 +1,41 @@
...
+      t.remove "registration_http_proxy_server"
+    end
+
+    Authentication.where(:authtype => %w[registration registration_http_proxy])
+                  .destroy_all
+
+    change_table "authentications" do |t|
+      t.remove "rhsm_sku"
+      t.remove "rhsm_pool_id"
+      t.remove "rhsm_server"
```
bdunne, line 20: These are unrelated.  They were added in https://github.com/ManageIQ/manageiq/pull/7779/files#diff-49eb1e9f4c7c64a5250f45cd08b58cd7e0832c08f18c26e06082a972b8d183e7R29-R31

bdunne: Although it looks like the feature requiring them was removed in https://github.com/ManageIQ/manageiq/pull/18515/files#diff-031eafbc5af9f5092abc3e9c07627e482c156a37c0c9e5ff798089e0e61b0464L1

Fryguy: Oh that's right - this is related to ovirt authetnications. @agrare can you chime in here?

agrare: I think that's actually code for deploying openshift via automate (that's pretty cool actually)

Wasn't aware that existed before, but no it is definitely no longer used

db/migrate/20210722111142_remove_rhsm_columns.rb
agrare: :+1: when we collect ssh keys from a provider we use this to store the provider native unique id, definitely still used

db/migrate/20210722111142_remove_rhsm_columns.rb
Fryguy: If that's the case can _all_ of these go away? https://github.com/ManageIQ/manageiq/pull/7779/files#diff-49eb1e9f4c7c64a5250f45cd08b58cd7e0832c08f18c26e06082a972b8d183e7R6-R31

agrare: Maybe?  I didn't check that none of those are used anymore.

Either way probably should be pulled out of this PR since it isn't related to rhn/etc...

Fryguy: > Maybe? I didn't check that none of those are used anymore.

They all went in at the same time for the same feature, so if the feature is gone, I assumed they all should go

> Either way probably should be pulled out of this PR since it isn't related to rhn/etc...

Totally agree - doesn't belong in this PR.

kbrock: Can't remember who asked me to delete `RegistrationSystem`  "while I was at it"

So the goal is now to separate out the `UpdateSystem` from the `RegistrationSystem` and actually grow the `RegistrationSystem` (satellite) PR to delete more fields?

db/migrate/20210722111142_remove_rhsm_columns.rb
```diff
@@ -0,0 +1,34 @@
+class RemoveRhsmColumns < ActiveRecord::Migration[6.0]
+  def up
+    change_table "miq_databases" do |t|
+      t.remove "registration_type"
+      t.remove "registration_organization"
+      t.remove "registration_server"
+      t.remove "registration_http_proxy_server"
```
bdunne, line 7: I think these columns belong in the migration above

db/migrate/20210722111142_remove_rhsm_columns.rb
Fryguy: There are three completely different "satellite/registration" things in our product that are unrelated to each other.

1. Our appliance had the ability to update the RPMs on the appliance itself
2. The Satellite/Foreman provider is used for managing satellites
3. The OpenShift provider had a capability to be able to deploy new openshifts, and part of that effort was the need to store various credentials for that purpose (including credentials to an rhsm system)

This PR should focus only on number 1. The `rhsm_` columns in the authentications table are part of number 3.  I can't recall offhand which number `RegistrationSystem` belongs to, but I _thought_ it was number 1 ...perhaps @bdunne knows better.

bdunne: Yeah, `RegistrationSystem` is part of 1

kbrock:
fixed copy paste and whitespace issues.

These new changes should be more focused on the columns that are just related to the registration system and update management

bdunne (approved):
👍🏻 LGTM

bdunne:
Also, I opened https://github.com/ManageIQ/manageiq-style/pull/25 to take care of the rubocop comment

db/migrate/20210810191933_remove_update_repo_names_from_settings.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveUpdateRepoNamesFromSettings < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
```
agrare, line 3: `SettingsChange` isn't an STI model so this isn't needed

agrare (approved):

Issue #479, opened 2020-05-13
The `rails generate migration` command should also generate a spec file


skateman:
This woule speed up things a little bit, maybe we should make it just optional.

cc @Fryguy

Fryguy:
I'm not sure about this one, since we only require specs for data migrations, and most migrations are just normal schema migrations.

skateman:
Maybe an optional one?

Fryguy:
I could see an optional argument during generation, yeah.  Do you know how to modify existing Rails generators?

skateman:
Not yet :laughing: will get back to you tomorrow :stuck_out_tongue:

kbrock:
generating a file with an up/down and having the correct name and spec name would be very nice
that is where my time is spent

PR #592, opened 2021-07-27
Add spec generator for migration generator

NickLaMuro:
Overwrites `rails generate migration` to allow a spec file to be generated when a migration is generated with the same file name.

QA
--

To test, you can run something like the following:

```ruby
$ bin/rails g migration foo_bar_baz -p
      create  db/migrate/20210727212808_foo_bar_baz.rb
      create  spec/migrations/20210727212808_foo_bar_baz_spec.rb
```

And just validate that the file names are as you would expect.  To validate the files are created properly, just remove the `-p` argument:

```ruby
$ bin/rails g migration test_two
      create  db/migrate/20210727213008_test_two.rb
      create  spec/migrations/20210727213008_test_two_spec.rb
```

Closes https://github.com/ManageIQ/manageiq-schema/issues/479

### DIFF
--- a/lib/generators/migration/migration_generator.rb
+++ b/lib/generators/migration/migration_generator.rb
@@ -1,0 +1,52 @@
+require 'rails/generators/active_record/migration/migration_generator'
+
+# Overwrites the default `ActiveRecord::Generators::MigrationGenerator` to
+# include our own spec file for each migration generated.
+#
+module ManageIQ::Schema
+  class MigrationGenerator < ActiveRecord::Generators::MigrationGenerator
+    # Using both `source_root` (standard) and `source_paths` here so that we
+    # can default to the rails sources for the `migration.rb` template, but
+    # then use our own for the `migration_spec.rb` file.
+    source_root ActiveRecord::Generators::MigrationGenerator.source_root
+    source_paths << File.expand_path('templates', __dir__)
+
+    # Overwrite the defaulted generated .namespace method so that this class is
+    # found as the `migration` class.
+    #
+    # Some Thor/Rails magic happeninghere ... but basically it allows this to
+    # overwrite and work properly
+    #
+    def self.namespace(name = nil)
+      return super if name
+
+      "migration"
+    end
+
+    def self.miq_schema_root
+      ManageIQ::Schema::Engine.root
+    end
+
+    # Overwrites the default method to do the original code, but then also
+    # generate the spec file.
+    def create_migration_file
+      super
+
+      migration_template("migration_spec.rb", File.join(spec_dir, "#{file_name}_spec.rb"))
+    end
+
+    private
+
+    def db_migrate_path
+      self.class.miq_schema_root.join("db", "migrate")
+    end
+
+    def spec_dir
+      self.class.miq_schema_root.join("spec", "migrations")
+    end
+
+    def spec_class_name
+      migration_class_name.sub(/Spec$/, "")
+    end
+  end
+end
--- a/lib/generators/migration/templates/migration_spec.rb.tt
+++ b/lib/generators/migration/templates/migration_spec.rb.tt
@@ -1,0 +1,39 @@
+require_migration
+
+# This is mostly necessary for data migrations, so feel free to delete this
+# file if you do no need it.
+describe <%= spec_class_name %> do
+  # let(:my_model_stub) { migration_stub(:MyModel) }
+
+  migration_context :up do
+    # Create data
+    #
+    # Example:
+    #
+    #   my_model_stub.create!(attributes)
+
+    migrate
+
+    # Ensure data exists
+    #
+    # Example:
+    #
+    #   expect(record).to have_attributes()
+  end
+
+  migration_context :down do
+    # Create data
+    #
+    # Example:
+    #
+    #   my_model_stub.create!(attributes)
+
+    migrate
+
+    # Ensure data exists
+    #
+    # Example:
+    #
+    #   expect(MyModel.count).to eq(0)
+  end
+end
--- a/spec/lib/generators/migration_generator_spec.rb
+++ b/spec/lib/generators/migration_generator_spec.rb
@@ -1,0 +1,55 @@
+require 'spec_helper'
+require 'rails/generators/test_case'
+require 'generators/migration/migration_generator'
+
+describe ManageIQ::Schema::MigrationGenerator do
+  include Rails::Generators::Testing::Behaviour
+  include Rails::Generators::Testing::SetupAndTeardown
+  include Rails::Generators::Testing::Assertions
+  include FileUtils
+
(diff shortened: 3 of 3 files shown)


lib/generators/migration/migration_generator.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails/generators/active_record/migration/migration_generator'
+
+# Overwirtes the default `ActiveRecord::Generators::MigrationGenerator` to
```
Fryguy, line 3:
```suggestion
# Overwrites the default `ActiveRecord::Generators::MigrationGenerator` to
```

:trollface:

lib/generators/migration/templates/migration_spec.rb.tt
```diff
@@ -0,0 +1,29 @@
+require_migration
+
+describe MoveMiqSetMemberships do
+  # let(:my_model_stub) { migration_stub(:MyModel) }
+
+  migration_context :up do
+    # Create data
```
Fryguy, line 7:
```suggestion
    # Create data
    #
    # Example:
    #
    #   my_model_stub.create!(attributes)
```

Fryguy:
This is pretty cool.  We may also want a note inside the generator file that the spec is only necessary for data migrations and should not be included for "normal" Rails migrations.

lib/generators/migration/templates/migration_spec.rb.tt
NickLaMuro: 👍

I think I stopped filling this out as it took me forever to figure out how to actually override it properly.

spec/lib/generators/migration_specs_generator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  before do 
+    tests MigrationSpecsGenerator
+    destination Rails.root.join('tmp/generators')
+    prepare_destination
+  end
+
+  it "runs the generator without errors" do
+    expect {
+      run_generator
+    }.not_to raise_error
```
NickLaMuro, line 16: Not entirely sure these specs even would run (if I didn't have the old model names here).

Will see if they do, but if not, I will probably delete this file.

NickLaMuro: Okay, got this working now.  Resolving.

kbrock:
Thank you.
You have made life for the developers just a little bit better. one migration at a time.

lib/generators/migration/templates/migration_spec.rb.tt
```diff
@@ -0,0 +1,37 @@
+require_migration
```
kbrock, line 1: I'm still very proud of this line -- so much magic to get us here

lib/generators/migration/migration_generator.rb
```diff
@@ -0,0 +1,56 @@
...
+    # overwrite and work properly
+    #
+    def self.namespace(name = nil)
+      return super if name
+
+      "migration"
+    end
+
+    def self.miq_schema_root
+      File.expand_path(File.join("..", "..", ".."), __dir__)
```
Fryguy, line 27: Would this be simpler?
```suggestion
      ManageIQ::Schema::Engine.root
```

Then you also have a Pathname object, so the other calls like db_migrate_path can piggyback on that

```ruby
def db_migrate_path
  miq_schema_root.join("db", "migrate")
end
```

lib/generators/migration/templates/migration_spec.rb.tt
```diff
@@ -0,0 +1,37 @@
...
+
+describe <%= spec_class_name %> do
+  # let(:<%= spec_symbol_name %>_stub) { migration_stub(:<%= spec_class_name %>) }
```
Fryguy, line 4: I don't think this is right - it looks like the stub has to do with the migration name, but it is kind of unrelated - it's more about the models defined in the migration.  I think the old way with the hardcode MyModelStub looked clearer.

spec/lib/generators/migration_generator_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  include Rails::Generators::Testing::SetupAndTeardown
+  include Rails::Generators::Testing::Assertions
+  include FileUtils
+
+  tests       ManageIQ::Schema::MigrationGenerator
+  destination Rails.root.join('tmp/generators')
+  
+  before do 
+    prepare_destination
+    allow(described_class).to receive(:miq_schema_root).and_return(destination_root)
```
Fryguy, line 16: _Technically_ this is returning a Pathname where the original method didn't, which makes them unmatched.  However, if you change it as I suggested in https://github.com/ManageIQ/manageiq-schema/pull/592/files#r678321714, then they'd both be Pathnames.

lib/generators/migration/migration_generator.rb
NickLaMuro: Done.

spec/lib/generators/migration_generator_spec.rb
NickLaMuro: Done.

lib/generators/migration/migration_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+    #
+    def self.namespace(name = nil)
+      return super if name
+
+      "migration"
+    end
+
+    def self.miq_schema_root
+      ManageIQ::Schema::Engine.root
+      # File.expand_path(File.join("..", "..", ".."), __dir__)
```
Fryguy, line 28: I think these commented bits can go away now.

Fryguy (approved):

Fryguy:
Updated the OP to link https://github.com/ManageIQ/manageiq-schema/issues/479

lib/generators/migration/migration_generator.rb
NickLaMuro: F... I kept these as comments there in case using `Pathname` was going to be a problem, but forgot to remove them once the tests passed.

Thanks for the ping.  Resolving.

Fryguy:
@NickLaMuro Can you explain the `-p` argument in the OP?  I feel like _not_ having it as the default will cause more trouble, because it creates an invalid spec test.

NickLaMuro:
@Fryguy it is short for `--pretend`, which means it is a `--no-op`.

Fryguy:
I think you must have a typo in the OP then?

```
$ bin/rails g migration test_two
      create  db/migrate/20210727213008_test_two.rb
      create  spec/migrations/20210727213008_test_two.rb
```

That second one should end in _spec.rb, right?

NickLaMuro:
@Fryguy Oh right, you are correct. Fixed

(and yes, I just took the first one and replaced output without actually running it... the secret is out...)

jrafanie:
This is sweet,  nice work @NickLaMuro

PR #595, opened 2021-07-29
Add `#options` to Endpoints

agrare:
Some data for an endpoint doesn't always fit in just the URL properties, an extra options jsonb column can be used to store endpoint-related info that isn't strictly about the URL.


agrare:
ref: https://github.com/ManageIQ/manageiq-providers-ibm_power_vc/pull/20#discussion_r673103117

PR #596, opened 2021-08-01
SDE-1726_add_wwpn_candidates

borisko123:
When creating host initiator, for FC ports, we need to fill in wwpn
So-called wwpn candidates will be populated to help make a choice

### DIFF
--- a/db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
+++ b/db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
@@ -1,0 +1,14 @@
+class CreateStorageHostsWwpnCandidates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :storage_hosts_wwpn_candidates do |t|
+      # stores WWPN candidate
+      t.string :candidate
+
+      t.string :ems_ref
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :physical_storage, :type => :bigint, :index => true
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20210925144400_create_wwpn_candidates.rb
+++ b/db/migrate/20210925144400_create_wwpn_candidates.rb
@@ -1,6 +1,6 @@
-class CreateStorageHostsWwpnCandidates < ActiveRecord::Migration[6.0]
+class CreateWwpnCandidates < ActiveRecord::Migration[6.0]
   def change
-    create_table :storage_hosts_wwpn_candidates do |t|
+    create_table :wwpn_candidates do |t|
       # stores WWPN candidate
       t.string :candidate
 
--- a/spec/support/table_list.txt
+++ b/spec/support/table_list.txt
@@ -309,6 +309,7 @@ showback_rates
 showback_tiers
 snapshots
 storage_files
+storage_hosts_wwpn_candidates
 storage_profile_storages
 storage_profiles
 storage_resources
--- a/spec/support/table_list.txt
+++ b/spec/support/table_list.txt
@@ -309,7 +309,6 @@ showback_rates
 showback_tiers
 snapshots
 storage_files
-storage_hosts_wwpn_candidates
 storage_profile_storages
 storage_profiles
 storage_resources
@@ -335,4 +334,5 @@ vms
 volume_mappings
 volumes
 windows_images
+wwpn_candidates
 zones


db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
```diff
@@ -0,0 +1,14 @@
+class CreateStorageHostsWwpnCandidates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :storage_hosts_wwpn_candidates do |t|
+      # stores WWPN candidate
+      t.string :candidates
```
Fryguy, line 5: Is this a string holding an array?  Having a table, I'd expect one candidate per row.

Can you describe the relationships a little more in the OP?

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
```diff
@@ -0,0 +1,14 @@
+class CreateStorageHostsWwpnCandidates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :storage_hosts_wwpn_candidates do |t|
```
agrare, line 3: Since this belongs to a physical_storage might `physical_storage_wwpn_candidates` be a better name?

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
bdunne: Will this contain all of the addresses that are not already in use?

db/migrate/20210726060426_add_fields_to_host_initiators.rb
```diff
@@ -0,0 +1,6 @@
+class AddFieldsToHostInitiators < ActiveRecord::Migration[6.0]
+  def change
+    add_column :host_initiators, :status, :string
```
Fryguy, line 3: What kind of values will be stored here?  I'm curious if we will need normalization or just want to store the raw values (or both?)

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
borisko123: 1.It will be one candidate per row. The name should be : 'candidate' . I will fix this, (this is mentioned by comment..)
2. This table will contain all of the addresses that are not already in use

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
borisko123: will change to this name

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
bdunne: Should this column be singular `candidate` then?

bdunne: Please correct me if I'm wrong (I don't have much experience with FiberChannel) but my understanding is that a WWPN would be something like `10:00:00:00:c9:b0:51:32`.
Are you planning to store trillions of rows per EMS (minus the few we discover from refresh) in this table?
Are we going to seed the table?

borisko123: we are going to store only WWPN 'candidates'. These are WWPN which are not assigned yet
usually it is maximum 512 per node. But this is theoretically, since these WWPN are usually published and assigned one by one, and after that they 'disappear', so the actual amount of these 'candidates'  will be dozens per storage at the best.

db/migrate/20210726060426_add_fields_to_host_initiators.rb
borisko123: included by mistake.  not related to this PR

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
borisko123: > Should this column be singular `candidate` then?

done

db/migrate/20210801104200_create_storage_hosts_wwpn_candidates.rb
agrare: @ereztheodorou the `storage_hosts` still is confusing though because we don't have that name anywhere in our database and the current table name is in the format of a rails join table (which would normally indicate that it joined storage_hosts model).

borisko123: @ereztheodorou @agrare so maybe simply the 'wwpn_candidates' will go?

agrare: Yeah that works if you don't want the physical_storage prefix

agrare (approved):

agrare:
@borisko123 changes look good can you rebase your commits into a sane history?  At a minimum drop the merge commits and with a simple change like this a single commit is probably sufficient

PR #598, opened 2021-08-08
added columns and props for health state of cloud volume

IlyaKessel:
added columns and props for the health state of cloud volume
image
image


IlyaKessel:
@agrare, we are using a status for indication of statuses like "pending creation", "pending deletion", modification and it can be changed for example by actions of storage admin.
health state is more for states like online or offline is it accessible or not it can be changed by the storage system because of some it the device itself.

agrare (approved):

PR #345, opened 2019-03-07
Remove container deployments and container deployment nodes

d-m-u:
Per https://bugzilla.redhat.com/show_bug.cgi?id=1672937 we're removing container deployment.

There are two commits here because we're also removing the Auth classes that were a part of the original container deployment PRs so this includes a data migration to remove those. 

Most of the relevant files in core were introduced in https://github.com/ManageIQ/manageiq/pull/7021. 

~~The UI PR that originally added container deployments was https://github.com/ManageIQ/manageiq/pull/7620.~~
Most of that code was removed already in 
~~https://github.com/ManageIQ/manageiq/pull/18095~~ and 
~~https://github.com/ManageIQ/manageiq-ui-classic/pull/4770~~

Bunch of related deletion PRs:

## Related: 
https://github.com/ManageIQ/manageiq/pull/18515
~~https://github.com/ManageIQ/manageiq-automation_engine/pull/301~~
~~https://github.com/ManageIQ/manageiq-content/pull/511~~
~~https://github.com/ManageIQ/manageiq-api/pull/560~~

**TLDR: take the code out, put it back in, take it out, soap on, soap off, no I don't even know where that reference is from...**

PR #599, opened 2021-08-12
Remove various authentication methods

kbrock:
We have some extra fields in our authentications table, first mentioned
in https://github.com/ManageIQ/manageiq-schema/pull/588/files#r676842653

The fields were added in https://github.com/ManageIQ/manageiq/pull/7779
But were removed in https://github.com/ManageIQ/manageiq/pull/18515

The schema migration removed all changes from the first pr except for
the columns in the authentications table https://github.com/ManageIQ/manageiq-schema/pull/345

This PR removes the extra columns from the authentications table

I added where the columns were mentioned in our main code in case we wanted some discussion on the matter.


Fryguy:
public_key I thought was used by other things, particularly when we query provider auth keys, but perhaps I'm wrong since it was added by this deployment code?  Maybe @agrare knows.

agrare:
`AuthKeyPair#public_key` is set by the GCE provider

kbrock:
**update:** no longer removing `public_key`.
grep'd all code including javascript and yaml again for phrases without using `\b` to get more hits.

only change was to a field that was undetermined, no changes to suggestions.

db/migrate/20210811194501_remove_deployment_from_authentications.rb
```diff
@@ -0,0 +1,69 @@
+class RemoveDeploymentFromAuthentications < ActiveRecord::Migration[6.0]
+  def up
+    change_table "authentications" do |t|
+      t.remove :challenge
+      t.remove :login
+      ## keeping: t.remove :public_key
```
agrare, line 6: Probably should remove this comment?

db/migrate/20210811194501_remove_deployment_from_authentications.rb
```diff
@@ -0,0 +1,69 @@
...
+      t.remove :rhsm_pool_id
+      t.remove :rhsm_server
+    end
+  end
+
+  def down
+    change_table "authentications" do |t|
+      t.boolean :challenge
+      t.boolean :login
+      ## keeping: t.text :public_key
```
agrare, line 40: Same

agrare:
Aside from the commented out change in the migration this LGTM
@Fryguy ?

db/migrate/20210811194501_remove_deployment_from_authentications.rb
kbrock: I liked having it in here but am more than happy to remove

kbrock:
removed column public_key code comment and put into git comment instead.

PR #600, opened 2021-08-16
More accurately check schema dumper lines

Fryguy:
Prior to this commit, the code would check that the line in the
schema.rb by using a substring match, which could be inaccurate,
particularly when a table name may be a substring of another table name.
This commit changes the lookups to only look at the full table names.

@agrare Please review.


Fryguy:
I intentionally did not change the last 2 tests, because they already do a "full string" check; they just happen to check the full schema line.

agrare (approved):

PR #602, opened 2021-08-20
Convert NetworkRouter#admin_state_up from varchar to boolean

kbrock:
We have this column as a boolean in NetworkPort#admin_state_up
but it is a string/text in NetworkRouter

This is a boolean column so converting it in the database
This helps the api handle proper typing for this column

fixes https://github.com/ManageIQ/manageiq-providers-openstack/issues/680

### DIFF
--- a/db/migrate/20210819145618_update_network_router_admin_state_up.rb
+++ b/db/migrate/20210819145618_update_network_router_admin_state_up.rb
@@ -1,0 +1,20 @@
+class UpdateNetworkRouterAdminStateUp < ActiveRecord::Migration[6.0]
+  class NetworkRouter < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    change_column :network_routers, :admin_state_up,
+      "BOOLEAN USING COALESCE(admin_state_up::BOOLEAN, FALSE)"
+
+    change_column_default :network_routers, :admin_state_up, false
+    change_column_null :network_routers, :admin_state_up, false
+  end
+
+  def down
+    change_column_null :network_routers, :admin_state_up, true
+    change_column_default :network_routers, :admin_state_up, nil
+
+    change_column :network_routers, :admin_state_up, "varchar"
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20210819145618_update_network_router_admin_state_up.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateNetworkRouterAdminStateUp < ActiveRecord::Migration[6.0]
+  class NetworkRouter < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    NetworkRouter.where(:admin_state_up => nil).update_all(:admin_state_up => false)
```
Fryguy, line 7: Please wrap in say_with_time.

Fryguy:
@agrare Please review.

agrare (approved):

kbrock:
comment: when there are this many tests, that means there was a ton of confusing stuff going on. I can delete a bunch of these test now that everything seems in order

db/migrate/20210819145618_update_network_router_admin_state_up.rb
```diff
@@ -0,0 +1,20 @@
+class UpdateNetworkRouterAdminStateUp < ActiveRecord::Migration[6.0]
+  class NetworkRouter < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    change_column :network_routers, :admin_state_up,
+      "boolean USING COALESCE(admin_state_up::BOOLEAN, FALSE)"
```
kbrock, line 8: this generates
```
ALTER COLUMN "admin_state_up" TYPE boolean USING COALESCE(admin_state_up::BOOLEAN)
```

since the column is currently nullable (and we want to remove nulls) we are removing nulls along the way using `COALESCE`

going from string to boolean requires an explicit cast anyway (`USING admin_state::BOOLEAN`), so it doesn't add too much complexity

kbrock:
updates:
- moved null removal and did as part of the `ALTER TABLE`
- uppercased `boolean` sql type

db/migrate/20210819145618_update_network_router_admin_state_up.rb
kbrock: thnx. like the new solution better.

db/migrate/20210819145618_update_network_router_admin_state_up.rb
kbrock: updated `boolean` to `BOOLEAN`

agrare (approved):

PR #603, opened 2021-08-23
Remove all instances of hardcoded ids and a review spec

### LINKED ISSUE #434: Avoid using hardcoded number literals for ids in specs
In general we should avoid using hardcoded number literal for ids in specs, because it ignores the reality of regions.  That is, if the database is created in a random region, then we are creating data from technically a different regin.  I'm not against hardcoding ids, as that is generally useful so we don't have to create extra unnecessary records, but when doing so we should have some helper method to get an id "in this region"

`git grep "_\?id\s*=>\?\s*[0-9]" spec` is a list of where we've done this.

Fryguy:
Closes #434

@jrafanie Please review.


Fryguy:
This is an example output from master:

```
Failures:

  1) No hardcoded ids need tests
     Failure/Error: expect(hardcoded_ids).to be_empty, hardcoded_ids_message(hardcoded_ids)

       Hardcoded id numbers were found in the following locations:

           spec/migrations/20181130203334_classification_parent_null_spec.rb:8:      c = classification_stub.create(:parent_id => 0)
           spec/migrations/20181130203334_classification_parent_null_spec.rb:15:      c = classification_stub.create(:parent_id => 1)
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:19:      :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:41:        :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:49:        :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:83:        :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:97:        :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:113:        :zone_id         => 99,
           spec/migrations/20201019211215_create_legacy_managers_spec.rb:121:        :zone_id         => 99,

       Hardcoded id numbers should not be present in a data migration spec, as
       they ignore region numbers, creating an invalid database setup. Instead,
       use the following helper to generate ids in the correct region:

           # Example, in region 1, requesting id number 2
           anonymous_class_with_id_regions.id_in_region(2, anonymous_class_with_id_regions.my_region_number)
           # => 1000000000002

       If you believe this file is a false postive, add the following as a
       tailing comment to the line in question

           # manageiq:disable HardcodedIds
     # ./spec/automated_review/no_hardcoded_ids_spec.rb:29:in `block (2 levels) in <top (required)>'
```

spec/automated_review/no_hardcoded_ids_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      tailing comment to the line in question
+
+          # manageiq:disable HardcodedIds
+
+    EOS
+  end
+
+  it "need tests" do
+    hardcoded_ids = Dir.chdir(ManageIQ::Schema::Engine.root) do
+      `grep -n "_\\?id\\s*=>\\?\\s*[0-9]" spec/**/*_spec.rb | grep -v "manageiq:disable HardcodedIds"`.strip.split("\n")
```
jrafanie, line 26: Is that regex ignoring ruby 1.9 format: `Classification.where(parent_id: 0)`?  🤣

jrafanie:
LGTM, other than the question above.  I'm fine if we assume hashrockets but thought it was fun to point out for all the :trollface:

spec/automated_review/no_hardcoded_ids_spec.rb
Fryguy: yeah technically :)

Fryguy: I tried adding in the 1.9 format, but I hit things like https://github.com/Fryguy/manageiq-schema/blob/f51b29941dc719c4e1870340151cd3f35fa7bc74/spec/migrations/20200212171537_remove_ovirt_sdk4_from_event_stream_spec.rb#L14, so I'm gonna punt for now

jrafanie: yeah, handling one format will probably have false positives on it's own.  But yes, since 1.8 hash syntax is still most common, we can punt on 1.9 and even drop this for 1.8 if false positives happen that frequently.

jrafanie:
> [ ] ❗ - [Line 21](https://github.com/Fryguy/manageiq-schema/blob/5de4ac64398c737f24bd2fb441f029f2d43669da/spec/automated_review/no_hardcoded_ids_spec.rb#L21), Col 1 - [Naming/HeredocDelimiterNaming](https://docs.rubocop.org/rubocop/1.13/cops_naming.html#namingheredocdelimiternaming) - Use meaningful heredoc delimiters.

Bot is trolling you with opinionated rubocops.  LOL.

spec/automated_review/no_hardcoded_ids_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      If you believe this file is a false postive, add the following as a
+      tailing comment to the line in question
+
+          # manageiq:disable HardcodedIds
+
+    EOS
+  end
+
+  it "need tests" do
```
chessbyte, line 24: this should be renamed to something more meaningful

spec/automated_review/no_hardcoded_ids_spec.rb
chessbyte: but wouldn't the ovirt example skip the grep because the ID has characters outside of 0-9?

jrafanie: > but wouldn't the ovirt example skip the grep because the ID has characters outside of 0-9?

yeah, we'd have to ensure all values following the hashrocket are bigints.  I'm not sure if that will still leave us with false positives or not.

Fryguy: > but wouldn't the ovirt example skip the grep because the ID has characters outside of 0-9?

If I spent enough time on it, yes :) I got up to `"_\?id\s*(=>\?|:)\s*[0-9]+\b"` but it still matched when it "starts with 0-9, but followed by a dash" - would need to specify "not immediately followed by a digit", but then you start getting into different grep implementations supporting negative lookahead or not, and I just couldn't be bothered anymore haha.

spec/automated_review/no_hardcoded_ids_spec.rb
Fryguy: Oh thanks - I copy-pasted and missed that

PR #610, opened 2021-09-29
Introducing Host Initiator Group

gregoryb-at-ibm:
Block storage devices have a configuration item commonly called "host cluster". Its role is to aggregate hosts on the storage into groups (in MIQ, storage hosts are called host-initiators). These host clusters allow an administrator to map a volume to several hosts in one operation.

This PR is part of this issue: https://github.com/ManageIQ/manageiq-providers-autosde/issues/93

### DIFF
--- a/db/migrate/20210810114405_create_host_initiator_groups.rb
+++ b/db/migrate/20210810114405_create_host_initiator_groups.rb
@@ -1,0 +1,15 @@
+class CreateHostInitiatorGroups < ActiveRecord::Migration[6.0]
+  def change
+    create_table :host_initiator_groups do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :physical_storage, :type => :bigint, :index => true
+
+      t.string :name
+      t.string :status
+      t.string :ems_ref
+      t.string :uid_ems
+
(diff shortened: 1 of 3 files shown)


db/migrate/20210810114405_create_host_initiator_groups.rb
```diff
@@ -0,0 +1,16 @@
+class CreateHostInitiatorGroups < ActiveRecord::Migration[6.0]
+  def change
+    create_table :host_initiator_groups do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :physical_storage, :type => :bigint, :index => true
+      t.references :host_initiator, :type => :bigint, :index => true
```
Fryguy, line 6: I was assuming a group would has_many initiators, but unless I'm reading it wrong, this makes it a 1:1

agrare: yeah I think you want to add `t.references :host_initiator_group` to `:host_initiators` assuming you want a group to have >1 initiator

gregoryb-at-ibm: Sorry, this has been an oversight.

PR #613, opened 2021-10-22
Remove cockpit from assigned services

kbrock:
part of ManageIQ/manageiq#21378

cockpit is no longer an available service
remove it from the assigned services table

### DIFF
--- a/db/migrate/20211022131607_drop_cockpit.rb
+++ b/db/migrate/20211022131607_drop_cockpit.rb
@@ -1,0 +1,31 @@
+class DropCockpit < ActiveRecord::Migration[6.0]
+  class AssignedServerRole < ActiveRecord::Base; end
+
+  class ServerRole < ActiveRecord::Base; end
+
+  class SettingsChange < ActiveRecord::Base; end
+
+  class MiqUserRole < ActiveRecord::Base; end
+
+  class MiqRolesFeature < ApplicationRecord; end
+
+  class MiqProductFeature < ApplicationRecord
+    # habtm :miq_user_roles :join_table => MiqRolesFeature
+  end
+
+  def up
+    say_with_time("Deleting cockpit configuration") do
+      if (cockpit = ServerRole.find_by(:name => "cockpit_ws"))
+        AssignedServerRole.where(:server_role_id => cockpit.id).delete_all
+        cockpit.delete
+      end
+
+      SettingsChange.where("key LIKE '%/cockpit_ws_worker/%'").delete_all
+
+      if (feature = MiqProductFeature.find_by(:identifier => "cockpit_console"))
+        MiqRolesFeature.where(:miq_product_feature_id => feature.id).delete_all
+        feature.delete
+      end
+    end
+  end
+end
--- a/spec/migrations/20211022131607_drop_cockpit_spec.rb
+++ b/spec/migrations/20211022131607_drop_cockpit_spec.rb
@@ -1,0 +1,47 @@
+require_migration
+
+describe DropCockpit do
+  let(:server_role_stub)          { migration_stub(:ServerRole) }
+  let(:assigned_server_role_stub) { migration_stub(:AssignedServerRole) }
+  let(:miq_product_feature_stub)  { migration_stub(:MiqProductFeature) }
+  let(:miq_user_role_stub)        { migration_stub(:MiqUserRole) }
+  let(:miq_role_feature_stub)     { migration_stub(:MiqRolesFeature) }
+
+  migration_context :up do
+    it "deletes only cockpit active services" do
+      cockpit = server_role_stub.find_or_create_by!(:name => "cockpit_ws")
+      other   = server_role_stub.create!(:name => "other_service")
+
+      assigned_server_role_stub.create!(:server_role_id => other.id)
+      3.times do
+        assigned_server_role_stub.create!(:server_role_id => cockpit.id)
+      end
+
+      expect(assigned_server_role_stub.where(:server_role_id => cockpit.id).count).to eq(3)
+      expect(assigned_server_role_stub.where(:server_role_id => other.id).count).to eq(1)
+
+      migrate
+
+      expect(assigned_server_role_stub.where(:server_role_id => other.id).count).to eq(1)
+      expect(assigned_server_role_stub.where(:server_role_id => cockpit.id).count).to eq(0)
+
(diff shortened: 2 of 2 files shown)


kbrock:
WIP: looks like we need to remove this from the assigned roles join table. touching this up and will un WIP

Fryguy:
@kbrock Is the removal done with https://github.com/ManageIQ/manageiq-schema/pull/613/files#diff-3722f65c2deae8fcdf7586e0385f0c705ea7888a1de6a76a1dfc65a15567462eR16 ?  If so, what else is left?

kbrock:
> @kbrock Is the removal done with https://github.com/ManageIQ/manageiq-schema/pull/613/files#diff-3722f65c2deae8fcdf7586e0385f0c705ea7888a1de6a76a1dfc65a15567462eR16 ? If so, what else is left (aside from minor stuff like the constant name and the bonus unneeded comment).

https://github.com/ManageIQ/manageiq/pull/21513#discussion_r738677663
https://github.com/ManageIQ/manageiq/pull/21513#discussion_r738680223

The only outstanding records I can think of are maybe authentications?

db/migrate/20211022131607_drop_cockpit.rb
```diff
@@ -0,0 +1,21 @@
...
+  class AssignedServerRole < ActiveRecord::Base
+  end
+
+  class ServerRole < ActiveRecord::Base
+  end
+
+  def up
+    cockpit = ServerRole.where(:name => COCKPIT).first
+
+    return if cockpit.nil?
```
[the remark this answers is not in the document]
kbrock: think I'll merge these into the action portion.

we are now doing a few things and we can put a guard clause for each.

db/migrate/20211022131607_drop_cockpit.rb
```diff
@@ -0,0 +1,21 @@
+class DropCockpit < ActiveRecord::Migration[6.0]
+  COCKPIT = "cockpit_ws".freeze
```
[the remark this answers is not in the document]
kbrock: This is only used once. Putting it into context seems like a better idea

kbrock:
@jrafanie you are the one who found most of the records that slipped through the cracks. Let me know if you get some more inspiration that would suggest another scragglier is left all alone.

db/migrate/20211022131607_drop_cockpit.rb
```diff
@@ -0,0 +1,35 @@
...
+  class MiqUserRole < ActiveRecord::Base
+    has_and_belongs_to_many :miq_product_features,
+                            :join_table => "miq_roles_features",
+                            :class_name => "DropCockpit::MiqProductFeature"
+  end
+
+  class MiqProductFeature < ApplicationRecord
+    has_and_belongs_to_many :miq_user_roles,
+                            :join_table => "miq_roles_features",
+                            :class_name => "DropCockpit::MiqUserRole"
```
Fryguy, line 17: habtm can cause problems in migrations because Rails creates these temporary models in the middle that don't play nice with cache clearing.  If we can get rid of these, that would be preferable, even though it might make the interface here clunky.

db/migrate/20211022131607_drop_cockpit.rb
```diff
@@ -0,0 +1,35 @@
...
+  class AssignedServerRole < ActiveRecord::Base; end
+
+  class ServerRole < ActiveRecord::Base; end
+
+  class SettingsChange < ActiveRecord::Base; end
+
+  class MiqUserRole < ActiveRecord::Base
+    has_and_belongs_to_many :miq_product_features,
+                            :join_table => "miq_roles_features",
+                            :class_name => "DropCockpit::MiqProductFeature"
```
Fryguy, line 11: Is this being used?  If not, can we drop it?

Fryguy:
Nice find on the product feature - I completely missed that on the first pass.

db/migrate/20211022131607_drop_cockpit.rb
kbrock: it was used in the tests. can approach from the other end. not as easy but it works

db/migrate/20211022131607_drop_cockpit.rb
kbrock: You are suggesting creating a `MiqRoleFeature` model with no id?

kbrock (commented):
@Fryguy just a few questions

(not sure why this is a review but...)

db/migrate/20211022131607_drop_cockpit.rb
Fryguy: > You are suggesting creating a `MiqRoleFeature` model with no id?

Yes, create an MiqRoleFeature model and do the through join more manually. (note, rhe table should already have an id as it's a requirement)

db/migrate/20211022131607_drop_cockpit.rb
Fryguy: If it's only for the tests, then it should only live in the tests, however, that may be moot as discussed in https://github.com/ManageIQ/manageiq-schema/pull/613#discussion_r744948932

kbrock:
ok, :habtm is all gone

db/migrate/20211022131607_drop_cockpit.rb
```diff
@@ -0,0 +1,31 @@
...
+  class ServerRole < ActiveRecord::Base; end
+
+  class SettingsChange < ActiveRecord::Base; end
+
+  class MiqUserRole < ActiveRecord::Base; end
+
+  class MiqRolesFeature < ApplicationRecord; end
+
+  class MiqProductFeature < ApplicationRecord
+    # habtm :miq_user_roles :join_table => MiqRoleFeature
```
Fryguy, line 13: Love this comment :+1:

```suggestion
    # habtm :miq_user_roles :join_table => MiqRolesFeature
```

Fryguy (approved):
LGTM - @jrafanie Please also review.

db/migrate/20211022131607_drop_cockpit.rb
kbrock: one I realized that the table still has an id, it was not so bad

db/migrate/20211022131607_drop_cockpit.rb
kbrock: gone

Fryguy (approved):

jrafanie (approved):

PR #616, opened 2021-11-04
Update dummy app to Rails 6.0.1.4

Fryguy:
This PR updates the dummy app to Rails 6.0.1.4 to match what we use in ManageIQ core.

```
rm -rf dummy/.git
```

---

I intentionally built the PR commits in a way that in the future, if we have to upgrade again, we can follow the commit's pattern.  The commits:

1. Delete the old app
2. Create the new app
3. Apply changes to the app in a single commit.


Gemfile
```diff
@@ -12,3 +10,6 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
+
+# Load Gemfile with dependencies from dummy app
+eval_gemfile(File.expand_path("spec/dummy/Gemfile", __dir__))
```
Fryguy, line 15: This is another point of discussion that I'm not sure about, and I'm honestly not sure how it even worked before.  Previously we had the following changes that I did not bring over:

https://github.com/ManageIQ/manageiq-schema/pull/616/files#diff-34e1b7b0905d8d8ffb401395c2e3bb17ee91c912b307a7586583c3ad132684b6L2-R1 and https://github.com/ManageIQ/manageiq-schema/pull/616/files#diff-ad366e89306810a068ae008a627d8fe6c27f88bf77bee35cd828f784bc6de062L14

Notice the paths - Previously they were `../../..` and now they are `..`.  This implies that the dummy app reached back into the plugin and used its Gemfile, but I don't understand how that could have possibly worked.  In fact, when I first upgraded to Rails 6, none of the rake tasks were available to me, such as `db:drop`, and I don't understand how it was possible before.

---

I think this change here, that uses the dummy's Gemfile in the way we do in ManageIQ is a more accurate reflection of what we do with other plugins.  There might be a more "correct" way of testing plugins with dummy apps, but I haven't found it yet.

Fryguy: Decided to do it the old way for now, which seems to work, so as not to change too many things at once.

Fryguy:
Decided to use a temporary override of belongs_to_required_by_default, and fix the individual migrations in follow up PRs.

lib/tasks_private/spec.rake
```diff
@@ -36,7 +36,8 @@ namespace :spec do
 
   desc "Run all non-migration specs"
   RSpec::Core::RakeTask.new(:non_migration => :initialize) do |t|
-    t.pattern = FileList["spec/**/*_spec.rb"].exclude("spec/migrations/**/*_spec.rb")
+    t.pattern = FileList["spec/**/*_spec.rb"].exclude("spec/dummy/**/*_spec.rb")
```
jrafanie, line 39: 🙇

spec/dummy/config/environments/development.rb
```diff
@@ -45,7 +53,7 @@
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
-  # Raises error for missing translations
+  # Raises error for missing translations.
```
jrafanie, line 56: I'm glad someone in upstream rails fixed this huge typo.  😆

jrafanie (approved):
LGTM

PR #618, opened 2021-11-19
Delete hawkular endpoints and authentications

agrare:
Remove any hawkular endpoint and authentication records as they are no longer supported

https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/453

### DIFF
--- a/db/migrate/20211119152213_remove_hawkular_authentications.rb
+++ b/db/migrate/20211119152213_remove_hawkular_authentications.rb
@@ -1,0 +1,17 @@
+class RemoveHawkularAuthentications < ActiveRecord::Migration[6.0]
+  class Authentication < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class Endpoint < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def up
+    say_with_time("Delete all hawkular authentications and endpoints") do
+      Authentication.in_my_region.where(:authtype => "hawkular").delete_all
+      Endpoint.in_my_region.where(:role => "hawkular").delete_all
+    end
+  end
+end
--- a/spec/migrations/20211119152213_remove_hawkular_authentications_spec.rb
+++ b/spec/migrations/20211119152213_remove_hawkular_authentications_spec.rb
@@ -1,0 +1,60 @@
+require_migration
+
+class RemoveHawkularAuthentications < ActiveRecord::Migration[6.0]
+  class Authentication < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :resource, :polymorphic => true
+  end
+
+  class Endpoint < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    belongs_to :resource, :polymorphic => true
+  end
+
+  class ExtManagementSystem < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    has_many :authentications, :as => :resource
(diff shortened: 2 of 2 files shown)


agrare:
The EMS and the `belongs_to resource` aren't strictly needed for the specs but I wanted to setup the env as closely as possible to what we'd see on a live system.

db/migrate/20211119152213_remove_hawkular_authentications.rb
```diff
@@ -0,0 +1,19 @@
+class RemoveHawkularAuthentications < ActiveRecord::Migration[6.0]
+  class Authentication < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :resource, :polymorphic => true
```
Fryguy, line 6: I don't see the resource being used in the migration, so do we need it?  If it's only for specs, I recommend just using resource_type, resource_id directly (which I think you already did)

db/migrate/20211119152213_remove_hawkular_authentications.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  class Endpoint < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    belongs_to :resource, :polymorphic => true
+  end
+
+  def up
+    Authentication.in_my_region.where(:authtype => "hawkular").delete_all
+    Endpoint.in_my_region.where(:role => "hawkular").delete_all
```
Fryguy, line 17: Please wrap in say_with_time.

spec/migrations/20211119152213_remove_hawkular_authentications_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_migration
+
+class RemoveHawkularAuthentications < ActiveRecord::Migration[6.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    has_many :authentications, :as => :resource
+    has_many :endpoints, :as => :resource
```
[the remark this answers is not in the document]
Fryguy: Oh no I see it below.

spec/migrations/20211119152213_remove_hawkular_authentications_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+describe RemoveHawkularAuthentications do
+  let(:auth_stub)     { migration_stub(:Authentication) }
+  let(:endpoint_stub) { migration_stub(:Endpoint) }
+  let(:ems_stub)      { migration_stub(:ExtManagementSystem) }
+
+  migration_context :up do
+    let(:kube) { ems_stub.create!(:type => "ManageIQ::Providers::Kubernetes::ContainerManager", :name => "kube") }
+
+    before do
+      auth_stub.create!(:type => "AuthToken", :authtype => "bearer", :userid => "_", :resource_id => kube.id, :resource_type => ems_stub)
```
Fryguy, line 22: Might be better to use kube.class here, to keep it consistent.

```suggestion
      auth_stub.create!(:type => "AuthToken", :authtype => "bearer", :userid => "_", :resource_id => kube.id, :resource_type => kube.class)
```

agrare: Just habit from `kube.class` usually being the namespaced subclass where resource_type is the base class.  It does work in this case though so updated

db/migrate/20211119152213_remove_hawkular_authentications.rb
agrare: It made the specs a little cleaner (I could do kube.endpoints for example), I'll move the `belongs_to :resource, :polymorphic => true` to the spec file

PR #619, opened 2021-11-24
Creation of host initiator group

gregoryb-at-ibm:
In this set of PRs we will allow users to create new host-initiator-groups on their storage systems through MIQ.

Part of https://github.com/ManageIQ/manageiq-providers-autosde/issues/108

image

image


agrare (approved):

agrare:
`HostInitiatorGroup` already has `include NewWithTypeStiMixin` so I think the missing `:type` column was just an oversight, also `HostInitiator` has a `:type` column as well.

PR #632, opened 2022-01-14
Drop unused pglogical references

jrafanie:
As of https://github.com/ManageIQ/manageiq/pull/18686 we moved from pglogical
to the pg-logical_replication using postgresql's native logical replication.

We switched as of ivanchuk.  It should be safe to drop this now.

Funny enough, this actually helps with getting ruby 3.0 support because the AR
extension signature for drop_table needs to change to work with ruby 2.x and
3.x.


Fryguy:
This _also_ helps with getting us to GitHub Actions (in a way), because after this we can archive the pg-pglogical gem, which I can't get GitHub Actions working on anyway due to the need to mess with the database service: https://github.com/ManageIQ/pg-pglogical/blob/29f4667183d80d56611011cc152422f485a0ff47/ci/before_install.sh#L3-L9

PR #635, opened 2022-02-02
Support rails 6.1

jrafanie:
* Use id_regions with 6.1 support
* Allow rails 6.1 but force local migration creation to still use 6.0
* Fix the global stubbing of the PG::Connection class that was causing failures in rails 6.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ cache: bundler
 rvm:
 - 2.6.6
 - 2.7.2
+env:
+  matrix:
+  - TEST_RAILS_VERSION=6.1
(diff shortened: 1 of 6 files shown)


Fryguy:
Still not sure I understand this, as we really only support 1 version of Rails at a time, but I think we can move forward for now.

jrafanie:
> Still not sure I understand this, as we really only support 1 version of Rails at a time, but I think we can move forward for now.

This was to add support for 6.1.  We can drop 6.0 when we move to 6.1 here and in the migration version.  We still have lots of gem releases before we're there.

PR #636, opened 2022-02-03
Switch to GitHub Actions

Fryguy:
@jrafanie Please review.

Note that you have to merge before it runs, so chicken-and-egg problem.  We can always revert.

Part of https://github.com/ManageIQ/manageiq/issues/21620


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby-version:
+        - '2.5'
```
jrafanie, line 11: why 2.5?  It's EOL since March 2021 I think.  It's not in the .travis.yml either.

Fryguy: copy paste fail from the other gem...i'll fix

jrafanie (approved):
YOLO merge.  We won't know if it's good until we do it.  🤞

PR #637, opened 2022-02-03
Ensure we connect to PG over TCP and not a socket in Github Actions

Fryguy:
@jrafanie Please review.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,7 @@ jobs:
         ports:
         - 5432:5432
     env:
+      PGHOST: localhost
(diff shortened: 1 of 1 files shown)


Fryguy:
I just realized github actions here might break cross-repo...we'll have to come up with a plan @agrare

kbrock (approved):
tend to do this for other github actions/ci I have setup

tangent: do we also need rails version in the `if` code climate check

jrafanie (approved):

Fryguy:
> tangent: do we also need rails version in the if code climate check

yeah, probably...good idea.

PR #638, opened 2022-02-04
Service request connected to Automation request

akhilkr128:
When an automation request is created for ansible, a sub-service request is created and it processes the ansible tasks. Currently, there is no connection between the automation request and the newly created request so we can't write a function to connect the existing requests. :-( But it will work for future requests.

### DIFF
--- a/db/migrate/20220202122417_add_parent_relation_to_miq_requests.rb
+++ b/db/migrate/20220202122417_add_parent_relation_to_miq_requests.rb
@@ -1,0 +1,5 @@
+class AddParentRelationToMiqRequests < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :miq_requests, :parent, :type => :bigint
+  end
+end


chessbyte:
@akhilkr128 This is an open-source project, so please do not provide links that community users cannot traverse.  If you do want to link things up, add the link to the public upstream issue/PR from the private downstream issue/PR, not vice-versa as you did here.

spec/migrations/20220202122417_add_parent_id_to_miq_requests_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_migration
+
+describe AddParentIdToMiqRequests do
+  let(:miq_request) { migration_stub(:MiqRequest) }
+
+  migration_context :up do
+    it "adds parent_id to  Miq Requests" do
+      migrate
+      miq_request.create!(:type => 'MiqRequest', :parent_id => 40)
```
agrare, line 9: You shouldn't use hard-coded IDs in schema specs, e.g. this ID is in region 0 and the region is randomized in spec tests.  It is better to actually create the other record and use that ID

spec/migrations/20220202122417_add_parent_id_to_miq_requests_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_migration
+
+describe AddParentIdToMiqRequests do
+  let(:miq_request) { migration_stub(:MiqRequest) }
+
+  migration_context :up do
```
agrare, line 6: @akhilkr128 simple schema changes like adding a column don't require spec tests, that is more for data migrations.  Just testing that after running the migration you can set a value to the newly created column is really just testing that `ActiveRecord::Migration` is working.

spec/migrations/20220202122417_add_parent_id_to_miq_requests_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it "adds parent_id to  Miq Requests" do
```
chessbyte, line 7:
```suggestion
    it "adds parent_id to MiqRequests" do
```

db/migrate/20220202122417_add_parent_id_to_miq_requests.rb
```diff
@@ -0,0 +1,9 @@
+class AddParentIdToMiqRequests < ActiveRecord::Migration[6.0]
+  class MiqRequest < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+  
+  def change
+    add_column :miq_requests, :parent_id, :bigint
```
agrare, line 7: Is this the id of another `miq_request` ?  I would have expected to see a `belongs_to :parent` relation in the `MiqReqeust` model over here https://github.com/ManageIQ/manageiq/pull/21681/files#diff-ecb22950a62fae5915a90e9d8ead61f3a7021f8ce173e9783f726784eb20b04d

Also usually good practice to use `add_reference` rather than `add_column` as that takes care of a lot of boiler-plate stuff for you like the type, index, etc...

db/migrate/20220202122417_add_parent_id_to_miq_requests.rb
```diff
@@ -0,0 +1,9 @@
+class AddParentIdToMiqRequests < ActiveRecord::Migration[6.0]
+  class MiqRequest < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
```
agrare, line 4: Unless you're using the model name in a data migration (like if you were doing `MiqRequest.in_my_region.update_all(:parent_id => nil)` or something like that) you don't need to define this

agrare:
Kicking github actions after #637

spec/migrations/20220202122417_add_parent_id_to_miq_requests_spec.rb
agrare: Hey look at that we have a spec for this https://github.com/ManageIQ/manageiq-schema/runs/5067896452?check_suite_focus=true#step:6:47

db/migrate/20220202122417_add_parent_id_to_miq_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddParentIdToMiqRequests < ActiveRecord::Migration[6.0]
```
Fryguy, line 1: This class name and file name doesn't match the body anymore.

db/migrate/20220202122417_add_parent_id_to_miq_requests.rb
```diff
@@ -0,0 +1,5 @@
...
+  def change
+    add_reference :miq_requests, :miq_requests
```
Fryguy, line 3: I think you want belongs_to (which is add_reference under the covers).  @agrare this is what you were suggesting right?

```suggestion
    add_reference :miq_requests, :parent, :type => :bigint
```

agrare: >  this is what you were suggesting right?

:+1: yes, exactly

akhilkr128: Changed

PR #643, opened 2022-03-28
Add new field to save customized message for validation

akhilkr128:
[issue](https://github.com/ManageIQ/manageiq/issues/21567)

Need a new field for saving the customized message.

### DIFF
--- a/db/migrate/20220324143540_add_validation_message_to_dialog_fields.rb
+++ b/db/migrate/20220324143540_add_validation_message_to_dialog_fields.rb
@@ -1,0 +1,5 @@
+class AddValidationMessageToDialogFields < ActiveRecord::Migration[6.0]
+  def change
+    add_column :dialog_fields, :validator_message, :string
+  end
+end


db/migrate/20220324143540_add_validation_message_to_dialog_fields.rb
```diff
@@ -0,0 +1,5 @@
+class AddValidationMessageToDialogFields < ActiveRecord::Migration[6.0]
+  def change
+    add_column :dialog_fields, :validator_message, :string
```
jrafanie, line 3: How about `failed_validation_message`?  That or something that matches the terminology we have in the UI for the message field for the failed validation.  If there's in intention to use this field for general informational purposes, the current value here is ok but if it's solely for failure messages, included failure or failed in the field would be helpful.

kbrock:
@akhilkr128 does the request to rename the field make sense to you?

db/migrate/20220324143540_add_validation_message_to_dialog_fields.rb
akhilkr128: `failed` was already removed from the UI since the message only helps the user input the data.

kbrock: @jrafanie you all set with this answer?

akhilkr128:
>@akhilkr128 does the request to rename the field make sense to you?

Completely understood what @jrafanie said. Previously in the UI there was a term `failed` I removed it later because the form won't be submitted and technically it won't fail. This message will notify the user instantly to help the user input data. I think the field name change is not needed. The current UI will look like this.


But if you guys think this field name change will increase code readability please let me know, I will make the changes.

db/migrate/20220324143540_add_validation_message_to_dialog_fields.rb
kbrock: another alternative is `validate_failure_message`

jrafanie (approved):

PR #645, opened 2022-04-26
Add physical server profile

gberginc:
This commit adds a new table for storing server profiles that are used
in Cisco Intersight provider.

Related PR from core: https://github.com/ManageIQ/manageiq/pull/21830

### DIFF
--- a/db/migrate/20220401093502_create_physical_server_profiles.rb
+++ b/db/migrate/20220401093502_create_physical_server_profiles.rb
@@ -1,0 +1,14 @@
+class CreatePhysicalServerProfiles < ActiveRecord::Migration[6.0]
+  def change
+    create_table :physical_server_profiles do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
+      t.string :name
+      t.string :type, :index => true
+      t.references :assigned_server, :type => :bigint, :index => true, :references => :physical_server
+      t.references :associated_server, :type => :bigint, :index => true, :references => :physical_server
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


gberginc:
@agrare This PR took the migration from https://github.com/ManageIQ/manageiq/pull/21830 for the server profiles for Cisco Intersight provider.

gberginc:
@miq-bot assign @agrare

db/migrate/20220401093502_create_physical_server_profiles.rb
```diff
@@ -0,0 +1,13 @@
+class CreatePhysicalServerProfiles < ActiveRecord::Migration[6.0]
+  def change
+    create_table :physical_server_profiles do |t|
+      t.references :ems, :type => :bigint, :index => true, :references => :ext_management_system
+      t.string :ems_ref
```
agrare, line 5: @gberginc you most likely want a `t.string :type, :index => true` column so that we can subclass these under the provider

gberginc: @agrare I am not sure I understand what you mean? Can you share a full example with me?

agrare: Sure thing, so if you want to be able to have a subclass e.g. `ManageIQ::Providers::CiscoIntersight::PhyiscalInfraManager::PhysicalServerProfile` then you have to have a `type` column for `STI`.  This is where rails stores the class name so that when you query the record the correct class is instantiated (see https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html)

You can check out https://github.com/ManageIQ/manageiq-schema/tree/master/db/migrate/20180327162308_create_physical_chassis.rb for an example of creating a table with a `:type` column (we're recently ensured an index is present for type so add `:index => true` to that example)

gberginc: Ah, got it. Thanks for the reference and the explanation.

I've now added the `:type` attribute.

agrare (approved):

PR #648, opened 2022-05-13
[WIP] Data migration for service de-explorization

jeffibm:
Data migration to convert `service_accord` to `service`

@miq-bot add-reviewer @kbrock 
@miq-bot assign @kbrock

**Dependant PR's**
Core - https://github.com/ManageIQ/manageiq/pull/21839
UI - https://github.com/ManageIQ/manageiq-ui-classic/pull/8229


spec/migrations/20220513071902_update_service_features_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  let(:feature_stub) { migration_stub :MiqProductFeature }
+  let(:roles_feature_stub) { migration_stub :MiqRolesFeature }
+
+  migration_context :up do
+    describe 'product features renaming' do
+      it 'renames the Services features' do
+        UpdateServiceFeatures::FEATURE_MAPPING.each_key do |identifier|
+          feature = feature_stub.create!(:identifier => identifier)
+          roles_feature_stub.create!(:miq_product_feature_id => feature.id, :miq_user_role_id => user_role_id)
+        end
```
kbrock, line 14: I'm having trouble reading this dynamic code.
Can you just explicitly type in the feature values for the before and after?

kbrock (changes requested):
1) MiqProductFeature is populated from the yaml file.
   So we do not need to be concerned about this value.

2) user is linked to the user_roles via miq_groups and entitlements
   don't need to test the user, miq groups or entitlements.

3) We may want to test with multiple user_role ids?
   Your test is probably fine. We can read the code to make sure it kept the same user_role_id.

4. The main tests will be checking the changes in the miq_roles_features table.
   We need to make sure a user_role that has the `service_request` product feature before the migration has the `service` product feature after the migration.

6. The roles of concern

Here are the data values of concern.
Since you are just using a single user_role_id, I guess this could be 4 specs.

|case name | before | after|
|-----|--------|------|
|none | neither| neither|
|service|`service`|`service`|
|service_accord|`service_accord`|`service`|
|both|`service_accord` and `service`|`service`|

db/migrate/20220520105024_update_service_user_roles.rb
```diff
@@ -0,0 +1,38 @@
...
+  class MiqProductFeature < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Correcting user created role feature sets") do
+      service         = MiqProductFeature.find_by(:identifier => "service")
+      service_accord  = MiqProductFeature.find_by(:identifier => "service_accord")
+
+      if service_accord
+        affected_user_roles(service_accord).each do |user_role|
+          if user_role.miq_product_features.include?(service_accord) # useless?
```
kbrock, line 15: lets see if this all works with this line removed.
I am assuming that the affected_user_roles will filter so this should not be necessary

db/migrate/20220520105024_update_service_user_roles.rb
```diff
@@ -0,0 +1,38 @@
...
+            unless user_role.miq_product_features.include?(service)
```
kbrock, line 16: so it looks like this line is not working.
It is trying to say that if the user already has a service, then we don't want to add another.

spec/migrations/20220520105024_update_service_user_roles_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(user_role.miq_product_features).to match_array([service])
+    end
+
+    it "converts service_accord to service without duplicate" do
+      user_role = user_role_stub.create!(:miq_product_features => [other, service, service_accord], :read_only => false)
+
+      migrate
+      user_role.reload
+
+      expect(user_role.miq_product_features).to match_array([other, service, service])
```
kbrock, line 36: We are testing the line I mentioned above. Specifically, if the user is already part of a service, then lets not add the service again.

```suggestion
      expect(user_role.miq_product_features).to match_array([other, service])
```

kbrock (commented):
Very close Jeffrey

We can go over again Monday AM

spec/migrations/20220520105024_update_service_user_roles_spec.rb
jeffibm: since its expecting 3 items, can we use like this

```
expect(user_role.miq_product_features.uniq).to match_array([other, service])
expect(user_role.miq_product_features).not_to match_array([service_accord])
```

PR #650, opened 2022-06-17
Migrate Authentication#become_method out of :options

agrare:
Having become_method in the options text column is causing issues for patching via the API, and become_username/become_password are already proper columns on the authentications table.

Add a :become_method column and migrate existing data out of the :options column.

### DIFF
--- a/db/migrate/20220617140542_move_become_method_from_options.rb
+++ b/db/migrate/20220617140542_move_become_method_from_options.rb
@@ -8,7 +8,7 @@ class MoveBecomeMethodFromOptions < ActiveRecord::Migration[6.0]
 
   def up
     say_with_time("Move Authentication become_method out of options text field") do
-      Authentication.in_my_region.find_each do |authentication|
+      Authentication.in_my_region.where("options LIKE '%become_method%'").find_each do |authentication|
         next unless authentication.options&.key?(:become_method)
 
         authentication.become_method = authentication.options.delete(:become_method)
--- a/db/migrate/20220617140542_move_become_method_from_options.rb
+++ b/db/migrate/20220617140542_move_become_method_from_options.rb
@@ -1,0 +1,29 @@
+class MoveBecomeMethodFromOptions < ActiveRecord::Migration[6.0]
+  class Authentication < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    serialize :options
+  end
+
+  def up
+    say_with_time("Move Authentication become_method out of options text field") do
+      Authentication.in_my_region.find_each do |authentication|
+        next unless authentication.options&.key?(:become_method)
+
+        authentication.become_method = authentication.options.delete(:become_method)
+        authentication.save!
+      end
+    end
+  end
+
+  def down
+    say_with_time("Move Authentication become_method back to options text field") do
+      Authentication.in_my_region.where.not(:become_method => nil).find_each do |authentication|
+        authentication.options = {} if authentication.options.nil?
+        authentication.options[:become_method] = authentication.become_method
+        authentication.save!
+      end
+    end
+  end
+end
--- a/db/migrate/20220617140542_move_become_method_from_options.rb
+++ b/db/migrate/20220617140542_move_become_method_from_options.rb
@@ -1,4 +1,19 @@
 class MoveBecomeMethodFromOptions < ActiveRecord::Migration[6.0]
+  ALLOWED_VALUES = %w[
+    sudo
+    su
+    pbrum
+    pfexec
+    doas
+    dzdo
+    pmrun
+    runas
+    enable
+    ksu
+    sesu
+    machinectl
+  ].freeze
+
   class Authentication < ActiveRecord::Base
     include ActiveRecord::IdRegions
     self.inheritance_column = :_type_disabled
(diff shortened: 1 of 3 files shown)


db/migrate/20220617140542_move_become_method_from_options.rb
```diff
@@ -0,0 +1,29 @@
...
+    serialize :options
+  end
+
+  def up
+    say_with_time("Move Authentication become_method out of options text field") do
+      Authentication.in_my_region.find_each do |authentication|
+        next unless authentication.options&.key?(:become_method)
+
+        authentication.become_method = authentication.options.delete(:become_method)
+        authentication.save
```
Fryguy, line 15: Prefer `save!` so failures will blow up, or prefer single-line with `update!`

db/migrate/20220617140542_move_become_method_from_options.rb
```diff
@@ -0,0 +1,29 @@
...
+  class Authentication < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    serialize :options
+  end
+
+  def up
+    say_with_time("Move Authentication become_method out of options text field") do
+      Authentication.in_my_region.find_each do |authentication|
```
kbrock, line 11: can we tack on a where clause?

```suggestion
      Authentication.in_my_region.where("options like '%become_method%'").find_each do |authentication|
```

Fryguy: capital `LIKE`, but I agree.

db/migrate/20220617140542_move_become_method_from_options.rb
```diff
@@ -0,0 +1,49 @@
...
+        authentication.save!
+      end
+    end
+  end
+
+  def down
+    say_with_time("Move Authentication become_method back to options text field") do
+      Authentication.in_my_region.where.not(:become_method => nil).find_each do |authentication|
+        authentication.options = {} if authentication.options.nil?
+        authentication.options[:become_method] = authentication.become_method
```
Fryguy, line 44: Should we also check the ALLOWED on reverse?  Feels like it can't hurt?

db/migrate/20220617140542_move_become_method_from_options.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+    end
+  end
+
+  def down
+    say_with_time("Move Authentication become_method back to options text field") do
+      Authentication.in_my_region.where.not(:become_method => nil).find_each do |authentication|
+        authentication.options = {} if authentication.options.nil?
+
+        authentication.options[:become_method] = authentication.become_method if ALLOWED_VALUES.include?(authentication.become_method)
```
Fryguy, line 45: If authentications.become_method is nil, then this won't set the key, but that's expected, right?  Additionally, if it's nil (or bad), and options was originally nil, then options will turn into an empty Hash, so is that also expected or should options be nil in that case?

db/migrate/20220617140542_move_become_method_from_options.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        authentication.save!
```
Fryguy, line 47: Follow up comment to https://github.com/ManageIQ/manageiq-schema/pull/650/files#r900454286, I'm wondering if this should be:


```suggestion
        if ALLOWED_VALUES.include?(authentication.become_method)
          authentication.options ||= {}
          authentication.options[:become_method] = authentication.become_method
        end

        authentication.save!
```

db/migrate/20220617140542_move_become_method_from_options.rb
agrare: Sure, I was thinking by the time we merge this we'd have validators on this column in core so that wouldn't happen

db/migrate/20220617140542_move_become_method_from_options.rb
agrare: Let me move the `options = {}` after checking allowed values

Fryguy (approved):

Fryguy:
https://github.com/ManageIQ/manageiq/pull/21946 is failing cross-repo-tests here: https://github.com/ManageIQ/manageiq-cross_repo-tests/runs/7031387659?check_suite_focus=true#step:6:260

The offending line of code is https://github.com/ManageIQ/manageiq-schema/blob/41e9d473f0d5d70fa68543a1e07dea9fa04d5922/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb#L325.  I'm wondering if the down migration is causing an issue since become_method is expected to be in the options here?: https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb#L251

Fryguy:
Gah, I'm sorry - bit of a false alarm - the actual line of code is here: https://github.com/ManageIQ/manageiq-schema/blob/41e9d473f0d5d70fa68543a1e07dea9fa04d5922/db/migrate/20190617191109_move_awx_credentials_to_authentications.rb#L321

Not sure how that can happen

Fryguy:
Actually wondering if this is fixed by the PG lockdown.  @jrafanie ?

jrafanie:
> Actually wondering if this is fixed by the PG lockdown. @jrafanie ?

yes, it was fixed by https://github.com/ManageIQ/manageiq/pull/21952
But they released 1.4.1 of pg which fixes the issue so I unlocked the pg dependency to only prevent usage of 1.4.0 in https://github.com/ManageIQ/manageiq/pull/21953

kbrock:
So then is this good to go?

PR #655, opened 2022-07-14
Rename messaging_type settings and default to Kafka

nasark:
- settings `/prototype/messaging_type` renamed to `/messaging_type` and `kafka` set as the default value (previously `miq_queue`)

@miq-bot add_labels enhancement, refactoring
@miq-bot assign @agrare


nasark:
Just a note that https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/message_configuration.rb#L189 will need to be updated once this is merged

nasark:
Closing since this is not necessary

PR #656, opened 2022-07-20
display events for physical storage

TonySchneider:
* Following this PR: https://github.com/ManageIQ/manageiq-providers-autosde/pull/151
This is an implementation of the monitoring page for physical storage where we'll show our critical alerts (storage events).

Before:
image

After:
image

# related PRs:
- https://github.com/ManageIQ/manageiq-providers-autosde/pull/154
- https://github.com/ManageIQ/manageiq/pull/22008
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8373

### DIFF
--- a/db/migrate/20220719151539_add_physical_storage_to_event_streams.rb
+++ b/db/migrate/20220719151539_add_physical_storage_to_event_streams.rb
@@ -1,0 +1,6 @@
+class AddPhysicalStorageToEventStreams < ActiveRecord::Migration[6.0]
+  def change
+    add_column :event_streams, :physical_storage_id, :bigint
+    add_column :event_streams, :physical_storage_name, :string
+  end
+end


db/migrate/20220719151539_add_physical_storage_to_event_streams.rb
```diff
@@ -0,0 +1,5 @@
+class AddPhysicalStorageToEventStreams < ActiveRecord::Migration[6.0]
+  def change
+    add_column :event_streams, :physical_storage_id, :bigint
+  end
```
agrare, line 4: @TonySchneider `process_object_in_event` also sets the `name` so best to add `add_column :event_streams, :physical_storage_name, :string` here as well

agrare (approved):
Looking good

PR #657, opened 2022-07-21
[wip] create request logs for diagnostics

akhilkr128:
[Feature](https://github.com/ManageIQ/manageiq/issues/21188) 

Request and Task Diagnostics


db/migrate/20220512123911_create_request_logs.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRequestLogs < ActiveRecord::Migration[6.0]
+  def change
+    create_table :request_logs do |t|
+      t.references :miq_requests, :type => :bigint
+      t.string :log_message
+      t.string :object_type
+      t.bigint :object_id
```
Fryguy, line 7: 3 comments here:

---

@akhilkr128 What is an `object` in this context?  I'm worried that this is too generic of a term.

---

Think this can be simplified to `t.references :object, :polymorphic => true`, but you'd have to verify.  (@agrare Do you know?)

---

Additionally, I'm wondering if this should have an index on it.

db/migrate/20220512123911_create_request_logs.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRequestLogs < ActiveRecord::Migration[6.0]
+  def change
+    create_table :request_logs do |t|
+      t.references :miq_requests, :type => :bigint
+      t.string :log_message
```
Fryguy, line 5: The `log_` prefix is redundant considering this is the request_logs table.  I think this can just be `message`.

Fryguy:
Thinking about log messages I think we need a few more columns (based on this reference: https://github.com/ManageIQ/manageiq-loggers/blob/1ca9f46b4042a0fdc92aab4346f62b3914062a40/lib/manageiq/loggers/json_logger.rb#L29-L36). 

However maybe some or all of these can wait for a future PR?  @jrafanie Thoughts?

- `severity`
- `pid`
- `tid`
- `request_id`

db/migrate/20220512123911_create_request_logs.rb
agrare: Yes this is the normal pattern we use: `t.references :resource, :type => :bigint, :polymorphic => true`

Fryguy: Ah thanks I missed the type - I'll update my comment / suggestion.

akhilkr128: I planned to give all the below values as strings to object type.
'ServiceAnsiblePlaybook'
'ServiceTemplateProvisionTask'
'ServiceTemplateProvisionRequest'
'ServiceTemplate'
'AutomationTask'
'AutomationRequest'

Fryguy: is that different from the request's type?

akhilkr128: I will try for the polymorphic relation. Will get back.

Fryguy: @akhilkr128 We had discussed maybe replacing the object_type/id with an miq_request_task_id, because I thought it was only MiqRequestTask and MiqRequest values (and we already had an miq_request_id).

However, now I see that this column could contain a number of different objects, namely: Service, MiqRequestTask, MiqRequest, and ServiceTemplate.  So, instead I think we should drop miq_request_id in favor of the polymorphic object_type/object_id.

Only other thing I would say here is use the term `resource` instead of `object`, since that's more consistent with the rest of the code base.

akhilkr128: Sure, make sense.

Fryguy:
I think `severity` needs to be a required field for this PR.  The goal of this is ultimately that users are trying to find their errors, so if they can't filter / search for errors, it's much less useful.

db/migrate/20220808064045_create_request_log.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRequestLog < ActiveRecord::Migration[6.0]
+  def change
+    create_table :request_logs do |t|
+      t.string :log_message
```
Fryguy, line 4: As mentioned above, `log_` is redundant here, so make this `message` only.

jrafanie:
This looks like audit logging of the tasks visible in the UI so you can see all the activity associated with that.

I think that's what we're describing here.  When I've looked at this in the past, the problem was the different types of tasks/requests do things differently so you'll need to learn how each works and ensure they cut records for each step in the process.  I don't know how to enumerate each type to test that each one cuts the right audit records.

db/migrate/20220808064045_create_request_log.rb
akhilkr128: done.

Fryguy (approved):
I think this is a good start

Fryguy:
Closing in favor of #681

PR #658, opened 2022-08-09
Fix IBM PowerVC STI classes

jaywcarman:
Database migrations to cover quite a few subclass fixes to the IBM PowerVC provider:
- ManageIQ/manageiq-providers-ibm_power_vc#52
- ManageIQ/manageiq-providers-ibm_power_vc#53
- ManageIQ/manageiq-providers-ibm_power_vc#55
- ManageIQ/manageiq-providers-ibm_power_vc#56

### DIFF
--- a/db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
+++ b/db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
@@ -1,0 +1,189 @@
+class FixIbmPowerVcStiClasses < ActiveRecord::Migration[6.0]
+  class ExtManagementSystem < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  class CloudNetwork < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class CloudSubnet < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class FloatingIp < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class NetworkPort < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class NetworkRouter < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class SecurityGroup < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class CloudVolume < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
+  class CloudVolumeBackup < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "FixIbmPowerVcStiClasses::ExtManagementSystem"
+  end
+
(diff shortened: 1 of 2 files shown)


db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def up
+    say_with_time("Fixing IBM PowerVC Network STI classes") do
+      network_klass = "ManageIQ::Providers::IbmPowerVc::NetworkManager"
+
+      network_managers = ExtManagementSystem.in_my_region.where(:type => network_klass)
+
+      CloudNetwork.in_my_region
+                  .where(:ext_management_system => network_managers)
+                  .update_all(:type => "#{network_klass}::CloudNetwork")
```
agrare, line 85: @jaywcarman were the types `nil` before or were they e.g. `Openstack::NetworkManager::CloudNetwork...`?

Wondering because of `ManageIQ::Providers::IbmPowerVc::NetworkManager::CloudNetwork::Private` and `ManageIQ::Providers::IbmPowerVc::NetworkManager::CloudNetwork::Public` are the classes which will be actually set

jaywcarman: I think the types were all set, just in the wrong namespace.

The parent ems types are correctly in `ManageIQ::Providers::IbmPowerVc::`, though, so I was trying to select via `.where(:ext_management_system => network_managers)`.

db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def down
+    say_with_time("Resetting IBM PowerVC Network STI classes") do
+      network_klass = "ManageIQ::Providers::IbmPowerVc::NetworkManager"
+
+      network_managers = ExtManagementSystem.in_my_region.where(:type => network_klass)
+
+      CloudNetwork.in_my_region
+                  .where(:ext_management_system => network_managers)
+                  .update_all(:type => nil)
```
jaywcarman, line 131: I'm not sure if these are supposed to be reset to `nil` or back to the `ManageIQ::Providers::Openstack::` types. With `nil` would they be reset by the next refresh update?

db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
agrare: Oh yeah that is fine, it is just for the cloud networks where the types are actually going to be `#{network_klass}::CloudNetwork::Public"` or `#{network_klass}::CloudNetwork::Private"` could we do:

```
      CloudNetwork.in_my_region
                  .where(:ext_management_system => network_managers)
                  .where(:type => "ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork::Public")
                  .update_all(::type => "#{network_klass}::CloudNetwork::Public")
```

agrare: (If they were `nil` we'd have no idea what they _should_ be)

db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
agrare: Ideally a down migration would set them to whatever they were before the up migration.

db/migrate/20220809152005_fix_ibm_power_vc_sti_classes.rb
jaywcarman: I just forced pushed an update with the Public and Private CloudNetwork types included.

agrare (approved):

spec/migrations/20220809152005_fix_ibm_power_vc_sti_classes_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(network_port.reload.type).to          eq("#{ibm_power_vc_network_klass}::NetworkPort")
+      expect(network_router.reload.type).to        eq("#{ibm_power_vc_network_klass}::NetworkRouter")
+      expect(security_group.reload.type).to        eq("#{ibm_power_vc_network_klass}::SecurityGroup")
+    end
+
+    it "Fixes IBM PowerVC Storage (Cinder) STI classes" do
+      ibm_power_vc_cinder_klass = "ManageIQ::Providers::IbmPowerVc::StorageManager::CinderManager"
+      openstack_cinder_klass = "ManageIQ::Providers::Openstack::StorageManager::CinderManager"
+
+      cinder_manager = ems_stub.create(:type => "#{ibm_power_vc_cinder_klass}")
```
agrare, line 48:
```suggestion
      cinder_manager = ems_stub.create(:type => ibm_power_vc_cinder_klass)
```

spec/migrations/20220809152005_fix_ibm_power_vc_sti_classes_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(cloud_subnet.reload.type).to          eq("#{ibm_power_vc_network_klass}::CloudSubnet")
+      expect(floating_ip.reload.type).to           eq("#{ibm_power_vc_network_klass}::FloatingIp")
+      expect(network_port.reload.type).to          eq("#{ibm_power_vc_network_klass}::NetworkPort")
+      expect(network_router.reload.type).to        eq("#{ibm_power_vc_network_klass}::NetworkRouter")
+      expect(security_group.reload.type).to        eq("#{ibm_power_vc_network_klass}::SecurityGroup")
+    end
+
+    it "Fixes IBM PowerVC Storage (Cinder) STI classes" do
+      ibm_power_vc_cinder_klass = "ManageIQ::Providers::IbmPowerVc::StorageManager::CinderManager"
+      openstack_cinder_klass = "ManageIQ::Providers::Openstack::StorageManager::CinderManager"
```
agrare, line 46: Were you going to use this to set the "wrong" types on the records?

agrare:
Ah rubocop found some good stuff

spec/migrations/20220809152005_fix_ibm_power_vc_sti_classes_spec.rb
jaywcarman: oops, copy pasted both of these when only the ibm_power_vc is needed

agrare (approved):

PR #659, opened 2022-08-11
Use rails61 by default

kbrock:
Default tests to rails 6.1

we can delete the 6.0 matrix option if we want.


jrafanie:
Updated https://github.com/ManageIQ/manageiq/issues/21516#issuecomment-1212166941 with this PR in case we need to revert

jrafanie (approved):
I'm good with keeping this for now and as we feel good with rails 6.1, we can drop this conditional

jrafanie:
All tests pass, we're still running the right ruby/rails matrix tests but now default to 6.1 as expected.

PR #661, opened 2022-08-28
added a `fixed` string field to model `event_streams`

OrGur1987:
This field indicates whether the event was fixed or not, but could also hold a more complex status, since it's a string rather than a boolean.

### DIFF
--- a/db/migrate/20220827142531_add_fixed_to_event_streams.rb
+++ b/db/migrate/20220827142531_add_fixed_to_event_streams.rb
@@ -1,0 +1,5 @@
+class AddFixedToEventStreams < ActiveRecord::Migration[6.0]
+  def change
+    add_column :event_streams, :fixed, :string
+  end
+end


db/migrate/20220827142531_add_fixed_to_event_streams.rb
```diff
@@ -0,0 +1,5 @@
+class AddFixedToEventStreams < ActiveRecord::Migration[6.0]
+  def change
+    add_column :event_streams, :fixed, :string
```
agrare, line 3: @OrGur1987 what does `fixed` mean in the context of an event?

agrare (changes requested):

db/migrate/20220827142531_add_fixed_to_event_streams.rb
OrGur1987: it means that the problem which caused the event has been resolved

agrare: Hm that sounds more like a different event type to me, I'm heavily leaning against this change since it isn't relevant for all of the other events that we have

agrare: NOTE we have a "chain_id" if you wanted to associate a "failed" event with a later "fixed" event, you can have the event type like "storage_drive_failure" and then "storage_drive_fixed"

OrGur1987:
`fixed` datum will be prepended to `event_type` field value, no need to change schema

PR #663, opened 2022-09-09
remove ws_cockpit role from /server/roles

kbrock:
Extracted from https://github.com/ManageIQ/manageiq/issues/22097

#613 / aka 01f6a17c had dropped cockpit from the server roles table, but did not remove it from the configuration overrides.

This removes the role from the settings

### DIFF
--- a/db/migrate/20220909190221_drop_cockpit_role.rb
+++ b/db/migrate/20220909190221_drop_cockpit_role.rb
@@ -1,0 +1,15 @@
+class DropCockpitRole < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Deleting cockpit role configuration") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%ws_cockpit%'").select('id', 'value').each do |row|
+        if row.value == 'ws_cockpit'
+          row.delete
+        else
+          row.update(:value => row.value.split(',').tap { |a| a.delete("ws_cockpit") }.join(','))
+        end
+      end
+    end
+  end
+end
--- a/spec/migrations/20220909190221_drop_cockpit_role_spec.rb
+++ b/spec/migrations/20220909190221_drop_cockpit_role_spec.rb
@@ -1,0 +1,23 @@
+require_migration
+
+describe DropCockpitRole do
+  let(:settings_change_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    it "removes roles in the beginning" do
+      only  = settings_change_stub.create!(:key => "/server/role", :value => "ws_cockpit")
+      first = settings_change_stub.create!(:key => "/server/role", :value => "ws_cockpit,database,websocket")
+      middle = settings_change_stub.create!(:key => "/server/role", :value => "database,ws_cockpit,websocket")
+      last  = settings_change_stub.create!(:key => "/server/role", :value => "database,websocket,ws_cockpit")
+      none  = settings_change_stub.create!(:key => "/server/role", :value => "database,websocket")
+
+      migrate
+
+      expect(settings_change_stub.exists?(:id => only.id)).to eq(false)
+      expect(first.reload.value).to eq("database,websocket")
+      expect(middle.reload.value).to eq("database,websocket")
+      expect(last.reload.value).to  eq("database,websocket")
+      expect(none.reload.value).to  eq("database,websocket")
+    end
+  end
+end


db/migrate/20220909190221_drop_cockpit_role.rb
```diff
@@ -0,0 +1,16 @@
...
+  class SettingsChange < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Deleting cockpit role configuration") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%ws_cockpit%'").select('id', 'value').each do |row|
+        if row.value == 'ws_cockpit'
+          row.delete
+        else
+          # if this string is in the middle of a string, add a comma, otherwise, no need
+          row.update(:value => row.value.sub(/(,?)ws_cockpit(,?)/) { ($1 == $2 && $1 == ',') ? ',' : ''})
```
Fryguy, line 11: Is it simpler/easier to read with to just split.delete.join?

```suggestion
          row.update(:value => row.split(",").tap { |r| r.delete("ws_cockpit") }.join(",")))
```

Fryguy: _technically_ it will update rows that don't change, but that's a minor penalty given the small number of rows in the table.

kbrock: I had this before, but the `delete` was a little ugly. I will put it back

kbrock:
update:
- using rails 6.0 migration
- using ruby arrays to remove string from serialized list of roles

db/migrate/20220909190221_drop_cockpit_role.rb
```diff
@@ -0,0 +1,15 @@
+class DropCockpitRole < ActiveRecord::Migration[6.0]
+  class SettingsChange < ActiveRecord::Base; end
+
+  def up
+    say_with_time("Deleting cockpit role configuration") do
+      SettingsChange.where("key = '/server/role' AND value LIKE '%ws_cockpit%'").select('id', 'value').each do |row|
+        if row.value == 'ws_cockpit'
+          row.delete
+        elsif row.value.include?("ws_cockpi")
```
Fryguy, line 9:
```suggestion
        elsif row.value.include?("ws_cockpit")
```

Fryguy: Although I don't thin this check is necessary because you check it at the query level with the `LIKE` clause.

kbrock: wonder why this was green with this bug.

Fryguy:
Minor, but can you also do the rubocops?  Otherwise this LGTM.

bryanfrancois:
Hi there,
I re-open this thread following an upgrade from Morphy to Petrosian yesterday.

The `cockpit_ws` role caused the `evmserverd` service to fail due to residual references in the `SettingsChange` table and the active server configuration (`Settings.server.role`). 

### Cleaned Up Residual Data:

Used the Rails console to query the `SettingsChange` table and deleted the `cockpit_ws` entries:

```ruby
SettingsChange.where("value LIKE ?", "%cockpit_ws%").destroy_all
```
Updated the `Settings.server.role` configuration to remove `cockpit_ws`:
```ruby
Settings.server.role = "event,notifier,remote_console,reporting,user_interface,web_services"
MiqServer.my_server.save
```

Ensured Configuration Consistency:

- Removed `cockpit_ws` references in /var/www/miq/vmdb/log/pg_inspector_server.yml file.
- Re-seeded the database using `bin/rails db:seed`to ensure proper role initialization.

Restarted the Service:

Successfully restarted the `evmserverd` service without further issues.

Fryguy:
Although this PR should have removed it from Settings.server.role value - I wonder if it was still present in the seed file.

Fryguy:
@bryanfrancois Do you know which SettingsChange keys had the cockpit_ws value inside them? This PR should have removed them from /server/role, so you must have hit it somewhere else?

bryanfrancois:
The `SettingsChange` key that had the `cockpit_ws` value was `/server/role`

kbrock:
@bryanfrancois thank you for responding.

Good news is this migration was written correctly and should have removed that record.
But it is curious that this record was not removed.

I heard that you ran the migrations (via `bundle exec rake db:migrate`), but for some reason it did not delete the necessary records.
And I'm also hearing that `pg_inspector_server.yml` was possibly the culprit.

ASIDE: I'm guessing that after deleting the `SettingsChange` record, simply reloading `MiqServer.my_server` via `MiqServer.my_server(true)` would have reloaded the `my_server` record with the proper settings and the `Settings.server.role = ...` was not necessary.

Wonder if there is an issue when attempting to delete a role that is actually active.

@Fryguy ideas where to look?

Fryguy:
I think we need a reproducer

bryanfrancois:
Maybe it could help : our migration path was `Morphy` to `Petrosian` directly.

PR #667, opened 2022-10-02
326047 Add capabilities schema and relations

galoiring:
We added capability and capability values schema. 
W/ @OrGur1987 

main repo PR: https://github.com/ManageIQ/manageiq/pull/22137
provider PR: https://github.com/ManageIQ/manageiq-providers-autosde/pull/182

Screen Shot 2022-10-02 at 18 13 04


ereztheodorou:
@agrare  The capability values table represents all possible values allowed for a specific capability name. so we need multiple values associated with a specific name. and if the name changes, it makes more sense to put it in a separate table vs updated all the records in the capability values table.

agrare:
@ereztheodorou sorry I'm still not seeing it.  Your capabilities jsonb column can have:
```
{"compression" => true, "thin_provision" => false}
```
On whatever record you want to store the capabilities for.

galoiring:
> @galoiring are you trying to show capabilities on inventory records, or present the user with a list of choices? Either way I don't think this is the way to go but need to know the use-case before presenting alternatives.

Firstly I want to present a list of choice , but I prefer @ereztheodorou to address this issue due to more complex design for this feature.

ereztheodorou:
@agrar For example we will possibly need to support a QOS capability in which a key with multiple values associated with it like: capability name: "QOS" value : "gold", value: "sliver", value: "bronze". We need to be able to filter according to the key and get all possible values and be able to associate them. 
And then later on we will need to be able to associate a specific subset of it to a storage.

galoiring:
> @Agrar For example we will possibly need to support a QOS capability in which a key with multiple values associated with it like: capability name: "QOS" value : "gold", value: "sliver", value: "bronze". We need to be able to filter according to the key and get all possible values and be able to associate them.
> And then later on we will need to be able to associate a specific subset of it to a specific storage.

@agrare

agrare:
Okay thanks @galoiring and when you pass the values to AutoSDE do you reference them by value or by ems_ref?

ereztheodorou:
@agrare we pass them by ems_ref (which is their UUID in AutoSDE)

galoiring:
@agrare ?

ereztheodorou:
@agrare Is there anything else needed here? we need to get this PR merged as it is fundamental to our provider and is holding up development.

Fryguy:
In general, we don't store simple enums in tables for a number of reasons (data migrations, backports, and internationalization are a couple that come to mind).  Instead, these enumerated values can easily be hardcoded into the model, where they can be applied in model validations if you want to verify the values before saving, or they can be referenced by the UI for presentation purposes in dropdowns.

PR #670, opened 2022-11-11
Remove VimStrings from CustomAttribute#serialized_value

agrare:
CustomAttribute records created by the old VimBroker style of VMware Refresher would have a serialized_value that includes VimStrings

The streaming refresh model which uses RbVmomi does not suffer the same potential problem but databases upgraded from before this change can still exhibit this issue.

PR #671, opened 2022-11-15
Remove VimString from MiqRequestTask#options

agrare:
There are occasions where a full VM object was serialized to the `#options` column of `MiqRequestTask` prior to the migration to remove the ems_ref_obj serialized column.  This means we still had serialized VimStrings in the options column causing load of these records to fail.

This is not something which is an issue after the migration to drop ems_ref_obj, but old miq_request_tasks could have this problem.

```
name: ems_ref_obj\n      value_before_type_cast: |\n        --- !ruby/string:VimString\n        str: domain-c4006\n        xsiType: :ManagedObjectReference\n        vimType: :ClusterComputeResource\n
```

Similar issue to https://github.com/ManageIQ/manageiq-schema/pull/670

### DIFF
--- a/db/migrate/20221114165219_remove_vim_string_from_miq_request_task_options.rb
+++ b/db/migrate/20221114165219_remove_vim_string_from_miq_request_task_options.rb
@@ -1,0 +1,38 @@
+class RemoveVimStringFromMiqRequestTaskOptions < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+  include MigrationHelper
+
+  class MiqRequestTask < ActiveRecord::Base
+    include ActiveRecord::IdRegions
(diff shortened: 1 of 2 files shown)


Fryguy:
Backported to `oparin` in commit 927644c70ac60062172b8f6846e67e6fc4036941.

```text
commit 927644c70ac60062172b8f6846e67e6fc4036941
Date:   Thu Nov 17 15:46:56 2022 -0500

    Merge pull request #671 from agrare/remove_vim_string_from_miq_request_tasks
    
    Remove VimString from MiqRequestTask#options
    
    (cherry picked from commit e0f61d54a8dc819561564f62cd3524c0f02f92e2)
```

PR #672, opened 2022-11-17
Transfer IsoDatastore records to Storage and drop table

nasark:
- schema changes to facilitate transfer of `:iso_datastore` records to `:storage` table
- iso datastores will be modelled as `Storage` objects with STI `:type => ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore`

Depends on:
- [ ] https://github.com/ManageIQ/manageiq-providers-ovirt/pull/621

@miq-bot add_labels enhancement, technical debt
@miq-bot assign @agrare

### DIFF
--- a/db/migrate/20221107195522_move_existing_iso_datastore_records.rb
+++ b/db/migrate/20221107195522_move_existing_iso_datastore_records.rb
@@ -1,0 +1,67 @@
+class MoveExistingIsoDatastoreRecords < ActiveRecord::Migration[6.1]
+  class ExtManagementSystem < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    has_many :iso_datastores, :foreign_key => :ems_id, :class_name => "MoveExistingIsoDatastoreRecords::IsoDatastore"
+  end
+
+  class Storage < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "MoveExistingIsoDatastoreRecords::ExtManagementSystem"
+    has_many :iso_images, :class_name => "MoveExistingIsoDatastoreRecords::IsoImage"
+  end
+
+  class IsoDatastore < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "MoveExistingIsoDatastoreRecords::ExtManagementSystem"
+    has_many :iso_images, :class_name => "MoveExistingIsoDatastoreRecords::IsoImage"
+  end
+
+  class IsoImage < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :storage, :class_name => "MoveExistingIsoDatastoreRecords::Storage"
+    belongs_to :iso_datastore, :class_name => "MoveExistingIsoDatastoreRecords::IsoDatastore"
+  end
+
+  def up
+    say_with_time("Transfer existing IsoDatastore records to Storage") do
+      add_reference :iso_images, :storage
+
+      IsoDatastore.in_my_region
+                  .joins(:ext_management_system)
+                  .where(:ext_management_systems => {:type => ["ManageIQ::Providers::Ovirt::InfraManager", "ManageIQ::Providers::Redhat::InfraManager"]})
+                  .each do |datastore|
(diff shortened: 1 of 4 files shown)


kbrock (approved):
- failure is legit but trivial - just need to remove the table from that list.
- cops hash/parens is legit byt trivial
- ignore timestamp cop

db/migrate/20221107195522_move_existing_iso_datastore_records.rb
```diff
@@ -0,0 +1,27 @@
...
+    include ActiveRecord::IdRegions
+
+    self.inheritance_column = :_type_disabled
+  end
+
+  class IsoDatastore < ActiveRecord::Base
+  end
+
+  def up
+    datastore_class_name = "ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore"
```
agrare, line 12: We'll need to handle `Ovirt::InfraManager` and `Redhat::InfraManager`.  I imagine the majority will be `Redhat::InfraManager` but we did release with the `Ovirt::InfraManager` so it is possible a customer has some ISO Datastores on this.

You can join on a where like (totally untested):
```
IsoDatastore.in_my_region.
            .joins(:ext_management_system)
            .where(:ext_management_systems => {:type => "ManageIQ::Providers::Ovirt::InfraManager"})
```

kbrock: - Do we think there will be a lot of these? I'm guessing no, but if so, I can help come up with the sql for this.

@agrare did you say we'll need to create 2 possible `datastore_class_name` values based upon the ems?

```ruby
IsoDatastore.in_my_region
            .joins(:ext_management_system)
            .where(:ext_management_systems => {:type => ["ManageIQ::Providers::Ovirt::InfraManager", "ManageIQ::Providers::Redhat::InfraManager"})
            .each do |datastore|
  datastore_class_name = "#{datastore.ems.type}::IsoDatastore"
  Storage.create!(:ems_id => datastore.ems_id, :type => datastore_class_name)
end
```

agrare: > Do we think there will be a lot of these? I'm guessing no, but if so, I can help come up with the sql for this.

Only Ovirt/RHV support IsoDatastores, so there are only two possible values

agrare:
@nasark reminded me about the `IsoImage` model, so we'll need to:
1. Add a `references` to iso_images -> storages
2. Update all iso_images on the old iso_datastore to set the new storage_id column
3. Delete the iso_datastore_id from the iso_images before dropping the iso_datastores table

db/migrate/20221107195522_move_existing_iso_datastore_records.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    IsoDatastore.all.each do |datastore|
```
Fryguy, line 14: Please wrap this and the down version in `say_with_time` prompts.

db/migrate/20221107195559_drop_iso_datastore.rb
```diff
@@ -0,0 +1,12 @@
+class DropIsoDatastore < ActiveRecord::Migration[6.1]
+  def up
+    drop_table :iso_datastores
+  end
+
+  def down
+    create_table :iso_datastores, force: :cascade do |t|
+      t.bigint :ems_id
+      t.datetime :last_refresh_on
```
Fryguy, line 9: Was this table really just 2 columns (and id)? Seems almost pointless 😮

nasark:
Thanks for the feedback all, I have made the changes and would appreciate another look whenever possible.

db/migrate/20221107195559_drop_iso_datastore.rb
agrare: Yes there could only be one per EMS so there wasn't even an ems_ref or name "needed" even though it would be helpful

spec/migrations/20221107195522_move_existing_iso_datastore_records_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  migration_context :up do
+    it "transfers datastore records to storages" do
+      emss = %w[Ovirt Redhat].map do |vendor|
+        ext_management_system_stub.create!(:type => "ManageIQ::Providers::#{vendor}::InfraManager")
+      end
+
+      emss.map do |ems|
+        datastore_stub.create!(:ext_management_system => ems)
+      end
```
agrare, line 16: No need for `.map` if you aren't using the return
```suggestion
      emss.each { |ems| datastore_stub.create!(:ext_management_system => ems) }
```

agrare:
@nasark can we add specs covering the IsoImage changes as well?

agrare (approved):

PR #675, opened 2022-12-01
[WIP] Change resource datatypes in resource_pools from integer to float

alizapeikes:
Resolves issue https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/108

image

### DIFF
--- a/db/migrate/20221201184037_change_resources_to_be_float_in_resource_pools.rb
+++ b/db/migrate/20221201184037_change_resources_to_be_float_in_resource_pools.rb
@@ -1,0 +1,13 @@
+class ChangeResourcesToBeFloatInResourcePools < ActiveRecord::Migration[6.1]
+  def up
+    %i[memory_reserve memory_limit memory_shares cpu_reserve cpu_limit cpu_shares].each do |column|
+      change_column :resource_pools, column, :float
+    end
+  end
+
+  def down
+    %i[memory_reserve memory_limit memory_shares cpu_reserve cpu_limit cpu_shares].each do |column|
+      change_column :resource_pools, column, :integer
+    end
+  end
+end


alizapeikes:
@agrare Can you do a cross-repo test with this PR?

db/migrate/20221201184037_change_resources_to_be_float_in_resource_pools.rb
```diff
@@ -0,0 +1,17 @@
+class ChangeResourcesToBeFloatInResourcePools < ActiveRecord::Migration[6.1]
+  class ResourcePools < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
```
Fryguy, line 5: This is not necessary in the migration because it's not used.

spec/migrations/20221201184037_change_resources_to_be_float_in_resource_pools_spec.rb
```diff
@@ -0,0 +1,43 @@
+require_migration
```
Fryguy, line 1: Since the migration is using Rails built-in methods and nothing custom we don't need a spec at all, so this can be deleted.

Fryguy:
As to whether we should change these to floats at all, I'm not sure - @agrare thoughts?

On the pro side, our metrics table has floats for everything, and the containers table uses floats for request_cpu_cores, limit_cpu_cores, and cpu_cores (but notably _not_ for memory because we store bytes-level values.

On the con side, flavors, hardwares, and vms table all have integers for things like cpu_reserve.  This makes me wonder if we should just change it everywhere, but I don't know the implications of that.

agrare:
> As to whether we should change these to floats at all

Yeah this is the main reason I pinged you @Fryguy.

I'm not sure how changing a column's datatype in a migration will work on a large database, probably will work flawlessly but I don't remember us doing this before.

Looking at the other providers that set this (namely VMware) I think the confusion is in the units.  VMware sets this to MHz, not fractions of whole CPUs which is why an integer is fine until we start aggregating up for metrics.

At a high level converting to a float should be fine for providers which have an integer type (unless we're over MAXFLOAT which I doubt), I do worry that code which expects this to be an integer might break if it is a suddenly a float (e.g. expects to assign to an integer) but if all metrics values are floats that is probably less of a concern.

I do think two providers setting the same column to two different units could be an issue though (MHz vs #CPUs)

agrare:
> I don't think you responded to my other concern which is

Sorry wasn't clear, my point about the different units was in response to this.  If HMC set cpu_reserve to MHz then I don't think that the float would be needed.  There has been talk about a float for the main vm.hardware.cpu_total_cores column since power can do fractional CPUs.

kbrock:
@jaywcarman @Fryguy @agrare may want to revisit if we want to do this across the board or just for power

jaywcarman:
Instead of changing the type we introduced new columns in https://github.com/ManageIQ/manageiq-schema/pull/708, which has been merged.

This PR is no longer needed and can be closed.

PR #676, opened 2022-12-04
Add capability values schema as jsonb field to storage models

galoiring:
We added capability values schema as jsonb field to physical storage, storage family, storage services and storage resources as requested by @agrare . 

Screen Shot 2022-12-04 at 15 39 23

related pr: https://github.com/ManageIQ/manageiq-api/pull/1190

W/ @OrGur1987

### DIFF
--- a/db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
+++ b/db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
@@ -1,0 +1,8 @@
+class AddCapabilitiesToPhysicalStorageModels < ActiveRecord::Migration[6.1]
+  def change
+    add_column :physical_storage_families, :capabilities, :jsonb, :default => {}
+    add_column :physical_storages, :capabilities, :jsonb, :default => {}
+    add_column :storage_resources, :capabilities, :jsonb, :default => {}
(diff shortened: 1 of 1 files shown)


db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
```diff
@@ -0,0 +1,8 @@
+class AddCapabilitiesToPhysicalStorageModels < ActiveRecord::Migration[6.1]
+  def change
+    add_column :physical_storage_families, :abstract_capability_values, :jsonb, :default => {}
+    add_column :physical_storages, :enabled_capability_values, :jsonb, :default => {}
+    add_column :storage_resources, :enabled_capability_values, :jsonb, :default => {}
+    add_column :storage_services, :enabled_capability_values, :jsonb, :default => {}
```
agrare, line 6: I think unless we are planning on storing other types of capabilities here (e.g. are we also going to have a `concrete_capability_values` or a `disabled_capability_values` column) we should keep this simple and call it `capabilities` in line with the other models

galoiring: Field names have changed to capabilities.

agrare (approved):

PR #677, opened 2023-01-10
fixed capability jsonb column default

galoiring:
Change default value from {} to [] so that field will save capabilities as a array.

https://github.com/ManageIQ/manageiq-ui-classic/pull/8584
https://github.com/ManageIQ/manageiq-providers-autosde/pull/195

### DIFF
--- a/db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
+++ b/db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
@@ -1,8 +1,8 @@
 class AddCapabilitiesToPhysicalStorageModels < ActiveRecord::Migration[6.1]
   def change
-    add_column :physical_storage_families, :capabilities, :jsonb, :default => {}
-    add_column :physical_storages, :capabilities, :jsonb, :default => {}
-    add_column :storage_resources, :capabilities, :jsonb, :default => {}
-    add_column :storage_services, :capabilities, :jsonb, :default => {}
+    add_column :physical_storage_families, :capabilities, :jsonb, :default => []
+    add_column :physical_storages, :capabilities, :jsonb, :default => []
+    add_column :storage_resources, :capabilities, :jsonb, :default => []
+    add_column :storage_services, :capabilities, :jsonb, :default => []
   end
 end


galoiring:
@agrare could you address this also please?
Thank you

Fryguy:
@agrare How do we do other capabilities fields?  Ideally we'd keep them all consistent at the model level.

agrare:
@Fryguy all of the other capabilities are:
```
{
  "capability_name" => boolean
}
```

so they are a hash at the top level.

db/migrate/20221116104013_add_capabilities_to_physical_storage_models.rb
```diff
@@ -1,8 +1,8 @@
 class AddCapabilitiesToPhysicalStorageModels < ActiveRecord::Migration[6.1]
   def change
-    add_column :physical_storage_families, :capabilities, :jsonb, :default => {}
-    add_column :physical_storages, :capabilities, :jsonb, :default => {}
-    add_column :storage_resources, :capabilities, :jsonb, :default => {}
-    add_column :storage_services, :capabilities, :jsonb, :default => {}
+    add_column :physical_storage_families, :capabilities, :jsonb, :default => []
+    add_column :physical_storages, :capabilities, :jsonb, :default => []
+    add_column :storage_resources, :capabilities, :jsonb, :default => []
+    add_column :storage_services, :capabilities, :jsonb, :default => []
```
Fryguy, line 6: Since developers likely have already gotten past this, we should create a new migration

Fryguy:
@agrare Yeah that's what I thought.  @galoiring What is the rationale for moving this to an Array?

OrGur1987:
Hi, thing is we collect the capabilities from our backend as an array of hashes: 
```
[
  {:name=>'compression', :value=>'true',:uuid=>'999..'}, ...
]
```
and we need to keep the uuid just like we save `ems_ref` for regular model records, so users could for instance create a new volume with selected capabilities (see PR https://github.com/ManageIQ/manageiq-providers-autosde/pull/194).

We initially wanted to save our capabilities in their own model but you preferred saving them in jsonb fields.

Fryguy:
Right, so that could also easily be:

```rb
{
  'compression' => {:value => 'true', :uuid => '999..'},
  ...
}
```

OrGur1987:
> ```ruby
> 'compression' => {:value => 'true', :uuid => '999..'}
> ```

but we will have both true and false for the same capability, so maybe we need something like:
```ruby
{
'compression' => [{:value => 'true', :uuid => '999..'}, {:value => 'false', :uuid => '888..'}],
'thin_provision => ...
}
```

Fryguy:
Oh, so this is a list of all _possible_ capabilities?  I would have expected that to be stored at the EMS level, not each individual record's level.  I thought this was the current capability for this specific record, which is why I thought it would only have one value.

OrGur1987:
we have capabilities on all levels:

- provider => all the capabilities of all `physical_storage_families` this provider has
- physical_storage_families => all capabilities a specific type of storage supports
- physical_storage => basically the capabilities that its family has, but we would like to give the "super" user an option to choose only some of the capabilities from the list of capabilities of the family, and limit the capabilities on a specific physical_storage. for instance: choose only compression=true, to disable creating non-compressed volumes on this storage.

and this trickles down to resources, volumes and services as well.
@ereztheodorou perhaps we should have a meeting with @Fryguy and @agrare to present this?

OrGur1987:
hi @agrare @Fryguy,
following our conversation last Tuesday, we would like to propose the following schema:
- on the provider level: save the full details as they are received from our backend: an array of hashes, each containing name, value and uuid. it already happens this way right now in `app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb`.
- on the lower levels (physical_storage_families, physical_storages, storage_resources, storage_services, cloud_volumes): save an array of uuids, just like a foreign key field keeps an id, and when textual details are needed for ui - retrieve them from the provider. this will require: 
  - a new migration file 
  - adjustment of `parser/storage_manager.parse_capabilities()` to pluck out the uuids
  - retrieval of textual details in relevant places in the ui repo.

agrare:
Hi @OrGur1987 what we were proposing was:

```
ExtManagementSystem.capabilities = {
"compression" => [{"uuid" => "29ac2ba7-1e55-411e-9828-472c935bb882", "value" => "True"}],
"thin_provision => [{"uuid" => "63a2dfbe-bb1b-408c-8e94-36c9d7791df2", "value" => "True"}]
}
```

Where the key is the name of the feature then you have the array of possible values.  On items where you want to say which specific feature is enabled not just what is available you could have just

```
CloudVolume.capabilities = {
  "compression" => "True",
  "thin_provision" => "False"
}
```

And if need be look up the values in the ext_management_system capabilities but this would make it very easy for the UI to display without going to other models.  If we're going to be looking up uuids by value your ext_management_system hash could even be e.g. `"compression" => {"True" => "29ac2ba7-1e55-411e-9828-472c935bb882", "False" => "something-else"}`

WDYT

OrGur1987:
> Hi @OrGur1987 what we were proposing was:
> 
> ```
> ExtManagementSystem.capabilities = {
> "compression" => [{"uuid" => "29ac2ba7-1e55-411e-9828-472c935bb882", "value" => "True"}],
> "thin_provision => [{"uuid" => "63a2dfbe-bb1b-408c-8e94-36c9d7791df2", "value" => "True"}]
> }
> ```
> 
> Where the key is the name of the feature then you have the array of possible values. On items where you want to say which specific feature is enabled not just what is available you could have just
> 
> ```
> CloudVolume.capabilities = {
>   "compression" => "True",
>   "thin_provision" => "False"
> }
> ```
> 
> And if need be look up the values in the ext_management_system capabilities but this would make it very easy for the UI to display without going to other models. If we're going to be looking up uuids by value your ext_management_system hash could even be e.g. `"compression" => {"True" => "29ac2ba7-1e55-411e-9828-472c935bb882", "False" => "something-else"}`
> 
> WDYT

I see, we'll give it a go next week and try to implement this throughout our provider

galoiring:
new pr was opened with the relevant changes:
Provider : https://github.com/ManageIQ/manageiq-providers-autosde/pull/213
UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/8656

galoiring:
@agrare @Fryguy following our conversation,
new pr was opened with the relevant changes:
Provider : https://github.com/ManageIQ/manageiq-providers-autosde/pull/213
UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/8656

PR #679, opened 2023-01-18
Add ManageIQ automate workflow tables

agrare:
Add tables for workflow and workflow instances.

https://github.com/ManageIQ/manageiq/issues/22311

### DIFF
--- a/db/migrate/20230117190802_create_workflow_migrations.rb
+++ b/db/migrate/20230117190802_create_workflow_migrations.rb
@@ -22,6 +22,7 @@ class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
       t.jsonb :workflow_content, :default => {}
       t.jsonb :credentials, :default => {}
       t.jsonb :context, :default => {}
+      t.jsonb :output, :default => {}
 
       t.timestamps
     end
--- a/db/migrate/20230117190802_create_workflow_migrations.rb
+++ b/db/migrate/20230117190802_create_workflow_migrations.rb
@@ -16,6 +16,7 @@ class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
       t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
       t.references :workflow, :type => :bigint, :index => true
       t.references :tenant,   :type => :bigint, :index => true
+      t.references :miq_task, :type => :bigint, :index => true
       t.string :type, :index => true
       t.string :userid
       t.string :status
--- a/db/migrate/20230117190802_create_workflow_migrations.rb
+++ b/db/migrate/20230117190802_create_workflow_migrations.rb
@@ -4,6 +4,8 @@ class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
       t.references :ems,           :type => :bigint, :index => true, :references => :ext_management_system
       t.references :tenant,        :type => :bigint, :index => true
       t.references :git_reference, :type => :bigint, :index => true
+      t.references :evm_owner,     :type => :bigint, :index => true, :references => :user
+      t.references :miq_group,     :type => :bigint, :index => true
       t.string :type, :index => true
       t.string :name
       t.jsonb :workflow_content, :default => {}
@@ -16,6 +18,8 @@ class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
       t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
       t.references :workflow, :type => :bigint, :index => true
       t.references :tenant,   :type => :bigint, :index => true
+      t.references :evm_owner,     :type => :bigint, :index => true, :references => :user
+      t.references :miq_group,     :type => :bigint, :index => true
       t.references :miq_task, :type => :bigint, :index => true
       t.string :type, :index => true
       t.string :userid
--- a/db/migrate/20230117190802_create_workflow_migrations.rb
+++ b/db/migrate/20230117190802_create_workflow_migrations.rb
@@ -15,7 +15,7 @@ class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
     create_table :workflow_instances do |t|
       t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
       t.references :workflow, :type => :bigint, :index => true
-      t.references :tenant
+      t.references :tenant,   :type => :bigint, :index => true
       t.string :type, :index => true
       t.string :userid
       t.string :status
--- a/db/migrate/20230117190802_create_workflow_migrations.rb
+++ b/db/migrate/20230117190802_create_workflow_migrations.rb
@@ -1,0 +1,29 @@
+class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :workflows do |t|
+      t.references :ems,           :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :tenant,        :type => :bigint, :index => true
+      t.references :git_reference, :type => :bigint, :index => true
+      t.string :type, :index => true
+      t.string :name
+      t.jsonb :workflow_content, :default => {}
+      t.jsonb :credentials, :default => {}
+
+      t.timestamps
+    end
+
+    create_table :workflow_instances do |t|
+      t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :workflow, :type => :bigint, :index => true
+      t.references :tenant
+      t.string :type, :index => true
+      t.string :userid
+      t.string :status
+      t.jsonb :workflow_content, :default => {}
+      t.jsonb :credentials, :default => {}
+      t.jsonb :context, :default => {}
+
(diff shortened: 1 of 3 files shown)


db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,26 @@
+class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :workflows do |t|
+      t.references :ems,           :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :tenant,        :type => :bigint, :index => true
+      t.references :git_reference, :type => :bigint, :index => true
+      t.string :type, :index => true
+      t.jsonb :payload, :default => {}
```
Fryguy, line 8: Not sure about the name payload, but can't come up with anything better.

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.timestamps
+    end
+
+    create_table :workflow_instances do |t|
+      t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :workflow, :type => :bigint, :index => true
+      t.references :tenant
+      t.string :type, :index => true
+      t.string :userid
+      t.string :state
```
Fryguy, line 19: Is the actual current step?  or something else like "running"?  How is this different from status? (I'm ok with the names, but state/status has always been confusing...wonder if "step" might be a clearer differentiator).

agrare: Oh yeah `state` is confusing, this is not the current state this matches the miq_request_task state/status so state would be "pending", "running", "finished" and status is "error", "warning"

We could probably collapse into a single "status" though where "queued", "running" are the "active" statuses and once finished "success", "warning", "error" are the "completed" statuses

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: yeah I was going to call it `t.jsonb :workflow` but `workflow.workflow` felt repetitive repetitive

agrare:
@Fryguy taking this out of WIP, think we can start formally reviewing it now

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,31 @@
...
+      t.jsonb :payload, :default => {}
+      t.jsonb :credentials, :default => {}
+
+      t.timestamps
+    end
+
+    create_table :workflow_instances do |t|
+      t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :workflow, :type => :bigint, :index => true
+      t.references :tenant,   :type => :bigint, :index => true
```
kbrock, line 18: The current code base is half way between group ownership and tenant ownership. Could we add miq group if here as well? (All tables)

kbrock (commented):
This is great stuff

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: :+1: great point I will add

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
+class CreateWorkflowMigrations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :workflows do |t|
+      t.references :ems,           :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :tenant,        :type => :bigint, :index => true
+      t.references :git_reference, :type => :bigint, :index => true
```
Fryguy, line 6: What was this again?  Are we using this? Wondering if we should add git stuff later in a separate set of commits for that feature.

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.references :evm_owner,     :type => :bigint, :index => true, :references => :user
+      t.references :miq_group,     :type => :bigint, :index => true
```
Fryguy, line 8: Do we need tenant here?

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.string :type, :index => true
```
Fryguy, line 9: Are we expecting STI with Workflows?

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    create_table :workflow_instances do |t|
+      t.references :ems,      :type => :bigint, :index => true, :references => :ext_management_system
+      t.references :workflow, :type => :bigint, :index => true
+      t.references :tenant,   :type => :bigint, :index => true
+      t.references :evm_owner,     :type => :bigint, :index => true, :references => :user
+      t.references :miq_group,     :type => :bigint, :index => true
+      t.references :miq_task, :type => :bigint, :index => true
+      t.string :type, :index => true
```
Fryguy, line 24: Same here - do we need type?

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: Yeah this was to support creating/updating these from a git repo.  I'll drop for now

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.string :userid
```
Fryguy, line 25: Is this different from evm_owner?

db/migrate/20230117190802_create_workflow_migrations.rb
Fryguy: Is payload the starting workflow content or the starting context content? Wonder if the word `initial_?` would be useful here.

agrare: Payload is the actual ASL JSON

db/migrate/20230117190802_create_workflow_migrations.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.string :status
+      t.jsonb :payload, :default => {}
```
agrare, line 27: NOTE this is a copy of the Workflow#payload above so we know what the ASL was when this ran in the case where it changes in the original Workflow record over time

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: I thought we needed both per this comment https://github.com/ManageIQ/manageiq-schema/pull/679/files#r1132272597

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: I am subclassing these in the provider plugin so the runtime logic isn't in core.

db/migrate/20230117190802_create_workflow_migrations.rb
agrare: I added evm_owner because it was in the OwnershipMixin but maybe we don't need both.  @kbrock @Fryguy what do you recommend for tracking ownership?  I'm used to seeing `userid` on things like miq_task and not evm_owner.  I also don't see `miq_group_id` used very much where tenant_id is used more.

db/migrate/20230117190802_create_workflow_migrations.rb
Fryguy: Oh sorry - didn't see it 3 lines up

db/migrate/20230117190802_create_workflow_migrations.rb
Fryguy: I'm ok with keeping payload, but wondering if `initial_workflow`, `workflow_content`, `initial_workflow_content`, or something like that would be clearer.

agrare: So the only reason I don't like `initial_*` is because the actual workflow json payload doesn't change during the execution.  I agree I don't love `payload` so no issues with changing the name.

agrare: I'll change it to `workflow_content` and we can throw ideas around if there is a better name than that.

Fryguy (approved):
LGTM - also want @kbrock to review.

agrare:
@kbrock can you take a look?  Want some insight into what we need (tenant/miq_group/evm_owner/userid) to use for RBAC with other collections like authentications.

I'm thinking we want at least tenant+miq_group (maybe evm_owner for who created the workflow), and userid is the user that is _running_ the workflow_instance

Fryguy:
Merging now to get this in.  We will probably have to verify a lot of RBAC cases in tests anyway and that will suss out any column changes we may need later.

PR #680, opened 2023-01-23
[WIP] use be_deleted instead of catching RecordNotFound

kbrock:
I'm a fan of [be_deleted](https://github.com/ManageIQ/manageiq/blob/master/spec/custom_matchers/be_deleted_spec.rb)

this change does nothing and drop if you don't like

### DIFF
--- a/spec/migrations/20180718132840_remove_transformation_product_setting_spec.rb
+++ b/spec/migrations/20180718132840_remove_transformation_product_setting_spec.rb
@@ -10,7 +10,7 @@ describe RemoveTransformationProductSetting do
 
       migrate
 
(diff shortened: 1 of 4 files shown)


Fryguy:
I like it but tests are failing?

kbrock:
ok. looks like we didn't include `manageiq/spec/shared/supports/**/*.rb`

Please double check that we want to do this. If not, then I'll come up with another way of using `be_deleted` in here.

kbrock:
ugh. yea, having trouble finding manageiq shared support. think that repo is far far away.

if we want this in a gem, not sure where it would go

kbrock:
WIP: darn. need to find a way to take this out of shared support and make available to other repos.

kbrock:
was worth a try. oh well

PR #685, opened 2023-03-22
Delete any Foreman Provisioning SettingsChanges

agrare:
We do not use the Forman ProvisioningManager RefreshWorker any longer so we should delete any associated settings_changes.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-foreman/pull/113

### DIFF
--- a/spec/migrations/20230322171110_delete_foreman_provisioning_refresh_settings_spec.rb
+++ b/spec/migrations/20230322171110_delete_foreman_provisioning_refresh_settings_spec.rb
@@ -1,0 +1,19 @@
+require_migration
+
+describe DeleteForemanProvisioningRefreshSettings do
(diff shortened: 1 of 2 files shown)


spec/migrations/20230322171110_delete_foreman_provisioning_refresh_settings_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_migration
+
+describe DeleteForemanProvisioningRefreshSettings do
+  let(:settings_change_stub) { migration_stub(:SettingsChange) }
+
+  migration_context :up do
+    it "Deletes Fpreman ProvisioningManager Refresh settings" do
```
jrafanie, line 7: typo: Fpreman

jrafanie:
Just waiting on 💚

PR #687, opened 2023-04-04
Add a filename and path to workflows

agrare:
Ref: https://github.com/ManageIQ/manageiq-ui-classic/issues/8715#issuecomment-1496181147

@Fryguy not sure about the naming, maybe `source_filename` and `source_path` ?

### DIFF
--- a/db/migrate/20230404171121_add_path_and_filename_to_workflows.rb
+++ b/db/migrate/20230404171121_add_path_and_filename_to_workflows.rb
@@ -1,0 +1,6 @@
+class AddPathAndFilenameToWorkflows < ActiveRecord::Migration[6.1]
+  def change
+    add_column :workflows, :file_path, :string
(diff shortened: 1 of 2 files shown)


agrare:
Yes this is for git-based workflows, if they are not using git then these would be blank and the name would be the only present component for uniqueness.  In this case we would be the single-source-of-truth and we can control name collisions prior to saving.

Fryguy:
> we can control name collisions prior to saving.

That's a good point. On that note, do we have a distinction between name and description in that table?

agrare:
> On that note, do we have a distinction between name and description in that table?

No description currently, so proposal is:
1. Description becomes the workflow `"Comment"` (which is currently being set as name)
2. Name becomes unique, and is either "repo/branch/file_path" for git-based or just a name locally authored workflows?

PR #688, opened 2023-04-07
Use configuration_scripts table for Workflows

agrare:
Rather than introduce new workflows/workflow_instances tables reuse the existing configuration_scripts table that ansible_tower and embedded_ansible use.

Dependents:
- [ ] https://github.com/ManageIQ/manageiq/pull/22477
- [ ] https://github.com/ManageIQ/manageiq-providers-workflows/pull/12

https://github.com/ManageIQ/manageiq/issues/22311

### DIFF
--- a/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
+++ b/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
@@ -3,6 +3,7 @@ class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
     change_table :configuration_scripts do |t|
       t.string     :run_by_userid
       t.jsonb      :payload
+      t.string     :payload_type
       t.jsonb      :credentials
       t.jsonb      :context
       t.jsonb      :output
--- a/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
+++ b/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
@@ -2,7 +2,7 @@ class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
   def change
     change_table :configuration_scripts do |t|
       t.string     :run_by_userid
-      t.jsonb      :payload
+      t.string     :payload
       t.string     :payload_type
       t.jsonb      :credentials
       t.jsonb      :context
--- a/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
+++ b/db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
@@ -1,0 +1,13 @@
+class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    change_table :configuration_scripts do |t|
+      t.string     :run_by_userid
+      t.jsonb      :payload
+      t.jsonb      :credentials
+      t.jsonb      :context
+      t.jsonb      :output
+      t.string     :status
+      t.references :miq_task
+    end
+  end
+end


db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
```diff
@@ -0,0 +1,12 @@
+class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :configuration_scripts, :miq_group,   :type => :bigint, :index => true
+    add_column    :configuration_scripts, :userid,      :string
+    add_column    :configuration_scripts, :payload,     :jsonb
+    add_column    :configuration_scripts, :credentials, :jsonb
+    add_column    :configuration_scripts, :context,     :jsonb
+    add_column    :configuration_scripts, :output,      :jsonb
+    add_column    :configuration_scripts, :status,      :string
+    add_reference :configuration_scripts, :miq_task,    :type => :bigint, :index => true
```
kbrock, line 10:
```suggestion
    change_table :configuration_scripts |t|
      t.references :miq_group
      t.string     :userid
      t.jsonb      :payload
      t.jsonb      :credentials
      t.jsonb      :context
      t.jsonb      :output
      t.string     :status
      t.references :miq_task
    end
```

Pretty sure `references` defaults to `:type => :bigint`, and `:index => true`
It does on 7.x, probably on 6.x

kbrock: I was unsure if we wanted `user_id` or `userid`, but we seem to use both everywhere, so not sure it makes a difference.

db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
```diff
@@ -0,0 +1,11 @@
+class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    add_column    :configuration_scripts, :payload,       :jsonb
+    add_column    :configuration_scripts, :credentials,   :jsonb
+    add_column    :configuration_scripts, :context,       :jsonb
+    add_column    :configuration_scripts, :output,        :jsonb
+    add_column    :configuration_scripts, :status,        :string
+    add_reference :configuration_scripts, :miq_task,      :type => :bigint, :index => true
+    add_column    :configuration_scripts, :run_by_userid, :string
```
agrare, line 9: Renamed this to `run_by_userid` from just `userid` so it is more obvious what this column is

db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
Fryguy: userid is the string, user_id is the integer - in this case we want the string.

db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
```diff
@@ -0,0 +1,14 @@
+class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    change_table :configuration_scripts do |t|
+      t.string     :run_by_userid
+      t.jsonb      :payload
+      t.string     :payload_type
```
agrare, line 6: @Fryguy we talked about having a payload_type so the UI can generically display this

db/migrate/20230407170430_add_payload_credentials_configuration_scripts.rb
```diff
@@ -0,0 +1,14 @@
+class AddPayloadCredentialsConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    change_table :configuration_scripts do |t|
+      t.string     :run_by_userid
+      t.jsonb      :payload
```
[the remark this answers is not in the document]
agrare: Change to a string column

PR #691, opened 2023-05-22
Converge batch_size and concurrent_requests

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/22510

2 settings concepts have been merged

Had originally intended to keep the batch_size, but there wasn't a good way to keep it
and avoid duplicates from the /performance/concurrent_requests keys.
It is so new that we are ignoring it for now. No one would have populated the value.

The values that are currently used are 20,1,1 -- which are way low.
locally I use 250,100,100 -- so may want to keep that in mind

### DIFF
--- a/db/migrate/20230522194044_move_concurrent_requests_settings_to_vmware.rb
+++ b/db/migrate/20230522194044_move_concurrent_requests_settings_to_vmware.rb
@@ -1,0 +1,25 @@
+class MoveConcurrentRequestsSettingsToVmware < ActiveRecord::Migration[6.1]
+  class SettingsChange < ActiveRecord::Base
+  end
+
+  def up
+    # global concurrent_requests
+    #   only currently read by vmware
+    #
+    # rename "/performance/concurrent_requests/{realtime,historical,}" =>
+    #        "/ems/ems_vmware/concurrent_requests/*"
+    say_with_time("Moving global concurrent requests to provider") do
+      SettingsChange.where("key LIKE ?", "/performance/concurrent_requests/%")
+                    .update_all("key = REPLACE(key, '/performance/concurrent_requests/', '/ems/ems_vmware/concurrent_requests/')")
+    end
+  end
+
+  def down
+    # rename "/ems/ems_vmware/concurrent_requests/*" =>
+    #        "/performance/concurrent_requests/{realtime,historical,}"
+    say_with_time("Moving provider concurrent requests to global") do
+      SettingsChange.where("key LIKE ?", "/ems/ems_vmware/concurrent_requests/%")
+                    .update_all("key = REPLACE(key, '/ems/ems_vmware/concurrent_requests/', '/performance/concurrent_requests/')")
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20230522194044_move_concurrent_requests_settings_to_vmware.rb
```diff
@@ -0,0 +1,40 @@
...
+  def up
+    # capture_batch_size
+    #   introduced d1b21ac5444ab07c 10/2022
+    #   only currently populated for vmware
+    #
+    # merged into concurrent_requests
+    #
+    # rename "/ems/ems_*/capture_batch_size" =>
+    #        "/ems/ems_*/concurrent_requests/realtime"
+    # update_all("key = replace(key, 'capture_batch_size', 'concurrent_requests/realtime')")
```
Fryguy, line 15: I don't understand this comment - it says to update, but instead the code below deletes?

db/migrate/20230522194044_move_concurrent_requests_settings_to_vmware.rb
```diff
@@ -0,0 +1,40 @@
...
+      SettingsChange.where("key LIKE ?", "/performance/concurrent_requests/%")
+                    .update_all("key = replace(key, '/performance/concurrent_requests/', '/ems/ems_vmware/concurrent_requests/')")
+    end
+  end
+
+  def down
+    say_with_time("Moving provider concurrent requests to global") do
+      # this is only used by vmware. And it has not been out long enough for people to use it.
+      # "/ems/ems_*/capture_batch_size" => "/ems/ems_*/concurrent_requests/realtime"
+      SettingsChange.where("key like ?", "/ems/ems_vmware/concurrent_requests/%")
```
Fryguy, line 36:
```suggestion
      SettingsChange.where("key LIKE ?", "/ems/ems_vmware/concurrent_requests/%")
```

db/migrate/20230522194044_move_concurrent_requests_settings_to_vmware.rb
```diff
@@ -0,0 +1,40 @@
...
+    #
+    # merged into concurrent_requests
+    #
+    # rename "/ems/ems_*/capture_batch_size" =>
+    #        "/ems/ems_*/concurrent_requests/realtime"
+    # update_all("key = replace(key, 'capture_batch_size', 'concurrent_requests/realtime')")
+    say_with_time("Removing old capture_batch_size keys") do
+      SettingsChange.where("key LIKE ?", "/ems/ems_vmware/capture_batch_size")
+                    .destroy_all
+    end
```
agrare, line 19: @kbrock at the point this schema change will run we won't have any settings keys like `/ems/ems_vmware/capture_batch_size` right?

I think we only need to rename `/performance/concurrent_requests/*` to `/ems/ems_vmware/concurrent_requests/*` because vmware was the only provider using this.

kbrock: ok, removed. thanks

kbrock:
update:

- removed `serialize :value` since it is unused
- made SQL uppercase
- dropped our changes to `capture_batch_size`

PR #696, opened 2023-07-10
Drop old_container_project_id

kbrock:
depends:
- https://github.com/ManageIQ/manageiq/pull/22610

### DIFF
--- a/db/migrate/20230709181609_drop_archived_container_projects.rb
+++ b/db/migrate/20230709181609_drop_archived_container_projects.rb
@@ -1,0 +1,18 @@
+class DropArchivedContainerProjects < ActiveRecord::Migration[6.1]
+  class ContainerGroup < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    ContainerGroup.where.not(:old_container_project_id => nil)
+                  .where(:container_project_id => nil)
+                  .update_all('container_project_id = old_container_project_id')
+  end
+
+  def down
+    # Sorry. No down available (but not a problem)
+
+    # The new format (with container_project_id assigned)
+    # is valid before and after this migration and has been for years.
+  end
+end
--- a/spec/migrations/20230709181609_drop_archived_container_projects_spec.rb
+++ b/spec/migrations/20230709181609_drop_archived_container_projects_spec.rb
@@ -1,0 +1,30 @@
+require_migration
+
+class DropArchivedContainerProjects
+  class ContainerProject < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+end
+
+describe DropArchivedContainerProjects do
+  let(:container_group_stub) { migration_stub(:ContainerGroup) }
+  let(:container_project_stub) { migration_stub(:ContainerProject) }
+
+  migration_context :up do
+    it "reassigns container_project_ids" do
+      cp_del_old = container_group_stub.create!(:deleted_on => Time.now.utc - 10.years)
+      cp_del = container_group_stub.create!(:deleted_on => Time.now.utc)
+      cp = container_group_stub.create!
+
+      touched1  = container_group_stub.create!(:container_project_id => nil, :old_container_project_id => cp_del_old.id, :deleted_on => Time.now.utc)
+      touched2  = container_group_stub.create!(:container_project_id => cp_del.id, :old_container_project_id => cp_del.id, :deleted_on => Time.now.utc)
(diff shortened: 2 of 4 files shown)


db/migrate/20230709181609_drop_archived_container_projects.rb
```diff
@@ -0,0 +1,12 @@
+class DropArchivedContainerProjects < ActiveRecord::Migration[6.1]
+  class ContainerProject < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    ContainerProject.where.not(:old_container_group_id => nil)
+                    .where(:container_group_id => nil)
+                    .update_all('container_group_id = old_container_group_id')
```
agrare, line 9: I think you want `ContainerGroup.where_not(:old_container_project_id => nil)...etc`

db/migrate/20230709181609_drop_archived_container_projects.rb
```diff
@@ -0,0 +1,11 @@
+class DropArchivedContainerProjects < ActiveRecord::Migration[6.1]
+  class ContainerGroup < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    ContainerGroup.where.not(:old_container_project_id => nil)
+                  .where(:container_project_id => nil)
+                  .update_all('container_project_id = old_container_project_id')
```
agrare, line 9:
```suggestion
    ContainerGroup.where.not(:old_container_project_id => nil)
                        .where(:container_project_id => nil)
                        .update_all('container_project_id = old_container_project_id')
```

spec/migrations/20230709181609_drop_archived_container_projects_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_migration
+
+describe DropArchivedContainerProjects do
+  let(:container_group_stub) { migration_stub(:ContainerGroup) }
+
+  migration_context :up do
+    it "reassigns container_project_ids"
+      untouched = container_group_stub.create!(:container_project_id => 1)
+      touched1  = container_group_stub.create!(:container_project_id => 3, :old_container_project_id => 3, :deleted_on => Time.now.utc)
+      touched2  = container_group_stub.create!(:container_project_id => nil, :old_container_project_id => 4, :deleted_on => Time.now.utc)
```
agrare, line 10: Probably doesn't matter, but could you create container_project records for these allowing the ID to be created in the right region versus hard-coded IDs?

kbrock:
update:
- tests passing locally (ugh - sorry)
- added project so the ids are reasonable
- fixed missed column name (double ugh)

db/migrate/20230709181609_drop_archived_container_projects.rb
```diff
@@ -0,0 +1,15 @@
+class DropArchivedContainerProjects < ActiveRecord::Migration[6.1]
+  class ContainerGroup < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  class ContainerProject < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
```
agrare, line 8: Minor but this class isn't used here and could be moved into the spec file since it is only being used for tests

db/migrate/20230709181609_drop_archived_container_projects.rb
```diff
@@ -0,0 +1,18 @@
...
+    ContainerGroup.where.not(:old_container_project_id => nil)
+                  .where(:container_project_id => nil)
+                  .update_all('container_project_id = old_container_project_id')
+  end
+
+  def down
+    # Sorry. No down available (but not a problem)
+
+    # The new format (with container_project_id assigned)
+    # is valid before and after this migration and has been for years.
```
agrare, line 16: So we probably could set `old_container_project_id = container_project_id` where deleted_on != null ... is there any case I'm missing where that wouldn't be right?

db/migrate/20230709181609_drop_archived_container_projects.rb
```diff
@@ -0,0 +1,18 @@
+class DropArchivedContainerProjects < ActiveRecord::Migration[6.1]
+  class ContainerGroup < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    ContainerGroup.where.not(:old_container_project_id => nil)
```
Fryguy, line 7: Should be wrapped in say_with_time

PR #699, opened 2023-08-04
Add container indexes

kbrock:
These are focused on speeding up the container related screens. These are also focused on the current implementation of the associations and their use of deleted_on IS NULL.

We probably want to circle back once our archived patterns change.

These are even more noticeable after we add a few virtual attributes that exercise these relationships in a non-N+1 way.

### DIFF
--- a/db/migrate/20230803225825_container_indexes.rb
+++ b/db/migrate/20230803225825_container_indexes.rb
@@ -1,0 +1,25 @@
+class ContainerIndexes < ActiveRecord::Migration[6.1]
+  def change
+    # container projects page
+    add_index :container_routes, :container_project_id
+    add_index :container_services, :container_project_id
+    # didn't show much, since only had 1 record
+    add_index :container_replicators, :container_project_id
+
+    add_index :container_groups, [:container_project_id, :id], :name => "index_container_groups_on_cpid_id_not_del", :where => "deleted_on IS NULL"
+    # container providers screen
+
+    # NOTE: this gets index only scans, but has a larger index size
+    add_index :taggings, [:taggable_type, :taggable_id, :tag_id], :name => "index_taggings_on_type_id_id"
+    remove_index :taggings, :column => [:taggable_id, :taggable_type], :name => "index_taggings_on_taggable_id_and_taggable_type"
+
+    # container Services
+    add_index :container_groups_container_services, [:container_service_id, :container_group_id], :name => "index_container_groups_on_csi_cgi"
+
+    # container groups
+    add_index :containers, [:container_group_id, :state]
+
+    # container nodes
+    add_index :container_conditions, [:container_entity_type, :container_entity_id, :name], :name => "index_container_conditions_on_cet_ceid_name"
+  end
+end


db/migrate/20230803225825_container_indexes.rb
```diff
@@ -0,0 +1,25 @@
+class ContainerIndexes < ActiveRecord::Migration[6.1]
+  def change
+    # container projects page
+    add_index :container_routes, :container_project_id
+    add_index :container_services, :container_project_id
+    # didn't show much, since only had 1 record
```
Fryguy, line 6: This comment doesn't help future readers, so let's remove the comment.  It's fine in the commit message and/or PR comments though)

db/migrate/20230803225825_container_indexes.rb
```diff
@@ -0,0 +1,25 @@
...
+    add_index :container_replicators, :container_project_id
+
+    add_index :container_groups, [:container_project_id, :id], :name => "index_container_groups_on_cpid_id_not_del", :where => "deleted_on IS NULL"
+    # container providers screen
+
+    # NOTE: this gets index only scans, but has a larger index size
+    add_index :taggings, [:taggable_type, :taggable_id, :tag_id], :name => "index_taggings_on_type_id_id"
+    remove_index :taggings, :column => [:taggable_id, :taggable_type], :name => "index_taggings_on_taggable_id_and_taggable_type"
+
+    # container Services
```
Fryguy, line 16: consistent caps

```suggestion
    # container services
```

db/migrate/20230803225825_container_indexes.rb
```diff
@@ -0,0 +1,25 @@
...
+    add_index :container_services, :container_project_id
+    # didn't show much, since only had 1 record
+    add_index :container_replicators, :container_project_id
+
+    add_index :container_groups, [:container_project_id, :id], :name => "index_container_groups_on_cpid_id_not_del", :where => "deleted_on IS NULL"
+    # container providers screen
+
+    # NOTE: this gets index only scans, but has a larger index size
+    add_index :taggings, [:taggable_type, :taggable_id, :tag_id], :name => "index_taggings_on_type_id_id"
+    remove_index :taggings, :column => [:taggable_id, :taggable_type], :name => "index_taggings_on_taggable_id_and_taggable_type"
```
Fryguy, line 14: I don't understand this at all.  Typically pattern for indexes is to put the highest cardinality first, unless there are queries where the lower cardinality one can stand alone.  Also I don't understand adding tag_id

db/migrate/20230803225825_container_indexes.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    # container Services
+    add_index :container_groups_container_services, [:container_service_id, :container_group_id], :name => "index_container_groups_on_csi_cgi"
+
+    # container groups
+    add_index :containers, [:container_group_id, :state]
+
+    # container nodes
+    add_index :container_conditions, [:container_entity_type, :container_entity_id, :name], :name => "index_container_conditions_on_cet_ceid_name"
```
Fryguy, line 23: I would expect id first.

kbrock: The virtual attribute that this is targeting is:

```
  (SELECT container_conditions.status
  FROM container_conditions
  WHERE container_conditions.name = 'Ready'
  AND container_conditions.container_entity_id = container_nodes.id
  AND container_conditions.container_entity_type = 'ContainerNode')
  AS ready_condition_status,
```

When we have 2 indexes, one with `id` first and the other with `type` first, the analyzer prefers the one with `id` first. When I delete the `id` first one, it is still able to use the `type` first one, and the query stats are similar.

Having `name, type, id` or `WHERE name = 'READY'` would be faster, but this way, the index can also be used for all `container_conditions`, not just the `ready` ones.

db/migrate/20230803225825_container_indexes.rb
Fryguy: I'm not against adding the tag_id - I'm questiong why taggable_type and taggable_id order were reversed - that's the part I'm questioning.

db/migrate/20230803225825_container_indexes.rb
Fryguy: > When we have 2 indexes, one with id first and the other with type first, the analyzer prefers the one with type first. When I delete the type first index, it is still able to use the id first index with similar query statistics.

I'm not sure that's an entirely accurate assessment.  Or rather, statistically, if they are the same, it will just pick one, but that doesn't then imply that type,id is better than id,type.

Based on all my readings, higher cardinality should always be first because it eliminates more rows and allows the index to execute faster, because the row elimination happens on the first pass of the index.  So id,type is nearly always preferable over type,id. The only time this shouldn't be true is if we have queries that don't provider the _id in the query at all.

> Having `name, type, id` ...

Not my suggestion - I think it should be `id, type, name`

PR #706, opened 2023-09-21
Remove VimStrings from MiqQueue

kbrock:
Events used to have vmware specific objects in the queue. These leaked through for a customer and it blew up.

This has not been in the queue for a number of years, but if someone delays upgrading and then doesn't empty the queue before upgrading... this will happen.

The solution is to remove entries

See also
- #671
- #670

### DIFF
--- a/db/migrate/20200427122455_remove_vim_strings_from_notifications.rb
+++ b/db/migrate/20200427122455_remove_vim_strings_from_notifications.rb
@@ -7,30 +7,10 @@ class RemoveVimStringsFromNotifications < ActiveRecord::Migration[5.1]
   end
 
   def up
-    say_with_time("Removing VimStrings from Notifications") do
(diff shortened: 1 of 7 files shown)


kbrock:
update:
- fixed rubocops
- fixed method rename that wasn't fully done.
- refactored other migrations to use this common function

Punted:
- `VimString` replace that followed a different pattern [ems_ref_obj](https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20191023180330_migrate_ems_ref_obj_to_ems_ref_type.rb) - think is not applicable.
- `VimString` replace of a binary column. [binary_blobs](https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20220114155819_remove_vim_types_from_binary_blob_parts.rb) follows the same pattern, but would add 25% more code from conditional logic. your call

agrare (approved):

PR #707, opened 2023-10-12
Support ruby 3.1

jrafanie:
* Update stubs to work with manageiq-password 1.2.0+
* For tests, use YAML.unsafe load in psych 4+

Part of https://github.com/ManageIQ/manageiq/issues/22696


jrafanie:
hmm, the bot comment didn't come back.  😦

PR #708, opened 2023-10-23
Add new resource_pools column for cpu_cores_*

jaywcarman:
Proposing this instead of #675 

Fixes https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/108

Related to:
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/474
- https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/147

The resource_pools table schema follows VMware resource pools. Because of this the 'cpu_reserve', 'cpu_limit', and 'cpu_shares' columns are all integers (representing units of Hz).

When IbmCloud::PowerVirtualServers and IbmPowerHmc providers used resource_pools to store information about Shared Processor Pools they stored CPU core shares in these 'cpu_*' columns (even though they are not in Hz). Since there is a unit mis-match the best way to store both in the same table is to create new columns 'cpu_core_*' (note the data type is float since fractional core counts are allowed).

### DIFF
--- a/db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
+++ b/db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
@@ -1,0 +1,34 @@
+class AddCpuCoresColumnsToResourcePools < ActiveRecord::Migration[6.0]
+  class ResourcePool < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :resource_pools, :cpu_cores_available, :float
+    add_column :resource_pools, :cpu_cores_reserve, :float
+    add_column :resource_pools, :cpu_cores_limit, :float
+
+    say_with_time('Removing non-applicable data from IbmCloud::PowerVirtualServer resource pools') do
+      ResourcePool.in_my_region.where(:type => %w[ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ResourcePool]).update_all(
+        :cpu_shares  => nil,
+        :cpu_reserve => nil,
+        :cpu_limit   => nil
+      )
+    end
+
+    say_with_time('Removing non-applicable data from IbmPowerHmc resource pools') do
+      ResourcePool.in_my_region.where(:type => %w[ManageIQ::Providers::IbmPowerHmc::InfraManager::ProcessorResourcePool]).update_all(
+        :cpu_shares  => nil,
+        :cpu_reserve => nil,
+        :cpu_limit   => nil
+      )
+    end
+  end
+
+  def down
+    remove_column :resource_pools, :cpu_cores_available
+    remove_column :resource_pools, :cpu_cores_reserve
+    remove_column :resource_pools, :cpu_cores_limit
+  end
+end
--- a/spec/migrations/20231018141450_add_cpu_cores_columns_to_resource_pools_spec.rb
+++ b/spec/migrations/20231018141450_add_cpu_cores_columns_to_resource_pools_spec.rb
@@ -1,0 +1,91 @@
+require_migration
+
+describe AddCpuCoresColumnsToResourcePools do
+  let(:resource_pools_stub) { migration_stub(:ResourcePool) }
+
+  migration_context :up do
+    it "Removes non-applicable data only from Power resource pools" do
+      ibm_cloud_power_vs_resource_pool = resource_pools_stub.create(
+        :type               => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ResourcePool",
+        :cpu_shares         => 1,
+        :cpu_reserve        => 2,
+        :cpu_limit          => 3,
+        :cpu_reserve_expand => true,
+        :is_default         => false
+      )
+
+      ibm_power_hmc_processor_resource_pool = resource_pools_stub.create(
+        :type               => "ManageIQ::Providers::IbmPowerHmc::InfraManager::ProcessorResourcePool",
+        :cpu_shares         => 1,
+        :cpu_reserve        => 2,
+        :cpu_limit          => -1,
+        :cpu_reserve_expand => true,
+        :is_default         => false
+      )
+
+      migrate
+
+      expect(ibm_cloud_power_vs_resource_pool.reload).to have_attributes(
+        :cpu_shares         => nil,
+        :cpu_reserve        => nil,
+        :cpu_limit          => nil,
+        :cpu_reserve_expand => true,
+        :is_default         => false
+      )
+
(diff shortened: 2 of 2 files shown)


spec/migrations/20231018141450_add_cpu_cores_columns_to_resource_pools_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        :type               => "ManageIQ::Providers::Vmware::InfraManager::ResourcePool",
+        :cpu_shares         => 1,
+        :cpu_reserve        => 2,
+        :cpu_limit          => 3,  # rand(-1..65536) ?
+        :cpu_reserve_expand => true,
+        :is_default         => false
+      )
+    end
+  end
+  migration_context :up do
```
Fryguy, line 67: If these are also up tests, then no need for a second migration_context :up

```suggestion

```

jaywcarman: Oops, I meant for this to be `migration_context :down do` instead of a duplicate up. Since down is removing the new columns, I only added a test to ensure any existing vmware resoure_pools are not affected.

Will force push to fix this.

db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
```diff
@@ -0,0 +1,34 @@
...
+    add_column :resource_pools, :cpu_cores_reserve, :float
+    add_column :resource_pools, :cpu_cores_limit, :float
+
+    say_with_time('Removing data non-applicable data from IbmCloud::PowerVirtualServer resource pools') do
+      ResourcePools.in_my_region.where(:type => %w[ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ResourcePool]).update_all(
+        :cpu_shares  => nil,
+        :cpu_reserve => nil,
+        :cpu_limit   => nil
+      )
+    end
```
kbrock, line 18: I think newly added columns will get a default of `null`.
So I think this sql is not doing anything.

kbrock (approved):

kbrock:
Are there changes to the refresh classes so these share columns will get null going forward?

kbrock (approved):

db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
```diff
@@ -0,0 +1,34 @@
+class AddCpuCoresColumnsToResourcePools < ActiveRecord::Migration[6.0]
+  class ResourcePools < ActiveRecord::Base
```
agrare, line 2: The class should be `ResourcePool` (surprised this work without a `self.table_name`)
https://github.com/ManageIQ/manageiq/blob/master/app/models/resource_pool.rb#L1

db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
```diff
@@ -0,0 +1,34 @@
...
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :resource_pools, :cpu_cores_available, :float
+    add_column :resource_pools, :cpu_cores_reserve, :float
+    add_column :resource_pools, :cpu_cores_limit, :float
+
+    say_with_time('Removing data non-applicable data from IbmCloud::PowerVirtualServer resource pools') do
```
agrare, line 12:
```suggestion
    say_with_time('Removing non-applicable data from IbmCloud::PowerVirtualServer resource pools') do
```

agrare:
@jaywcarman I fixed the description but can you update the commit message to fix the typos? (`resoure pools.` and `informatino`)

db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
```diff
@@ -0,0 +1,34 @@
...
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    add_column :resource_pools, :cpu_cores_available, :float
+    add_column :resource_pools, :cpu_cores_reserve, :float
+    add_column :resource_pools, :cpu_cores_limit, :float
+
+    say_with_time('Removing non-applicable data from IbmCloud::PowerVirtualServer resource pools') do
+      ResourcePools.in_my_region.where(:type => %w[ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ResourcePool]).update_all(
```
agrare, line 13:
```suggestion
      ResourcePool.in_my_region.where(:type => %w[ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ResourcePool]).update_all(
```

db/migrate/20231018141450_add_cpu_cores_columns_to_resource_pools.rb
```diff
@@ -0,0 +1,34 @@
...
+        :cpu_shares  => nil,
+        :cpu_reserve => nil,
+        :cpu_limit   => nil
+      )
+    end
+
+    say_with_time('Removing non-applicable data from IbmPowerHmc resource pools') do
+      ResourcePools.in_my_region.where(:type => %w[ManageIQ::Providers::IbmPowerHmc::InfraManager::ProcessorResourcePool]).update_all(
```
agrare, line 21:
```suggestion
      ResourcePool.in_my_region.where(:type => %w[ManageIQ::Providers::IbmPowerHmc::InfraManager::ProcessorResourcePool]).update_all(
```

spec/migrations/20231018141450_add_cpu_cores_columns_to_resource_pools_spec.rb
```diff
@@ -0,0 +1,91 @@
+require_migration
+
+describe AddCpuCoresColumnsToResourcePools do
+  let(:resource_pools_stub) { migration_stub(:ResourcePools) }
```
agrare, line 4:
```suggestion
  let(:resource_pools_stub) { migration_stub(:ResourcePool) }
```

agrare:
@jaywcarman I think we discussed this on a call earlier but the cpu_(shares,limit,reserve) are in MHz not in `#cpus`.  This could work with fractional cores (as long as it is divisible by a MHz) if you know the pool's CPU speed, is that something that is possible to get on ibm_cloud power / powerhmc?

jaywcarman:
> if you know the pool's CPU speed, is that something that is possible to get on ibm_cloud power / powerhmc?

Host CPU clock speed is not available through PowerVS. For HMC I don't see it in available in our current SDK schemas but it seems like something we might be able to get. Even if we can collect clock speed, it would be messy because the Power hosts are able to run in modes with variable clock speeds (based on workload and cores available).

agrare (approved):

PR #710, opened 2023-11-08
Add payload validity attributes to configuration_scripts

agrare:
Add columns to store if the payload of a configuration_script is valid or not, and if not what the error is to help the user fix their scripts.

### DIFF
--- a/db/migrate/20231108172936_add_valid_to_configuration_scripts.rb
+++ b/db/migrate/20231108172936_add_valid_to_configuration_scripts.rb
@@ -1,6 +1,6 @@
 class AddValidToConfigurationScripts < ActiveRecord::Migration[6.1]
   def change
-    add_column :configuration_scripts, :payload_valid, :boolean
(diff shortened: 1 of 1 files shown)


db/migrate/20231108172936_add_valid_to_configuration_scripts.rb
```diff
@@ -0,0 +1,6 @@
+class AddValidToConfigurationScripts < ActiveRecord::Migration[6.1]
+  def change
+    add_column :configuration_scripts, :payload_valid, :boolean
```
Fryguy, line 3: Should we set a default of true (or in Ruby use default_value_for)?  Reason I'm asking is that ansible playbooks will be seen as "nil" valid, which is falsey in Ruby, so `.payload_valid?` will return false.

agrare: Done

Fryguy:
Moving this to yes? until the parent PR is ready.

PR #712, opened 2023-11-14
Remove default from config_scripts payload_valid

agrare:
Only set payload_valid if it was checked otherwise leave it as `nil`

https://github.com/ManageIQ/manageiq-providers-workflows/pull/59#issuecomment-1810299583


agrare:
ping @Fryguy

kbrock:
ping. Is this ready?

kbrock:
@Fryguy Is there a timeframe that is best to merge these schema PRs?

Fryguy:
This one fell to the backburner, but I originally had concerns/thoughts around how to handle providers that don't support this column, and thus the value would be nil, and what that means (i.e. since nil is falsey, are they therefore invalid by default?)

Fryguy:
Closing in favor of the approach in #712

PR #714, opened 2024-01-09
Fix OrchestrationStack#configuration_script reference

agrare:
There are some `OrchestrationStack` types (AWX and AnsibleTower Jobs) which were using `orchestration_template_id` to reference a configuration_script.  This was causing issues with the other side of the relationship, namely if there was an `OrchestrationTemplate` with the same ID as the `ConfigurationScript` then that template would think it had active stacks and could not be deleted.

There is already a `OrchestrationStack#configuration_script_base_id` but that is used to track the `Playbook` aka the `ConfigurationScriptPayload` where this is used to track the `JobTemplate` which is a running `ConfigurationScript`.  Ideally I would have named these `configuration_script_payload_id` and `configuration_script_id` so it was obvious that the existing one is for the payload but I didn't want to get into renaming columns/indexes/etc... right now.

Dependents:
* https://github.com/ManageIQ/manageiq-providers-awx/pull/22
* https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/302

https://github.com/ManageIQ/manageiq/issues/22813

### DIFF
--- a/db/migrate/20240108211241_fix_orchestration_stack_configuration_script_reference.rb
+++ b/db/migrate/20240108211241_fix_orchestration_stack_configuration_script_reference.rb
@@ -1,0 +1,22 @@
+class FixOrchestrationStackConfigurationScriptReference < ActiveRecord::Migration[6.1]
+  class OrchestrationStack < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  TYPES = %w[ManageIQ::Providers::Awx::AutomationManager::Job ManageIQ::Providers::AnsibleTower::AutomationManager::Job].freeze
+
+  def up
+    say_with_time("Fixing OrchestrationStack#configuration_script foreign key") do
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all('configuration_script_id = orchestration_template_id')
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all(:orchestration_template_id => nil)
+    end
+  end
+
+  def down
+    say_with_time("Resetting OrchestrationStack#configuration_script foreign key") do
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all('orchestration_template_id = configuration_script_id')
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all(:configuration_script_id => nil)
+    end
+  end
+end
--- a/db/migrate/20240105173129_add_orchestration_stack_configuration_script.rb
+++ b/db/migrate/20240105173129_add_orchestration_stack_configuration_script.rb
@@ -1,0 +1,5 @@
+class AddOrchestrationStackConfigurationScript < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :orchestration_stacks, :configuration_script, :type => :bigint, :index => true
+  end
+end
--- a/spec/migrations/20240108211241_fix_orchestration_stack_configuration_script_reference_spec.rb
+++ b/spec/migrations/20240108211241_fix_orchestration_stack_configuration_script_reference_spec.rb
@@ -1,0 +1,84 @@
+require_migration
+
+class FixOrchestrationStackConfigurationScriptReference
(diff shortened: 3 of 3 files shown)


agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/301, https://github.com/ManageIQ/manageiq-providers-awx/pull/22

agrare:
@miq-bot cross-repo-tests manageiq, manageiq-api, manageiq-ui-classic, https://github.com/ManageIQ/manageiq-providers-awx/pull/22, https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/302

agrare:
@jrafanie kbrock (won't ping him right now since he's on vacation and I _know_ he'll actually respond :laughing: ) I know neither of you are very familiar with ansible_tower but wonder if you could review this?

High level an Ansible Tower Job (OrchestrationStack) has a reference to a playbook (ConfigurationScriptPayload) via configuration_script_base_id and to a job_template (ConfigurationScript) via orchestration_template_id.  That last one is a problem because if there is an OrchestrationTemplate with the same ID as the ConfigurationScript/JobTemplate then it thinks it has active stacks when it doesn't.

db/migrate/20240108211241_fix_orchestration_stack_configuration_script_reference.rb
```diff
@@ -0,0 +1,22 @@
...
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+  end
+
+  TYPES = %w[ManageIQ::Providers::Awx::AutomationManager::Job ManageIQ::Providers::AnsibleTower::AutomationManager::Job].freeze
+
+  def up
+    say_with_time("Fixing OrchestrationStack#configuration_script foreign key") do
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all('configuration_script_id = orchestration_template_id')
+      OrchestrationStack.in_my_region.where(:type => TYPES).update_all(:orchestration_template_id => nil)
```
Fryguy, line 12: I think think can be done in a single line:

```suggestion
      OrchestrationStack.in_my_region.where(:type => TYPES).update_all(
        :configuration_script_id   => :orchestration_template_id,
        :orchestration_template_id => nil
      )
```

agrare: Pretty sure I tried that and it didn't work...let me double check

agrare: Yeah no, to set one column to the value of another you need the `'columnA = columnB'` syntax, `update_all(:configuration_script_id   => :orchestration_template_id)` just sets `:configuration_script_id=0` in this case

agrare: Plus I figured they're in the same transaction so the performance difference will be negligible

Fryguy: oh wild, I was sure that worked 🤯

Fryguy (approved):

db/migrate/20240108211241_fix_orchestration_stack_configuration_script_reference.rb
agrare: Yeah I'm assuming AR can't tell the difference between you wanting to set it to the column `:orchestration_template_id` or the symbol value `:orchestration_template_id`

Fryguy:
Backported to `quinteros` in commit f5adf88df27144a90391fb34d3652d5553a370bc.

```text
commit f5adf88df27144a90391fb34d3652d5553a370bc
Date:   Thu Jan 11 15:00:18 2024 -0500

    Merge pull request #714 from agrare/fix_orchestration_stack_configuration_script_reference
    
    Fix OrchestrationStack#configuration_script reference
    
    (cherry picked from commit e63f02a662eabf1b6e34e72fb15671796268873f)
```

PR #715, opened 2024-01-12
Allow rails 7 gems in gemspec

jrafanie:
Allow rails 7 gems in gemspec

Depends on:

- [x] https://github.com/ManageIQ/activerecord-id_regions/pull/33
- [x] New id_regions release (0.4.0)
- [x] Extracted backward compatible changes: https://github.com/ManageIQ/manageiq-schema/pull/735
- [x] Fix failures creating associations in migration specs due to ActiveRecord::AssociationTypeMismatch errors.  Rails 7 only. (associations classes not matching stub models)
- [x] Fix "secrets" related test failures with rails 7 (mock/stubs on stub models)
- [x] Cleanup ENV stubbing 20180606155924_move_ansible_container_secrets_into_database (workaround inability to mock/stub stub models).

The commits should be reviewable one by one.

Rails 7 changed the way migrations are run.  The migration class
constant is now removed and the file is reloaded between migrations.

This means tests should not use expect/allow on migration stub classes
as the constant stubbed is not the same constant the migration is run against.

Additionally, comparisons of objects created before the migration with ones
after will also not work.  In that case, you can assert by id or other
attributes instead of comparing objects.

Rails also checks association classes are correct and will also not match
expectations.  For example, instead of using host to create or query
the host association, use host_id instead.  This also works with has_manys
such as miq_product_feature_ids.

See https://github.com/rails/rails/pull/42198

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,10 +13,12 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
-case ENV['TEST_RAILS_VERSION']
-when "6.0"
-  # Default local bundling to use 6.0 for generating migrations
-  gem "rails",  "~>6.0.4"
-else
-  gem "rails",  "~>6.1.6"
-end
+minimum_version =
+  case ENV.fetch('TEST_RAILS_VERSION', nil)
+  when "7.0"
+    "~>7.0.8"
+  else
+    # Default local bundling to use this version for generating migrations
+    "~>6.1.4"
+  end
+gem "rails", minimum_version
--- a/db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
+++ b/db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
@@ -14,14 +14,26 @@ class MoveAnsibleContainerSecretsIntoDatabase < ActiveRecord::Migration[5.0]
     update_authentications
   end
 
-  def self.read_token(file)
-    File.read(file)
-  end
-
   private
 
+  def token_file
+    if Rails.env.test? && ENV['TOKEN_FILE']
+      ENV['TOKEN_FILE']
+    else
+      TOKEN_FILE
+    end
+  end
+
+  def ca_cert_file
+    if Rails.env.test? && ENV['CA_CERT_FILE']
+      ENV['CA_CERT_FILE']
+    else
+      CA_CERT_FILE
+    end
+  end
+
   def containerized?
-    File.exist?(TOKEN_FILE) && File.exist?(CA_CERT_FILE)
+    File.exist?(token_file) && File.exist?(ca_cert_file)
   end
 
   def update_authentications
@@ -92,8 +104,8 @@ class MoveAnsibleContainerSecretsIntoDatabase < ActiveRecord::Migration[5.0]
   def request_params
     {
       'Accept'         => "application/json",
-      'Authorization'  => "Bearer #{self.class.read_token(TOKEN_FILE)}",
-      :ssl_ca_cert     => CA_CERT_FILE,
+      'Authorization'  => "Bearer #{File.read(token_file)}",
+      :ssl_ca_cert     => ca_cert_file,
       :ssl_verify_mode => OpenSSL::SSL::VERIFY_PEER
     }
   end
--- a/spec/migrations/20180606155924_move_ansible_container_secrets_into_database_spec.rb
+++ b/spec/migrations/20180606155924_move_ansible_container_secrets_into_database_spec.rb
@@ -59,13 +59,19 @@ describe MoveAnsibleContainerSecretsIntoDatabase do
       end
 
       before do
-        stub_const("MoveAnsibleContainerSecretsIntoDatabase::TOKEN_FILE", token_path)
-        stub_const("MoveAnsibleContainerSecretsIntoDatabase::CA_CERT_FILE", cert_path)
         expect(URI::HTTPS).to receive(:build).with({
           :host => kube_host,
           :port => kube_port,
           :path => "/api/v1/namespaces/#{namespace}/secrets/ansible-secrets"
         }).and_return(uri_stub)
+
+        ENV["CA_CERT_FILE"] = cert_path
+        ENV["TOKEN_FILE"] = token_path
+      end
+
+      after do
+        ENV.delete("CA_CERT_FILE")
+        ENV.delete("TOKEN_FILE")
       end
 
       it "doesn't add any authentications if the secret is not found" do
@@ -111,7 +117,9 @@ describe MoveAnsibleContainerSecretsIntoDatabase do
   end
 
   def expect_request
-    expect(described_class).to receive(:read_token).with(token_path).and_return("totally-a-token")
+    token.write("totally-a-token")
+    token.close
+
     response = double("RequestIO", :read => secret_json)
     expect(uri_stub).to receive(:open).with({
       'Accept'         => "application/json",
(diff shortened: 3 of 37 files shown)


jrafanie:
Oh, these are fun errors... looks like our `migration_stub` method is returning a cached constant and any associations you define in your migration "stub" fail under rails 7 because it's not a refresh association class.

kbrock:
sigh. ping if you need to pair on this.
no ideas at this time

jrafanie:
Ok, all green, now to clean up the ENV stubbing migration `20180606155924_move_ansible_container_secrets_into_database`

Gemfile
```diff
@@ -13,10 +13,12 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
...
-else
-  gem "rails",  "~>6.1.6"
-end
+minimum_version =
+  case ENV['TEST_RAILS_VERSION']
+  when "7.0"
+    "~>7.0.8"
+  else
+    # Default local bundling to use this version for generating migrations
+    "~>6.1.4"
```
jrafanie, line 22: We're still defaulting to rails 6.1 for new migrations.

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
```diff
@@ -14,14 +14,26 @@ def up
...
+      TOKEN_FILE
+    end
   end
 
-  private
+  def ca_cert_file
+    if Rails.env.test? && ENV['CA_CERT_FILE']
+      ENV['CA_CERT_FILE']
+    else
+      CA_CERT_FILE
```
jrafanie, line 31: These methods were added to inject different behavior in test vs production.  See b5cb2e7a857e2886636f3fd0142cf642b2084e5d

spec/migrations/20180606155924_move_ansible_container_secrets_into_database_spec.rb
```diff
@@ -111,7 +117,9 @@
   end
 
   def expect_request
-    expect(described_class).to receive(:read_token).with(token_path).and_return("totally-a-token")
+    token.write("totally-a-token")
+    token.close
```
jrafanie, line 121: Avoid the expectation on the migration stub constant since it's been removed and re-loaded by the migration.

spec/migrations/20180712122000_remove_host_provisioning_spec.rb
```diff
@@ -7,9 +7,9 @@
     let(:miq_approval_stub) { migration_stub :MiqApproval }
 
     it 'only removes Host Provision request instances' do
-      miq_request_stub.create!(:type => 'MiqProvisionRequest', :miq_approvals => [miq_approval_stub.create!])
-      miq_request_stub.create!(:type => 'MiqHostProvisionRequest', :miq_approvals => [miq_approval_stub.create!])
-      miq_request_stub.create!(:type => 'ServiceTemplateProvisionRequest', :miq_approvals => [miq_approval_stub.create!])
+      miq_request_stub.create!(:type => 'MiqProvisionRequest', :miq_approval_ids => [miq_approval_stub.create!.id])
+      miq_request_stub.create!(:type => 'MiqHostProvisionRequest', :miq_approval_ids => [miq_approval_stub.create!.id])
+      miq_request_stub.create!(:type => 'ServiceTemplateProvisionRequest', :miq_approval_ids => [miq_approval_stub.create!.id])
```
jrafanie, line 12: avoid calling the association methods because rails does a class lookup check which will fail with rails 7 since our stub class constant has been removed by rails 7 by the time this is run and so it doesn't match the expected association class object.

spec/migrations/20190617191109_move_awx_credentials_to_authentications_spec.rb
```diff
@@ -97,7 +97,8 @@
         :type => "ManageIQ::Providers::EmbeddedAnsible::AutomationManager::VaultCredential"
       )
       expect(auth.manager_ref).to be_nil
-      expect(described_class).to_not receive(:pg_connection)
+      allow(PG::Connection).to receive(:new).and_call_original
+      expect(PG::Connection).to_not receive(:new).with(a_hash_including(:dbname => "awx"))
```
jrafanie, line 101: Again, avoid expectations on the migration class constant as it's been removed and re-loaded by the time these expectations are reached.

jrafanie:
Ok, all green and cleaned up.  The commits should be reviewable one by one.

Fryguy:
Can you explain why we need the `id` changes?  There are a gazillion of them.  Are the relationships lost or something in the cached stubs?

jrafanie:
> Can you explain why we need the `id` changes? There are a gazillion of them. Are the relationships lost or something in the cached stubs?

Sure, I had it in each commit.  I put it it in the description:

Rails 7 changed the way migrations are run.  The migration class
constant is now removed and the file is reloaded between migrations.

This means tests should not use expect/allow on migration stub classes
as the constant stubbed is not the same constant the migration is run against.

Additionally, comparisons of objects created before the migration with ones
after will also not work.  In that case, you can assert by id or other
attributes instead of comparing objects.

Rails also checks association classes are correct and will also not match
expectations.  For example, instead of using host to create or query
the host association, use host_id instead.  This also works with has_manys
such as miq_product_feature_ids.

See https://github.com/rails/rails/pull/42198

kbrock:
@jrafanie so you are converting from relations to ids so we can avoid defining relations all together? (e.g.: `ext_management_system => ems` to `ems_id => ems.id`)

Seems those are very non-rails specific and can be pulled out of here?

jrafanie:
> @jrafanie so you are converting from relations to ids so we can avoid defining relations all together? (e.g.: `ext_management_system => ems` to `ems_id => ems.id`)

Not specifically.  I think that's beyond the scope here.  The problem is that the association itself has logic to verify the class for the association.  This verification doesn't work if you're trying to create an association with an object that is not the expected constant.   This worked in 6.1 because the migration was loaded only once and the stub model constant remained.  With rails 7, the migration stub model constants are removed and the migration is loaded again, so the classes aren't equivalent.  What I'm doing here is avoiding the association method when I'm creating or comparing objects.

The migrations that define the associations can still use those association methods but I'm trying to avoid them in the tests for the issues found with rail 7.

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
kbrock: can we change the definition of CA_CERT_FILE instead?

```
CA_CERT_FILE = ENV[] || "const".freeze
```

spec/migrations/20180606155924_move_ansible_container_secrets_into_database_spec.rb
kbrock: I also like avoiding `expect` all together.

👍

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
kbrock: no. ignore

can we just do:

```ruby
def ca_cert_file
  ENV.fetch("CA_CERT_FILE", "/run/secrets/kubernetes.io/serviceaccount/ca.crt")
end
```

or maybe `ENV[].presence || ""`

jrafanie: > and do these need to be test only?

Yes, my objective was to not change the migration to add test code into it.  I think what is here is a fair enough workaround to avoid test code in migrations while still being able to work with rails 7 neutering the ability to do `allow`/`expect` on the stub migration models.

jrafanie: > ```ruby
> def ca_cert_file
>   ENV.fetch("CA_CERT_FILE", "/run/secrets/kubernetes.io/serviceaccount/ca.crt")
> end
> ```

I was trying not to change the behavior of production migrations.  While the proposed code reads, I really didn't want to change the behavior of past migrations.

PR #717, opened 2024-02-14
Remove Openstack NetworkManager metrics worker records

agrare:
This worker class was removed in https://github.com/ManageIQ/manageiq-providers-openstack/pull/548, we should delete any leftover worker records.

Fixes https://github.com/ManageIQ/manageiq/issues/22891

### DIFF
--- a/db/migrate/20240214165402_remove_openstack_network_manager_metrics_collector_workers.rb
+++ b/db/migrate/20240214165402_remove_openstack_network_manager_metrics_collector_workers.rb
@@ -1,0 +1,11 @@
+class RemoveOpenstackNetworkManagerMetricsCollectorWorkers < ActiveRecord::Migration[6.1]
+  class MiqWorker < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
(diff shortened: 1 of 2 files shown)


db/migrate/20240214165402_remove_openstack_network_manager_metrics_collector_workers.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveOpenstackNetworkManagerMetricsCollectorWorkers < ActiveRecord::Migration[6.1]
+  class MiqWorker < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    MiqWorker.where(:type => "ManageIQ::Providers::Openstack::NetworkManager::MetricsCollectorWorker").delete_all
```
Fryguy, line 7: Can you add `say_with_time` ?

agrare: :+1: Done

Fryguy (approved):

PR #721, opened 2024-02-29
[WIP] Add simple column defaults... still need to verify none had prior db …

jrafanie:
…defaults

### DIFF
--- a/db/migrate/20240228190145_add_column_defaults.rb
+++ b/db/migrate/20240228190145_add_column_defaults.rb
@@ -1,0 +1,57 @@
+class AddColumnDefaults < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default :assigned_server_roles, :active,                   :from => nil, :to => false
+    change_column_default :blacklisted_events, :enabled,                     :from => nil, :to => true
+    change_column_default :classifications, :read_only,                      :from => nil, :to => false
+    change_column_default :classifications, :show,                           :from => nil, :to => true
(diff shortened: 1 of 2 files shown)


kbrock:
1. Think `allow_null => false` is good here.
2. The allow null updates all the columns to the default value.
3. But since the update is part of a schema change, it blocks the whole database. So people tend to manually update the database and then do the schema change (i.e.: disallow null) in a separate transaction. This makes the change very verbose. Maybe we want a helper?

```
{
  :table1 => {
    :column1 => default,
    :column2 => default
  }
}



def migrate_null_defaults(table_name, columns)
  # can we do a single query?
  columns.each do |column_name, default|
    change_column_default(table_name, column_name, default)
  end
  model.where(columns.transform_values {|n, v| v = nil}).update_all(columns)

  # can we do a single query?
  # verify this is that much different
  columns.each do |column_name, default|
    change_column_null(table_name, column_name, false, default)
  end
end
```

kbrock:
I see you closed https://github.com/ManageIQ/manageiq/pull/22923

Do we still want to go forward with this?
Personally I really like having these defaults in the database, thought I know we've been mixed about changing the db schema.

jrafanie:
> Personally I really like having these defaults in the database, thought I know we've been mixed about changing the db schema.

yeah, that's why I didn't close this one 😉

jrafanie:
I don't know when I'll get back to this.  Closing for now.

PR #733, opened 2024-05-01
Bump linux_admin to 3.0 for net-ssh fix

agrare:
Required for: https://github.com/ManageIQ/manageiq/pull/23014

### DIFF
--- a/manageiq-schema.gemspec
+++ b/manageiq-schema.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "ancestry"
   spec.add_dependency "activerecord-id_regions", "~> 0.3.2"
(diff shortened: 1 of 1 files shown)


manageiq-schema.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "ancestry"
   spec.add_dependency "activerecord-id_regions", "~> 0.3.2"
-  spec.add_dependency "linux_admin",             "~> 2.0"
+  spec.add_dependency "linux_admin",             "~> 3.0"
```
bdunne, line 23: Should we loosen this dependency?
```suggestion
  spec.add_dependency "linux_admin",             ">= 2.0", "< 4"
```

agrare: Done

PR #735, opened 2024-05-22
Backward compatible rails 7 changes

jrafanie:
Extracted from #715

These are backward compatible changes to fix failures in rails 7.  Interestingly, detected the typo preventing the down migration from running and also the weird migration stub issue where the it's defined in the test itself.


jrafanie:
cc @kbrock

db/migrate/20220111170303_fix_child_container_manager_sti.rb
```diff
@@ -46,18 +46,17 @@ def up
         ServiceParametersSet.in_my_region.where(:ext_management_system => managers).update_all(:type => "#{provider_klass}::ServiceParametersSet")
       end
     end
+  end
```
jrafanie, line 49: oops!  I can't believe 6.1 didn't fail previously with this

spec/migrations/20190201173316_add_missing_ems_id_to_switch_spec.rb
```diff
@@ -1,35 +1,5 @@
...
-    self.inheritance_column = :_type_disabled
-
-    belongs_to :ext_management_system, :foreign_key => "ems_id", :class_name => "AddMissingEmsIdToSwitch::ExtManagementSystem"
-    belongs_to :host, :class_name => "AddMissingEmsIdToSwitch::Host"
-
-    has_many :hosts, :through => :host_switches, :class_name => "AddMissingEmsIdToSwitch::Host"
-    has_many :host_switches, :class_name => "AddMissingEmsIdToSwitch::HostSwitch"
-  end
-end
-
```
jrafanie, line 32: There is no reason to re-open the migration class to define stubs... this was actually causing issues in the tests in rails 7 because it couldn't find the `has_many :host_switches, :class_name => "AddMissingEmsIdToSwitch::HostSwitch"` class `AddMissingEmsIdToSwitch::HostSwitch`.  It makes sense to move this to the migration class, even if the migration isn't using the stubs.

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
```diff
@@ -54,7 +54,7 @@ def update_authentications
     database_auth_find_args = db_args.merge(
       :name     => "Ansible Database Authentication",
       :authtype => "ansible_database_password",
-      :userid   => ApplicationRecord.configurations[Rails.env]["username"],
+      :userid   => ApplicationRecord.configurations.configs_for(:env_name => Rails.env).first.configuration_hash["username"],
```
Fryguy, line 57: wow the new way is gross - wonder if we should move it into a spec helper method.

db/migrate/20190201173316_add_missing_ems_id_to_switch.rb
```diff
@@ -1,4 +1,32 @@
...
+
+  class Switch < ActiveRecord::Base
+    self.inheritance_column = :_type_disabled
+
+    belongs_to :ext_management_system, :foreign_key => "ems_id", :class_name => "AddMissingEmsIdToSwitch::ExtManagementSystem"
+    belongs_to :host, :class_name => "AddMissingEmsIdToSwitch::Host"
+
+    has_many :hosts, :through => :host_switches, :class_name => "AddMissingEmsIdToSwitch::Host"
+    has_many :host_switches, :class_name => "AddMissingEmsIdToSwitch::HostSwitch"
+  end
```
Fryguy, line 28: Why is any of this necessary?  It's not used in the script itself.

db/migrate/20220111170303_fix_child_container_manager_sti.rb
Fryguy: wow

spec/migrations/20190201173316_add_missing_ems_id_to_switch_spec.rb
Fryguy: The reason these are in the test only is because the migration itself doesn't use those classes.  At "runtime" we don't want to have to create stubs we don't need, because it can only introduce potential problems.

That being said, the way it was doing it before by reopening the migration is also wrong.  In other migrations where we need a "new" migration stub, we have a helper method called `new_migration_stub`.  Granted, it only creates a basic stub class, but it's possible we could pass a block to it to allow further manipulation.

Fryguy:
```
$ git grep "new_migration_stub"
spec/migrations/20200221200855_add_ancestry_to_miq_ae_namespace_spec.rb:4:  let(:ns_stub) { new_migration_stub(:miq_ae_namespaces) }
spec/migrations/20210519204247_move_miq_set_memberships_spec.rb:7:  let(:set_stub)        { new_migration_stub(:miq_sets) }
spec/migrations/20210519204247_move_miq_set_memberships_spec.rb:8:  let(:item_stub)       { new_migration_stub(:scan_items) }
spec/migrations/20210519204247_move_miq_set_memberships_spec.rb:9:  let(:rp_stub)         { new_migration_stub(:resource_pools) }
spec/migrations/20210824231354_add_timestamps_to_binary_blobs_spec.rb:4:  let(:binary_blob_stub) { new_migration_stub(:binary_blobs) }
spec/support/migration_helper.rb:62:      def new_migration_stub(table_name)
```

db/migrate/20190201173316_add_missing_ems_id_to_switch.rb
Fryguy: Oh I see - See my other comment.

spec/migrations/20190201173316_add_missing_ems_id_to_switch_spec.rb
jrafanie: yeah, we can possibly add the block case.  I don't see the potential problem of defining a migration stub within the migration class if it's only used in the test.  It feels wrong to be in the test itself unless it's a basic `new_migration_stub`.

jrafanie: To be fair, our reuse of migration stubs through multiple tests are problematic under rails 7 so we may have to revisit this anyway.

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
jrafanie: I'll check to see if there's a less gross way to do this.  Although, I believe we had changes like this for appliance console and other areas already.

spec/migrations/20190201173316_add_missing_ems_id_to_switch_spec.rb
jrafanie: I'll drop this change from this PR.  I don't have a solution that I like for now.

Fryguy: Oh wow - ok yeah - we'll have to discuss - sounds like we need to "reimplement" some of the stub handling for Rails 7 - what a pain.

Fryguy:
Does this PR upgrade the dummy app to Rails 7?  Asking because it's changing some of the files within the dummy app.

db/migrate/20180606155924_move_ansible_container_secrets_into_database.rb
jrafanie: a little better but still the way rails wants it:

```ruby
ApplicationRecord.configurations.configs_for(:env_name => Rails.env, :name => "primary").configuration_hash["username"]
```

jrafanie:
> Does this PR upgrade the dummy app to Rails 7? Asking because it's changing some of the files within the dummy app.

No, I'm just dropping some sprockets related files that aren't needed in the dummy app for testing migrations... those sprockets files assume sprocket-rails is required by rails... which is not the case with rails 7 so dropping the needless files is one less thing to deal with for rails 7.

kbrock (approved):
LGTM

Thought these were required for 6.1

Anything special outstanding?
This looks merge ready to me

jrafanie:
> Thought these were required for 6.1

I think they were deprecated in 6.1 and dropped on 7.0.  I guess we never saw the deprecations in 6.1.

PR #736, opened 2024-05-29
Remove code for unsupported rails versions

jrafanie:
We already dropped support for rails 6 and lower so these conditionals are not needed.

Discovered when working on rails 7 support in:
https://github.com/ManageIQ/manageiq-schema/pull/715

NOTE1: Hiding whitespace helps. 
NOTE2: This is completely standalone from #715.

### DIFF
--- a/spec/support/migration_helper.rb
+++ b/spec/support/migration_helper.rb
@@ -107,15 +107,9 @@ module Spec
 
       def migrate_to(version)
         suppress_migration_messages do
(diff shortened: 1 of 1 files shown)


jrafanie:
This is pretty straight forward.  We haven't supported schema repo with rails versions less than 6 so I'm dropping that unused code here.

Fryguy (approved):

PR #740, opened 2024-07-22
Ensure a minimum version of manageiq-style

Fryguy:
@jrafanie Please review.

Because schema works a little differently, we need to set this minimum level.


jrafanie (approved):
:shipit:

PR #750, opened 2024-08-30
Update dummy app to Rails 7.0.8.4

Fryguy:
This PR updates the dummy app to Rails 7.0.8.4 to match what we use in ManageIQ core.

```
rm -rf dummy/.git
```

I intentionally built the PR commits in a way that in the future, if we have to upgrade again, we can follow the commit's pattern. The commits:

- Delete the old app
- Create the new app
- Apply changes to the app in a single commit compiled from a number of commits over time

This pattern is similar to the one I took for https://github.com/ManageIQ/manageiq-schema/pull/616, and includes a cherry-pick of the commit from that PR.


spec/dummy/config/environments/test.rb
```diff
@@ -6,17 +8,18 @@
...
   config.cache_classes = true
 
-  # Do not eager load code on boot. This avoids loading your whole application
-  # just for the purpose of running a single test. If you are using a tool that
-  # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = false
+  # Eager loading loads your whole application. When running a single test locally,
+  # this probably isn't necessary. It's a good idea to do in a continuous integration
+  # system, or in some way before deploying your code.
+  config.eager_load = ENV["CI"].present?
```
Fryguy, line 17: This is the only one I'm not sure about, but I don't think it will hurt anything in reality.

jrafanie:
Very nice!  Side question, should we create new migrations with rails 7 and should we do it in this PR? See: https://github.com/ManageIQ/manageiq-schema/blob/master/Gemfile#L22

jrafanie:
> Very nice! Side question, should we create new migrations with rails 7 and should we do it in this PR? See: https://github.com/ManageIQ/manageiq-schema/blob/master/Gemfile#L22

lol, you did change it... my bad

Fryguy:
yeah I think we can start creating new migrations with Rails 7.  Worst case scenario if we need to downgrade, it's a simple find/replace.

spec/dummy/config/environments/test.rb
jrafanie: I agree.  I don't think it's a big deal.  It might lead to test contamination by loading things earlier than previously but who knows. It's a small dummy app.

Fryguy:
Backported to `radjabov` in commit 65993a71027cfef79d249f5066ce5074c2ab6b17.

```text
commit 65993a71027cfef79d249f5066ce5074c2ab6b17
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Fri Aug 30 14:35:28 2024 -0400

    Merge pull request #750 from Fryguy/dummy_rails7
    
    Update dummy app to Rails 7.0.8.4
    
    (cherry picked from commit 08097e7b5651c1b6d8966160ce93908051abe7f4)
```

PR #756, opened 2024-10-04
Add rails 7.1 and 7.2 support

jrafanie:
Depends on:

- [x] https://github.com/ManageIQ/activerecord-id_regions/pull/40 Merge and release

High level:

YAML `serialize` is going out of style and we should convert as many to json or other types if possible.  The interface for `serialize` changed so I have a bridge that changes our existing migrations to use the 7.1 interface with a layer to convert it back to 7.0 format if using 7.0.  I think this way is easier to read than the prior commit: [Prepend serialize in 7.1+ to convert positional arguments to kwargs](https://github.com/ManageIQ/manageiq-schema/pull/756/commits/737278c55b7571fc8b3f0e0d690c533fd65eb161)


spec/dummy/config/application.rb
```diff
@@ -23,6 +23,11 @@ class Application < Rails::Application
     # migration specs to honor it.
     config.active_record.belongs_to_required_by_default = false
 
+    # Note, you can't pass kwargs :coder => YAML to serialize until rails 7.1 as it was a positional
+    # argument previously.  To avoid a case statement in all usages of serialize, we're defaulting
+    # all serialized columns to YAML for rails 7.1+ here. Ideally, we would use JSON if we find we can
+    # use a simpler datatype. See: https://github.com/rails/rails/pull/47463
+    config.active_record.default_column_serializer = YAML if Rails.version >= "7.1"
```
jrafanie, line 30: TODO: Do we want to inject this for each of the migrations and use the "bridge" in a similar way I'm handling `:type` above?  We'd need to specify `:coder => YAML` for nearly all the migrations that do `serialize`.  This approach could help guide us on what to do in core in the models.

jrafanie: In other words, we can specify the column serializer/coder in each migration or here only once.  The negative of doing it here is it's less explicit and we're less likely to actually change to json or another / more supported coder.

cc @Fryguy @kbrock

jrafanie: On the flip side, I assume you can override the default in each migration and specify an override coder one by one so any that are serializing basic objects can likely specify json while the default for others will be to use yaml.

jrafanie:
Ok, https://github.com/ManageIQ/activerecord-id_regions/pull/40 was merged and released as 0.5.0 so this is now ready for review.

spec/dummy/config/initializers/serialize_positional_to_kwargs_bridge.rb
```diff
@@ -0,0 +1,19 @@
+module Dummy
+  module SerializePositionalToKwargsBridge
+    def serialize(*args, **options)
```
jrafanie, line 3: This is the largest change in rails 7.1+ for this repository.  Other than the "coder" change above, there isn't an option to default the `type` so what I did was accepted the new interface: kwargs for type and built a bridge to convert kwargs back to positional arguments when run with rails 7.0.

spec/dummy/config/application.rb
Fryguy: I'm trying to understand the implications of this change.

Ideally, ActiveRecord::Migration would know to use coder yaml for serialize on older migration versions that are indicated with the `[7.0]` thing in the constant name. To me, it's a bug that they didn't do this, so maybe we should open a bug with them.

That aside, I'm concerned if we set a global default, then it will break future migrations. So, considering that I think it's better, though more verbose, to set the coder => yaml on every migration.  We're already changing it for the :type change, so it's not like it's more lines of code.  This way we can do it the "right way" moving forward.

---

Related question, what is the coder on the model side?  Do we need a similar change in core changing every serialized column to say what coder it is?

spec/dummy/config/initializers/serialize_positional_to_kwargs_bridge.rb
```diff
@@ -0,0 +1,19 @@
+module Dummy
```
Fryguy, line 1: This feels like it should be part of the plugin itself, and not dummy.  The dummy app is only for tests here, so I presume we will hit this same issue in application code where the dummy won't be present.  Perhaps it could be done similar to how we [patch migrate for clearing caches](https://github.com/ManageIQ/manageiq-schema/blob/master/lib/manageiq/schema/migrate_with_cleared_schema_cache.rb).

spec/dummy/config/application.rb
kbrock: What is in our way from converting all our serialized fields from yaml => json?
What do we see as the biggest barier

- Time to write migration code
- Time to run migrations on customer code (upgrades)
- Possible datatype incompatibility (symbols, dates, active record)
- Test that app is still working

jrafanie: > Ideally, ActiveRecord::Migration would know to use coder yaml for serialize on older migration versions that are indicated with the `[7.0]` thing in the constant name. To me, it's a bug that they didn't do this, so maybe we should open a bug with them.

Technically, this is a gray area because we're defining stub models in migrations.  I don't think they can guard for that type of thing.  Basically anything you put in a stub model could require their knowledge of and include versioned logic for handling them.  I can see why they wouldn't want to handle that.

If they're text fields, you can define your accessors and do YAML dump/load or to_json / JSON parse in those accessors.  Perhaps, that's what we should do in stub models.  We avoid defining has_many/belongs_to in them for similar reasons so we can make changes and avoid automagic stuff done by those methods.

jrafanie: > That aside, I'm concerned if we set a global default, then it will break future migrations. So, considering that I think it's better, though more verbose, to set the coder => yaml on every migration. We're already changing it for the :type change, so it's not like it's more lines of code. This way we can do it the "right way" moving forward.

Yes, I agree.  there is no global for :type column and I did the YAML default coder first so I came to a similar conclusion.  

I first did the `:type` column with 7.0 as the target by converting 7.1+ kwargs to positional.  That was pretty nasty looking so instead I changed it to target 7.1+ kwargs interface and convert kwargs to positional arguments in case you're running 7.0.  I like that better.  That's what is here now.

I would be fine with doing the same with YAML as the coder in each migration.  It's pretty easy to undo the default coder and fix all the broken serialize calls missing a coder and handle it the same way as I converted kwargs to positional in the event it's running on rails 7.0

> Related question, what is the coder on the model side? Do we need a similar change in core changing every serialized column to say what coder it is?

Depends on version.  YAML is the default in 7.0. In 7.1 it's `nil`.

Yes, we have the same problem to solve in core and plugins.  That's why I started here as it's much easier to review on a much smaller dummy application with models as only stub models (thank you data migrations).

See also: https://www.github.com/rails/rails/pull/47422 https://www.github.com/rails/rails/pull/47463

jrafanie: > What is in our way from converting all our serialized fields from yaml => json? What do we see as the biggest barier
> 
> * Time to write migration code
> * Time to run migrations on customer code (upgrades)
> * Possible datatype incompatibility (symbols, dates, active record)
> * Test that app is still working

Anyplace we store MiqExpression, MiqReport, etc. objects in serialized columns is where I'd be most concerned.  I don't know if basic types would come back in the same format we're expecting but I would think those are much easier to deal with than custom objects.

spec/dummy/config/initializers/serialize_positional_to_kwargs_bridge.rb
jrafanie: Nice, I can look into that.  Good idea.

spec/dummy/config/application.rb
jrafanie: > Related question, what is the coder on the model side? Do we need a similar change in core changing every serialized column to say what coder it is?

You can see some of the [very wet stuff on the core PR](https://github.com/ManageIQ/manageiq/pull/23225/files)... so far, type is where I did changes since I could specify a default coder using `config.active_record.default_column_serializer = YAML`.  I would prefer a common approach for both and I like the idea of writing serialize in the `rails 7.1+` kwargs way and the bridge module would convert it to positional arguments for rails 7.0.

Fryguy: > What is in our way from converting all our serialized fields from yaml => json?

non-json types, notable Ruby objects and DateTime.  If we know for a fact a column doesn't have those it's an easy conversion.

However, that's moving forward - we still need to deal with older migrations regardless.

Fryguy: > Technically, this is a gray area because we're defining stub models in migrations. I don't think they can guard for that type of thing.

'nuff said.  My brain didn't click that this was a model problem, not really a migration problem.

spec/lib/generators/migration_generator_spec.rb
```diff
@@ -3,7 +3,12 @@
 require 'generators/migration/migration_generator'
 
 describe ManageIQ::Schema::MigrationGenerator do
-  include Rails::Generators::Testing::Behaviour
+  if Rails.version >= "7.1"
+    include Rails::Generators::Testing::Behavior
+  else
+    include Rails::Generators::Testing::Behaviour
```
jrafanie, line 9: 🇬🇧

spec/lib/generators/migration_generator_spec.rb
```diff
@@ -3,7 +3,12 @@
...
+  end
```
Fryguy, line 10: Really? 😞

spec/dummy/config/application.rb
jrafanie: By the way, I kept the first attempt commit.  I had the bridge module implementing "7.0 format" we don't need to change model calls to serialize but it's ugly:
https://github.com/ManageIQ/manageiq-schema/pull/756/commits/737278c55b7571fc8b3f0e0d690c533fd65eb161

Compare that to the "7.1 format" where we convert 7.0 format to 7.1 we have now https://github.com/ManageIQ/manageiq-schema/pull/756/commits/410fbfcdabbec411accaf0091e59f4abf3fd3164

I kept it around to help decide which way forward would be best.

I'll mark this was WIP again and apply the changes suggested here:
1) try moving the module to the plugin instead of the dummy app
2) treat coder the same way as type, convert all serialize calls to specify a YAML coder and remove the default column serializer setting.

spec/lib/generators/migration_generator_spec.rb
jrafanie: It was more a "consistency in naming" vs. 🇬🇧 / 🇺🇸  spelling fight. https://www.github.com/rails/rails/pull/45180

I do prefer the fight though.  It's more fun.

spec/dummy/config/application.rb
jrafanie: Ok, I tried removing `config.active_record.default_column_serializer = YAML if Rails.version >= "7.1"` and defaulted to setting `:coder => YAML` throughout but that didn't work well with 7.0.  I removed that change.  We'll carry this for now and revisit it once we're off of 7.0.

jrafanie:
Ok, I'm taking this out of WIP.  I suspect we'll do something similar in serialize calls in the models in core.

PR #757, opened 2024-10-17
Add a migration to encrypt database password using scram-sha-256

bdunne:
Related:
- https://github.com/ManageIQ/manageiq-appliance/pull/388
- https://github.com/ManageIQ/manageiq-pods/pull/1157

CP4AIOPS-3003

### DIFF
--- a/db/migrate/20241017013023_reencrypt_password_scramsha.rb
+++ b/db/migrate/20241017013023_reencrypt_password_scramsha.rb
@@ -1,0 +1,13 @@
+class ReencryptPasswordScramsha < ActiveRecord::Migration[6.1]
+  def up
+    say_with_time('Reencrypting database user password with scram-sha-256') do
+      db_config = ActiveRecord::Base.connection_db_config.configuration_hash
+      username = db_config[:username]
+      password = connection.raw_connection.encrypt_password(db_config[:password], username, "scram-sha-256")
+
+      connection.execute <<-SQL
+        ALTER ROLE #{username} WITH PASSWORD '#{password}';
+      SQL
+    end
+  end
+end
--- a/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
+++ b/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
@@ -1,0 +1,22 @@
+require_migration
+
+# This is mostly necessary for data migrations, so feel free to delete this
+# file if you do no need it.
+describe ReencryptPasswordScramsha do
+  migration_context :up do
+    it "Ensures that the user password is stored as scram-sha-256" do
+      migrate
+
+      username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
+
+      users_and_passwords = ActiveRecord::Base.connection.execute <<-SQL
+        SELECT rolname, rolpassword FROM pg_authid WHERE rolcanlogin;
+      SQL
+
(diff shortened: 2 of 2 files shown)


db/migrate/20241017013023_reencrypt_password_scramsha.rb
```diff
@@ -0,0 +1,17 @@
...
+      database_yml = YAML.safe_load(Rails.root.join("config", "database.yml"))
+      username = database_yml[Rails.env]["username"]
+      password = MiqPassword.try_decrypt(database_yml[Rails.env]["password"])
+
+      execute <<-SQL
+        SET password_encryption='scram-sha-256';
+      SQL
+
+      execute <<-SQL
+        ALTER ROLE #{username} WITH PASSWORD '#{password}';
```
Fryguy, line 13: As discussed, let's not send the bare password, but instead pre-encrypt it.

```
password = connection.raw_connection.encrypt_password(username, password, "scram-sha-256")
```

Fryguy:
Note that if anyone has their database locally setup as md5 (or just on a old version), this will break their local database.  I'm ok with that, but we might need some instructions either here or in guides/dev-setup on the changes needed to unbreak them.

db/migrate/20241017013023_reencrypt_password_scramsha.rb
```diff
@@ -0,0 +1,17 @@
+class ReencryptPasswordScramsha < ActiveRecord::Migration[6.1]
+  def up
+    say_with_time('Reencrypting database user password with scram-sha-256') do
+      database_yml = YAML.safe_load(Rails.root.join("config", "database.yml"))
```
Fryguy, line 4: There are cases where we could use DATABASE_URL, so I think it's better to use the database configuration instead. Additionally, database confiuguration should have already decrypted the password for us, so there's no need to further try_decrypt:

```
ActiveRecord::Base.connection_db_config.configuration_hash
# => {:adapter=>"postgresql", :encoding=>"utf8", :username=>"root", :pool=>"5", :wait_timeout=>5, :min_messages=>"notice", :database=>"vmdb_development", :password=>"smartvm", :port=>5432, :host=>"localhost"}
```

db/migrate/20241017013023_reencrypt_password_scramsha.rb
bdunne: Done

db/migrate/20241017013023_reencrypt_password_scramsha.rb
bdunne: Done

Fryguy (approved):

kbrock:
I think the only concern here is the plaintext password will be printed into the pg logs in some cases. This is from what I've read and I have not researched/looked too deep into this

Fryguy:
Backported to `radjabov` in commit 37e87fa519bb9df9f8fe41d9c005b9b1790b4a32.

```text
commit 37e87fa519bb9df9f8fe41d9c005b9b1790b4a32
Date:   Wed Oct 23 16:29:10 2024 -0400

    Merge pull request #757 from bdunne/scramsha
    
    Add a migration to encrypt database password using scram-sha-256
    
    (cherry picked from commit 3762bf55c1f5037c7e5421ec426ecbb9dad12fdb)
```

Fryguy:
@kbrock the way it's done here encryption the password first (in memory) then sends that, so it should not be in the logs

PR #763, opened 2024-10-24
Handle cases where the database connection does not use a password

bdunne:
Introduced in #757

### DIFF
--- a/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
+++ b/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
@@ -5,18 +5,31 @@ require_migration
 describe ReencryptPasswordScramsha do
   migration_context :up do
     it "Ensures that the user password is stored as scram-sha-256" do
-      migrate
+      allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      users_and_passwords = ActiveRecord::Base.connection.execute <<-SQL
-        SELECT rolname, rolpassword FROM pg_authid WHERE rolcanlogin;
-      SQL
+      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
+      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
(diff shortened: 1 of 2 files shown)


spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
```diff
@@ -5,18 +5,31 @@
...
     it "Ensures that the user password is stored as scram-sha-256" do
-      migrate
+      allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      users_and_passwords = ActiveRecord::Base.connection.execute <<-SQL
-        SELECT rolname, rolpassword FROM pg_authid WHERE rolcanlogin;
-      SQL
+      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
```
bdunne, line 12: This was an interesting find.

agrare: Don't love adding "real" code only to support tests, but maybe worth it in this case.  You could add a method that returns the connection_db_config in the migration class and stub that method in this spec

Fryguy: Why exactly 10 times as opposed to something like at least once?

Fryguy:
@miq-bot cross-repo-test manageiq, manageiq-ui-classic, manageiq-api

spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
agrare: I'm guessing because on the 11th time he needs to return a different payload

Fryguy:
I'm going to merge this to unstick the rest of the repositories but I think it's worth looking into a better way to test this.

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit 7400a8abb8e617b81a06945179db8dbdf2118c95.

```text
commit 7400a8abb8e617b81a06945179db8dbdf2118c95
Date:   Thu Oct 24 14:36:23 2024 -0400

    Merge pull request #763 from bdunne/no_password
    
    Handle cases where the database connection does not use a password
    
    (cherry picked from commit 6c25437ca4b62489ef801d2ab21f00da1674699c)
```

PR #776, opened 2024-12-12
Allow the rails method to be called and change the result only when called from the migration

### LINKED ISSUE #768: Fix scramsha migration test failures on rails 7.1
From https://github.com/ManageIQ/manageiq-schema/pull/767
Seen in https://github.com/ManageIQ/manageiq-schema/pull/766

```
1) ReencryptPasswordScramsha#up Ensures that the user password is stored as scram-sha-256
     Failure/Error: expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original

       (#<ActiveRecord::DatabaseConfigurations::HashConfig:0x00007f4cdb38ad18 @env_name="test", @name="primary", @configuration_hash={:adapter=>"postgresql", :encoding=>"utf8", :username=>"root", :***"smartvm", :pool=>3, :wait_timeout=>5, :min_messages=>"warning", :database=>"dummy_test"}>).configuration_hash(*(any args))
           expected: 10 times with any arguments
           received: 1 time with any arguments
     # ./spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb:12:in `block (3 levels) in <top (required)>'
     # ./spec/support/migration_helper.rb:77:in `clearing_caches'
     # ./spec/support/migration_helper.rb:13:in `block (2 levels) in migration_context'

  2) ReencryptPasswordScramsha#up Handles connections with no password
     Failure/Error: super

       (#<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x00007f4cda51acb0 @transaction_manager=#<ActiveRecord::ConnectionAdapters::TransactionManager:0x00007f4cda920a80 @stack=[#<ActiveRecord::ConnectionAdapters::RealTransaction:0x00007f4cd3e9c[47](https://github.com/ManageIQ/manageiq-schema/actions/runs/11858845607/job/33050405714?pr=766#step:6:48)0 @connection=#<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x00007f4cda51acb0 ...>, @state=#<ActiveRecord::ConnectionAdapters::TransactionState:0x00007f4cd3e9c420 @state=nil, @children=[#<ActiveRecord::ConnectionAdapters::TransactionState:0x00007f4cd395a930 @state=nil, @children=nil>]>, @records=nil, @isolation_level=nil, @materialized=true, @joinable=false, @run_commit_callbacks=true, @lazy_enrollment_records=nil, @dirty=true, @instrumenter=#<ActiveRecord::ConnectionAdapters::TransactionInstrumenter:0x00007f4cd3e9c290 @handle=#<ActiveSupport::Notifications::Fanout::Handle:0x00007f4cd3ea1718 @name="transaction.active_record", @id="4321fe7d6758014edd25", @payload={:connection=>#<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x00007f4cda[51](https://github.com/ManageIQ/manageiq-schema/actions/runs/11858845607/job/33050405714?pr=766#step:6:52)acb0 ...>}, @groups=[],
```

jrafanie:
In rails 7.0, it's called 10 times plus 1 more on the next line.  In rails 7.1+, it's called only once.  From a test perspective, we don't care if it's even called. We only care that the SQL query is executed with the scramsha password or not.

To change the username/password only in calls from the migration itself and not from
rails, we hack around the caller_locations to make the change only for the migration callers.

With this, I believe we can expect test success on rails 7.1 and 7.2.

It was added in 742841cf8954be1cb5843dca33945b5fc001cf8d but I think this change emphasizes what we really care about testing and removes an assertion on internals that we don't really have interest in testing.

Fixes #768

### DIFF
--- a/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
+++ b/spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
@@ -9,9 +9,9 @@ describe ReencryptPasswordScramsha do
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
-        original_method.call(*args, &block).dup.tap { |i| i[:password] ||= "abc" }
+      allow(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
+        # set a value for any calls originating from the migration file, not from rails itself
+        original_method.call(*args, &block).dup.tap {|i| i[:password] ||= "abc" if caller_locations.any? {|loc| loc.path.include?("db/migrate/20241017013023_reencrypt_password_scramsha.rb")} }
       end
       expect(ActiveRecord::Base.connection).to receive(:execute).with(a_string_matching(/ALTER ROLE #{username} WITH PASSWORD \'SCRAM-SHA-256.*\'\;/))
 
@@ -23,9 +23,9 @@ describe ReencryptPasswordScramsha do
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
-        original_method.call(*args, &block).dup.tap { |i| i.delete(:password) }
+      allow(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
+        # set a value for any calls originating from the migration file, not from rails itself
+        original_method.call(*args, &block).dup.tap { |i| i.delete(:password) if caller_locations.any? {|loc| loc.path.include?("db/migrate/20241017013023_reencrypt_password_scramsha.rb")} }
       end
       expect(ActiveRecord::Base.connection).not_to receive(:execute).with(a_string_matching(/ALTER ROLE.*\'\;/))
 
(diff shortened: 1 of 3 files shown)


spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
```diff
@@ -23,9 +23,9 @@
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
-        original_method.call(*args, &block).dup.tap { |i| i.delete(:password) }
+      allow(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
+        # set a value for any calls originating from the migration file, not from rails itself
+        original_method.call(*args, &block).dup.tap { |i| i.delete(:password) if caller_locations.any? {|loc| loc.path.include?("db/migrate/20241017013023_reencrypt_password_scramsha.rb")} }
```
jrafanie, line 28: Note, this is still super brittle but that's because the test is modifying the database configuration in the context of the test and code outside our migration can be impacted by that change.  This is why we had 10 calls originally in rails 7.0 and only 1 in 7.1.

spec/migrations/20241017013023_reencrypt_password_scramsha_spec.rb
```diff
@@ -9,9 +9,9 @@
 
       username = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
 
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).exactly(10).times.and_call_original
-      expect(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
-        original_method.call(*args, &block).dup.tap { |i| i[:password] ||= "abc" }
+      allow(ActiveRecord::Base.connection_db_config).to receive(:configuration_hash).and_wrap_original do |original_method, *args, &block|
+        # set a value for any calls originating from the migration file, not from rails itself
+        original_method.call(*args, &block).dup.tap {|i| i[:password] ||= "abc" if caller_locations.any? {|loc| loc.path.include?("db/migrate/20241017013023_reencrypt_password_scramsha.rb")} }
```
Fryguy, line 14: Extremely minor but would it be worth extracting this helper code into a new helper method and then using that here?

https://github.com/ManageIQ/manageiq-schema/blob/7af242029e4c111510c37bf32911e3749414a51c/spec/support/migration_name_helper.rb#L2-L3

jrafanie: sure, I'll take a look when I come back to this.  👍

PR #783, opened 2025-02-17
Create SettingsChanges from BlacklistedEvents

agrare:
Create SettingsChange records for any user created BlacklistedEvents.

This makes the assumption that all `BlacklistedEvent.where(:system => true)` already have SettingsChange records because that is how they would be seeded.

This only considers user generated BlacklistedEvent records via rails-console that would have `system => false`

Since we can't know which MiqServer a BlacklistedEvent would have been created for this uses `MiqRegion.my_region` as the `SettingsChange#resource`

Required for:
* https://github.com/ManageIQ/manageiq/pull/23320

https://github.com/ManageIQ/manageiq/issues/19707

### DIFF
--- a/db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
+++ b/db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
@@ -3,6 +3,10 @@ class MoveBlacklistedEventsToSettingsChanges < ActiveRecord::Migration[7.0]
     include ActiveRecord::IdRegions
   end
 
+  class MiqRegion < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
   class SettingsChange < ActiveRecord::Base
     include ActiveRecord::IdRegions
 
@@ -52,6 +56,9 @@ class MoveBlacklistedEventsToSettingsChanges < ActiveRecord::Migration[7.0]
 
   def up
     say_with_time("Moving BlacklistedEvent records to SettingsChanges") do
+      my_region_id = MiqRegion.find_by(:region => MiqRegion.my_region_number)&.id
+      return if my_region_id.nil?
+
       BlacklistedEvent.in_my_region.where(:system => false, :enabled => true).group_by(&:provider_model).each do |provider_model, blacklisted_events|
         addition_blacklisted_events = blacklisted_events.pluck(:event_name) - DEFAULT_BLACKLISTED_EVENTS[provider_model]
         next if addition_blacklisted_events.empty?
@@ -59,7 +66,7 @@ class MoveBlacklistedEventsToSettingsChanges < ActiveRecord::Migration[7.0]
         settings_change_key = PROVIDER_MODEL_TO_SETTINGS_KEY[provider_model]
         next if settings_change_key.nil?
 
-        settings_change = SettingsChange.find_or_initialize_by(:resource_type => "MiqRegion", :resource_id => SettingsChange.my_region_number, :key => settings_change_key)
+        settings_change = SettingsChange.find_or_initialize_by(:resource_type => "MiqRegion", :resource_id => my_region_id, :key => settings_change_key)
         settings_change.update!(:value => DEFAULT_BLACKLISTED_EVENTS[provider_model] + addition_blacklisted_events)
       end
     end
--- a/db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
+++ b/db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
@@ -1,0 +1,84 @@
+class MoveBlacklistedEventsToSettingsChanges < ActiveRecord::Migration[7.0]
+  class BlacklistedEvent < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  class SettingsChange < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+
+    serialize :value
+  end
+
+  # The list of BlacklistedEvents without any changes to Settings
+  # Any BlacklistedEvents that are not in this list must be added to SettingsChanges
+  DEFAULT_BLACKLISTED_EVENTS = {
+    "ManageIQ::Providers::Amazon::CloudManager"                      => ["ConfigurationSnapshotDeliveryCompleted", "ConfigurationSnapshotDeliveryStarted", "ConfigurationSnapshotDeliveryFailed"],
+    "ManageIQ::Providers::Amazon::ContainerManager"                  => ["ConfigurationSnapshotDeliveryCompleted", "ConfigurationSnapshotDeliveryStarted", "ConfigurationSnapshotDeliveryFailed"],
+    "ManageIQ::Providers::Amazon::NetworkManager"                    => ["ConfigurationSnapshotDeliveryCompleted", "ConfigurationSnapshotDeliveryStarted", "ConfigurationSnapshotDeliveryFailed"],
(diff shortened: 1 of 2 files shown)


agrare:
Still WIP but this is what the conversion might look like, the list of default blacklisted events was just taken from seeding a clean database with no additional filtered events created.

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  SETTINGS_KEY_TO_PROVIDER_MODEL = {
+    "/ems/ems_vmware/blacklisted_event_names" => "ManageIQ::Providers::Vmware::InfraManager"
+  }.freeze
+
+  def up
+    say_with_time("Moving BlacklistedEvent records to SettingsChanges") do
+      BlacklistedEvent.in_my_region.where(:system => false, :enabled => true).group_by(&:provider_model).each do |provider_model, blacklisted_events|
+        addition_blacklisted_events = blacklisted_events.pluck(:event_name) - DEFAULT_BLACKLISTED_EVENTS[provider_model]
+        next if addition_blacklisted_events.empty?
```
agrare, line 58: NOTE I'm not sure it is worth it performance wise to build a bunch of `.or.where(:provider_model => a, :event_name => b)` vs just looping through them all

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  # TODO add the rest of the provider settings keys
+  # This is the mapping of BlacklistedEvent#provider_model -> SettingsChange#key
+  PROVIDER_MODEL_TO_SETTINGS_KEY = {
+    "ManageIQ::Providers::Vmware::InfraManager" => "/ems/ems_vmware/blacklisted_event_names"
+  }.freeze
+
+  SETTINGS_KEY_TO_PROVIDER_MODEL = {
+    "/ems/ems_vmware/blacklisted_event_names" => "ManageIQ::Providers::Vmware::InfraManager"
+  }.freeze
```
Fryguy, line 52: Assuming the right and left are both unique

```suggestion
  SETTINGS_KEY_TO_PROVIDER_MODEL = PROVIDER_MODEL_TO_SETTINGS_KEY.invert
```

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
Fryguy: +1 for just looping through them all.

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
agrare: I'm hoping so yeah, it uses the vendor not the provider and some providers override it so I need to prove that out

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
```diff
@@ -0,0 +1,85 @@
...
+  def up
+    say_with_time("Moving BlacklistedEvent records to SettingsChanges") do
+      BlacklistedEvent.in_my_region.where(:system => false, :enabled => true).group_by(&:provider_model).each do |provider_model, blacklisted_events|
+        addition_blacklisted_events = blacklisted_events.pluck(:event_name) - DEFAULT_BLACKLISTED_EVENTS[provider_model]
+        next if addition_blacklisted_events.empty?
+
+        settings_change_key = PROVIDER_MODEL_TO_SETTINGS_KEY[provider_model]
+        next if settings_change_key.nil?
+
+        settings_change = SettingsChange.find_or_initialize_by(:resource_type => "MiqRegion", :resource_id => SettingsChange.my_region_number, :key => settings_change_key)
```
Fryguy, line 63: The region number doesn't match the MiqRegion `id` column...For example, in region number `0`, the id would probably be `1`, and in region `1` the id would probably be `1000000000001`.

```suggestion
        settings_change = SettingsChange.find_or_initialize_by(:resource_type => "MiqRegion", :resource_id => MiqRegion.my_region.id, :key => settings_change_key)
```

Fryguy: You're probably going to have to add an MiqRegion stub for this, annoyingly...you can copy it from another migration.

Fryguy:
This approach looks good to me.  Found one bug above with the region id, but otherwise looks good.

db/migrate/20250214153835_move_blacklisted_events_to_settings_changes.rb
```diff
@@ -0,0 +1,119 @@
...
+    "ManageIQ::Providers::Ovirt::NetworkManager"                 => "/ems/ems_ovirt_network/blacklisted_event_names",
+    "ManageIQ::Providers::Redfish::PhysicalInfraManager"         => "/ems/ems_redfish/blacklisted_event_names",
+    "ManageIQ::Providers::Redhat::InfraManager"                  => "/ems/ems_redhat/blacklisted_event_names",
+    "ManageIQ::Providers::Redhat::NetworkManager"                => "/ems/ems_redhat_network/blacklisted_event_names",
+    "ManageIQ::Providers::Vmware::InfraManager"                  => "/ems/ems_vmware/blacklisted_event_names",
+    "ManageIQ::Providers::Vmware::CloudManager"                  => "/ems/ems_vmware_cloud/blacklisted_event_names",
+    "ManageIQ::Providers::Vmware::NetworkManager"                => "/ems/ems_vmware_cloud_network/blacklisted_event_names",
+    "ManageIQ::Providers::Vmware::ContainerManager"              => "/ems/ems_vmware_tanzu/blacklisted_event_names",
+    "ManageIQ::Providers::Workflows::AutomationManager"          => "/ems/ems_workflows/blacklisted_event_names"
+  }.freeze
```
agrare, line 81: This was a bit interesting.  Because this was using the vendor instead of the ems_type there were some classes (e.g. Amazon::CloudManager, Amazon::NetworkManager, Amazon::ContainerManager) which had duplicate blacklisted events because they were all using the `Settings.ems.ems_amazon.blacklisted_event_names` setting.

I need to have this full list above so I can skip all of these as default, but I don't want or need to replicate that here and this can be a straight map of 1 EMS class to 1 settings key.

I got the list of settings keys by iterating through `Settings.ems.keys` where there was a `Settings.ems.send(key).key?(:blacklisted_event_names)` key.  I then manually assigned what I knew to be the "main" EMS class.

agrare: Also interesting, there seems to have been some copy&paste going on here as there are a number of providers with a blacklisted_event_names setting but no event catcher.

```
>> PROVIDER_MODEL_TO_SETTINGS_KEY.keys.select { |klass| "#{klass}::EventCatcher".safe_constantize.to_s != "#{klass}::EventCatcher" }
=> 
["ManageIQ::Providers::EmbeddedTerraform::AutomationManager",
 "ManageIQ::Providers::Foreman::ConfigurationManager",
 "ManageIQ::Providers::IbmTerraform::ConfigurationManager",
 "ManageIQ::Providers::Nsxt::NetworkManager",
 "ManageIQ::Providers::Ovirt::NetworkManager",
 "ManageIQ::Providers::Redhat::NetworkManager",
 "ManageIQ::Providers::Vmware::NetworkManager",
 "ManageIQ::Providers::Workflows::AutomationManager"]
```

Fryguy (approved):

PR #792, opened 2025-06-03
Drop tenant use_config_for_attributes

kbrock:
no longer using SettingsChanges for the root tenant name

part of https://github.com/ManageIQ/manageiq/pull/23463

### DIFF
--- a/db/migrate/20250603164430_update_tenant_name.rb
+++ b/db/migrate/20250603164430_update_tenant_name.rb
@@ -1,0 +1,22 @@
+class UpdateTenantName < ActiveRecord::Migration[7.0]
+  class SettingsChange < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  class Tenant < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def change
+    say_with_time("Updating root tenant's name from settings") do
+      sc = SettingsChange.in_my_region.find_by(:key => "/server/company")
+      Tenant.in_my_region.find_by(:ancestry => nil, :use_config_for_attributes => true)&.update(:name => sc&.value || "My Company", :use_config_for_attributes => false)
+      SettingsChange.in_my_region.where(:key => "/server/company").destroy_all
+    end
+  end
+
+  def revert
+    # no real way to revert
+    # the record remains as :use_config_for_attributes => false, and Tenant.root.name is used
+  end
+end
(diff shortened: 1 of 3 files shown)


kbrock:
update:
- fixed issue with empty database
- added test case for no settings override (using default)

db/migrate/20250603164430_update_tenant_name.rb
```diff
@@ -0,0 +1,21 @@
...
+  def change
+    # there may be multiple of these, but it should only be
+    sc = SettingsChange.in_my_region.find_by(:key => "/server/company")
+    Tenant.find_by(:ancestry => nil, :use_config_for_attributes => true)&.update(:name => sc&.value || "My Company", :use_config_for_attributes => false)
+    SettingsChange.in_my_region.where(:key => "/server/company").destroy_all
+  end
+
+  def revert
+    # no real way to revert
+    # but since the tenant record no longer looks into server/company record, it is fine
```
Fryguy, line 19: In a revert situation we'd be using old code, so it _would be loking at the settings.  I think this comment should be removed.

kbrock:
update:
- added say_with_time

db/migrate/20250603164430_update_tenant_name.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  class Tenant < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def change
+    say_with_time("Updating root tenant's name from settings") do
+      # there may be multiple of these, but it should only be
+      sc = SettingsChange.in_my_region.find_by(:key => "/server/company")
+      Tenant.find_by(:ancestry => nil, :use_config_for_attributes => true)&.update(:name => sc&.value || "My Company", :use_config_for_attributes => false)
```
Fryguy, line 14: I think there's a bug here? What about the case where the Tenant has a name _and_ there isn't a value in /server/company. Then I think this changes the Tenant name to whatever the user set to "My Company".  Is that even an expected combination?

db/migrate/20250603164430_update_tenant_name.rb
```diff
@@ -0,0 +1,23 @@
...
+    include ActiveRecord::IdRegions
+  end
+
+  class Tenant < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+  end
+
+  def change
+    say_with_time("Updating root tenant's name from settings") do
+      # there may be multiple of these, but it should only be
```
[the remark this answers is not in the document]
kbrock: if a region has multiple servers each with a different company name, then this is the use case that has been removed. Anything goes.

kbrock:
update:
- `in_my_region` fix

PR #799, opened 2025-07-30
show_exceptions no longer supports true|false

jrafanie:
See: https://www.github.com/rails/rails/commit/ec2c2666c255029593dda2f78e89b457cf8509fc

They now support :none, :rescuable, and :all

See also https://github.com/ManageIQ/manageiq/pull/23531


jrafanie:
See also discussion in https://github.com/ManageIQ/manageiq/pull/23531#issuecomment-3137403298

PR #802, opened 2025-08-06
Update dummy app to 7.2 drop 7.0

jrafanie:
This is a copy of the pattern from https://github.com/ManageIQ/manageiq-schema/pull/750

This PR updates the dummy app to Rails 7.2.2.1 to match what we use in ManageIQ core.

```
rm -rf dummy/.git
```

I intentionally built the PR commits in a way that in the future, if we have to upgrade again, we can follow the commit's pattern. The commits:

- Delete the old app
- Create the new app
- Apply changes to the app in a single commit compiled from a number of commits over time

### Dummy
* Cherrypicked and edited c89dbcfd89bead6f2dc7cfc1a7ef3e32b62e1073 (the rolled up commit with changes for the dummy app
  * Dropped 7e34b02c551d25892511e702029900fb049ec39b (dummy/Gemfile is removed - no need to patch it)
  * Added 9ad9001485588cdd8742e4f54cb95c42c0499e34 (YAML coder default)

### For schema (I can extract if desired)
* Drop rails 7.0 support and serialize positional arg bridge
* Default to 7.2 for new migrations and code coverage


Fryguy:
So, I suggest squashing https://github.com/ManageIQ/manageiq-schema/pull/802/commits/e992889a66d8b431505840be6c9c2b7ba8c0e262 into https://github.com/ManageIQ/manageiq-schema/pull/802/commits/f49e640d6d3d809152b3235fcd0001f3e521bf18, because in the future we'll want to cherry-pick that "3rd" commit when we go to 8.0

.github/workflows/ci.yaml
```diff
@@ -16,7 +16,6 @@ jobs:
         - '3.3'
         - '3.4'
         rails-version:
-        - '7.1'
```
jrafanie, line 19: Since dummy app is now 7.2, we can't test with 7.1.

lib/manageiq/schema/engine.rb
```diff
@@ -9,13 +9,11 @@ class Engine < ::Rails::Engine
 
       ActiveSupport.on_load(:active_record) do
         require_relative 'migrate_with_cleared_schema_cache'
-        require_relative 'serialize_positional_to_kwargs_bridge'
```
jrafanie, line 12: Can be deleted after dropping 7.0 support. 🔥

manageiq-schema.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "manageiq-password",       ">= 1.2.0", "< 2"
   spec.add_dependency "more_core_extensions",    ">= 3.5", "< 5"
   spec.add_dependency "pg"
-  spec.add_dependency "rails",                   ">=7.0.8.7", "<8.0"
+  spec.add_dependency "rails",                   ">=7.2.2.1", "<8.0"
```
jrafanie, line 28: Drop 7.0 and 7.1 since we no longer test them and 7.2 is the default for the dummy app, migrations, and rails defaults.

PR #819, opened 2025-10-29
Upgrade rails to 7.2.3

Fryguy:


jrafanie (approved):
Related to https://github.com/ManageIQ/manageiq/pull/23641.  

Merging as is knowing the security failure is wrong.  https://www.github.com/rubysec/ruby-advisory-db/issues/915 describes the problem and https://www.github.com/rubysec/ruby-advisory-db/pull/916 should fix it.

PR #821, opened 2025-11-05
Drop miq_schedules.file_depot_id

Fryguy:
Depends on
- [x] https://github.com/ManageIQ/manageiq/pull/23649


jrafanie (approved):
LGTM... so many removals!

PR #824, opened 2025-12-12
Support Rails 8, drop EOL ruby 3.1

jrafanie:
* Add support for rails 8.0, drop EOL ruby 3.1
* Overload migrate, forwarding args/kwargs, in a compatible way
* 8.0 supports inherits in schema dump but doesn't handle table order
* Change metric* tables schema dumper tests for new behavior
* Add a warning so we're reminded to remove the backport patch and conditionals
* Add documentation in the README for verifying the schema dumper
* Support db:migrate:reset by looking for no versions(7.2) or 0(8.0)

### DIFF
--- a/db/migrate/20171030131403_collapsed_initial_migration.rb
+++ b/db/migrate/20171030131403_collapsed_initial_migration.rb
@@ -10,10 +10,12 @@ class CollapsedInitialMigration < ActiveRecord::Migration[5.0]
     # If a user has not upgraded to at least ManageIQ Gaprindashvili, then they
     #   will have entries in their schema_migrations table, but will not have
     #   migration 20171030131403. This migration will run and disallow that case.
-    if table_exists?("schema_migrations") &&
-       connection.select_value("SELECT COUNT(*) FROM schema_migrations") != 0
-      raise "This database cannot be migrated. You must first upgrade this " \
-            "database using the ManageIQ Gaprindashvili release."
+    if table_exists?("schema_migrations")
+      versions = connection.select_values("SELECT version FROM schema_migrations")
+      if versions.present? && versions != ["0"]
+        raise "This database cannot be migrated. You must first upgrade this " \
+              "database using the ManageIQ Gaprindashvili release."
+      end
     end
 
     # This collapsed initial migration is partly generated from Rails' schema
--- a/lib/tasks_private/spec.rake
+++ b/lib/tasks_private/spec.rake
@@ -13,10 +13,8 @@ namespace :spec do
     puts "** Creating database with REGION #{ENV["REGION"]}"
   end
 
-  # TODO: Leverage rails 8 feature: "db:migrate loads from schema by default for fresh databases"
-  # Note, db:migrate:reset does drop, create, load schema, migrate, our collapsed migration expects no schema_migrations
-  # version, but this is populated after schema load.
-  task :setup_db => [:setup_region, "db:drop", "db:create", "db:migrate"]
+  # Note, db:migrate:reset does drop, create, load schema, migrate
+  task :setup_db => [:setup_region, "db:migrate:reset"]
 
   desc "Prepare all specs"
   task :setup => [:initialize, :setup_db]


.github/workflows/ci.yaml
```diff
@@ -19,11 +19,11 @@ jobs:
     strategy:
       matrix:
         ruby-version:
-        - '3.1'
```
Fryguy: Oh nice - I missed this one - my rudimentary "is this an MIQ repo" checked exactly 3.1 and 3.3 (which is a weird combo, and why I thought it was a safe heuristic), but this repo apparently also had 3.4

Fryguy (commented):
A few suggestions

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -4,19 +4,110 @@ module SchemaDumper
       METRIC_ROLLUP_TABLE_REGEXP = /(?<METRIC_TYPE>metrics?_?.*)_(?<CHILD_TABLE_NUM>\d\d)$/.freeze
 
       def tables(stream)
-        super
+        ignore_inherited_tables { super }
+
+        inherited_tables(stream)
```
Fryguy: As discussed, instead of having a copy of the tables method, let's instead leverage your block yielding pattern twice in a row.


```suggestion
        ignoring_inherited_tables { super }
        ignoring_parent_tables { super }
```

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -41,10 +95,14 @@ def miq_metric_table_sequences(stream)
       # dumped prior to `metrics_base`, etc.
       #
       def miq_metric_table_inheritances(stream)
-        inherited_metrics_tables.each do |(table, inherit_from)|
+        return if Rails.version >= "8.0"
```
jrafanie: At this point, this seems to be the only rails 8 specific difference... inherits is handled by the schema dumper so we don't need to do this in rails 8.

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -84,28 +142,21 @@ def triggers(stream)
 
       private
 
-      def track_miq_metric_table_inheritance(table)
-        return unless table.match?(METRIC_ROLLUP_TABLE_REGEXP)
-        return unless (inherit_from = determine_table_parent(table))
-
-        inherited_metrics_tables << [table, inherit_from]
+      def base_tables
```
Fryguy: Move this method after base_inherited_tables_partition

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -4,19 +4,110 @@ module SchemaDumper
...
+          before = @ignore_tables.dup
+          @ignore_tables = @ignore_tables | inherited_table_names
+          puts "*** ignore_tables: #{@ignore_tables.inspect}"
+        end
+
+        # byebug
+
+        yield
+
+        @ignore_tables = before if Rails.version >= "8.0"
```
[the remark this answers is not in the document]
jrafanie: This is still true although now we do 2 passes: 
1) ignore_tables has the subtables so we can the first pass with the base tables
2) ignore_tables has the base tables so we can do a second pass with the inherited tables

lib/manageiq/schema/schema_dumper.rb
```diff
@@ -4,9 +4,9 @@ module SchemaDumper
       METRIC_ROLLUP_TABLE_REGEXP = /(?<METRIC_TYPE>metrics?_?.*)_(?<CHILD_TABLE_NUM>\d\d)$/.freeze
 
       def tables(stream)
-        ignore_inherited_tables { super }
+        ignoring_tables(inherited_tables, stream) { super }
+        ignoring_tables(base_tables, stream) { super }
```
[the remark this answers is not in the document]
jrafanie: I kept the stream parameter to consistent with the other method calls in this method.

jrafanie:
> * Support db:migrate:reset by looking for no versions(7.2) or 0(8.0)

☝️ 
@Fryguy I added another commit to drop the one TODO remaining.

Fryguy (approved):

PR #826, opened 2025-12-16
Split inherit/constraint for usage in collapsed migration/schema.rb

jrafanie:
We need to do inheritance in both migrations and from schema dump/load.

For constraints, we don't need to do them from schema load as t.check_constraints is now included for each metric_* subtable in the dumped schema.rb, so we no longer pass the conditions from schema.rb dump/load.

Simplify the interface to add_miq_metric_table_constraint to pass the array of conditions inline.

Note, this issue was found in #824 and extracted and simplified here.

### DIFF
--- a/lib/manageiq/schema/schema_dumper.rb
+++ b/lib/manageiq/schema/schema_dumper.rb
@@ -6,7 +6,7 @@ module ManageIQ
       def tables(stream)
         super
         miq_metric_table_sequences(stream)
-        miq_metric_table_constraints(stream)
+        miq_metric_table_inheritances(stream)
         miq_metric_views(stream)
         triggers(stream)
       end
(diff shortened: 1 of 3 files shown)


lib/manageiq/schema/schema_dumper.rb
```diff
@@ -52,8 +52,7 @@ def miq_metric_table_constraints(stream)
           conditions        = [primary_condition, "realtime", child_table_num]
 
           stream.puts "  add_miq_metric_table_inheritance #{child_table}, " \
-                          "#{inherit_from.inspect}, "                       \
-                          ":conditions => #{conditions.inspect}"
+                          "#{inherit_from.inspect} "
```
Fryguy: You can remove a number of the local variables now such as conditions, primary_condition, and child_table_num.

Fryguy:
@jrafanie , I found out that check constraints were added to migrations in Rails 6.1 (https://blog.saeloun.com/2021/01/08/rails-6-check-constraints-database-migrations/).

So, if we change collapsed_initial_migration version we don't have to go too far. (Not for this PR - but we can try in a follow up)

PR #852, opened 2026-04-01
Convert role/group names/descriptions to IDs in visibility for CustomButton and MiqWidget

jrafanie:
Summary: role names and group descriptions can change, so we can't associate buttons and widgets with them via the visibility column, we need to migrate these hardcoded names to something that will not change, the role or group's id.

## NOTE

- [x] Co-dependency... While this is standalone, it should be merged only if both this PR and https://github.com/ManageIQ/manageiq-ui-classic/pull/9954 are ready to go since the UI will continue to save group description/role names in the database if schema is merged but the UI is not.

###  Convert role names to IDs in visibility for CustomButton and MiqWidget

    - Preserve existing IDs when mixed with names
    - Return an empty array when no valid roles remain, adminsitrator can set
      visibility
    - Log warnings for invalid role names/IDs

### Convert group descriptions to IDs in visibility column

    Widgets store group visibility by description, which breaks when groups
    are renamed. This migration converts existing group descriptions to IDs.

    - Only affects MiqWidget (custom buttons don't support group visibility)
    - Return an empty array when no valid groups remain, adminsitrator can set
      visibility

### DIFF
--- a/db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
+++ b/db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
@@ -67,9 +67,14 @@ class ConvertRoleNamesToIdsInVisibility < ActiveRecord::Migration[8.0]
     valid_ids = []
     invalid_names = []
 
+    role_names = roles.select { |r| r.kind_of?(String) && r != "_ALL_" }
+
+    # fetch all roles at once
+    roles_by_name = MiqUserRole.in_my_region.where(:name => role_names).index_by(&:name)
+
     roles.each do |role_name_or_id|
       if role_name_or_id.kind_of?(String) && role_name_or_id != "_ALL_"
-        role = MiqUserRole.in_my_region.where(:name => role_name_or_id).first
+        role = roles_by_name[role_name_or_id]
         if role
           valid_ids << role.id
         else
@@ -103,9 +108,15 @@ class ConvertRoleNamesToIdsInVisibility < ActiveRecord::Migration[8.0]
     valid_names = []
     invalid_ids = []
 
+    # extract role IDs for the upfront query
+    role_ids = roles.select { |r| r.kind_of?(Integer) || (r.kind_of?(String) && r.match?(/^\d+$/)) }.collect(&:to_i)
+
+    # fetch all roles at once
+    roles_by_id = MiqUserRole.in_my_region.where(:id => role_ids).index_by(&:id)
+
     roles.each do |role_id|
       if role_id.kind_of?(Integer) || (role_id.kind_of?(String) && role_id.match?(/^\d+$/))
-        role = MiqUserRole.in_my_region.where(:id => role_id.to_i).first
+        role = roles_by_id[role_id.to_i]
         if role
           valid_names << role.name
         else


db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
```diff
@@ -0,0 +1,143 @@
+class ConvertRoleNamesToIdsInVisibility < ActiveRecord::Migration[8.0]
+  class CustomButton < ActiveRecord::Base
+    include ActiveRecord::IdRegions
+    self.inheritance_column = :_type_disabled
+    serialize :visibility
```
Fryguy: I thought you had to give a coder now for serialization?

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+  end
+
+  def build_role_ids(roles, record)
+    valid_ids = []
+    invalid_names = []
+
+    roles.each do |role_name_or_id|
+      if role_name_or_id.is_a?(String) && role_name_or_id != "_ALL_"
+        role = MiqUserRole.in_my_region.find_by("name = ?", role_name_or_id)
```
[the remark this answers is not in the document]
jrafanie: Good point.

db/migrate/20260401000000_convert_group_descriptions_to_ids_in_visibility.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    if invalid_descriptions.any?
+      record_info = record_identifier(record)
+      message = "WARNING: #{record_info} has invalid group descriptions that will be removed: #{invalid_descriptions.join(', ')}"
+      say(message, true)
+      Rails.logger.warn(message)
+    end
+
+    # Default to _ALL_ if no valid groups found
```
jrafanie: Same question as above... but now just for widgets

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
jrafanie: done

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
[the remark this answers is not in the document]
jrafanie: done

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
```diff
@@ -0,0 +1,145 @@
...
+    end
+
+    if invalid_names.any?
+      record_info = record_identifier(record)
+      message = "WARNING: #{record_info} has invalid role names that will be removed: #{invalid_names.join(', ')}"
+      say(message, true)
+      Rails.logger.warn(message)
+    end
+
+    # Default to _ALL_ if no valid roles found
```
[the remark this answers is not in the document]
jrafanie: As discussed, we now will leave the visibility as empty `[]` in the event no roles remain as visible to the object.  The admin can fix this by granting visibility to the right roles.

db/migrate/20260401000000_convert_group_descriptions_to_ids_in_visibility.rb
jrafanie: Same as above, but for groups.

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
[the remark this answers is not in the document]
jrafanie: done

db/migrate/20260401000000_convert_group_descriptions_to_ids_in_visibility.rb
jrafanie: done

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+  end
+
+  def build_role_ids(roles, record)
+    valid_ids = []
+    invalid_names = []
+
+    roles.each do |role_name_or_id|
+      if role_name_or_id.kind_of?(String) && role_name_or_id != "_ALL_"
+        role = MiqUserRole.in_my_region.where(:name => role_name_or_id).first
```
Fryguy: This is an N+1.  There should be a way to fetch all of it up front. Something like

```ruby
roles_names = roles.select { |role_name_or_id| role_name_or_id.kind_of?(String) && role_name_or_id != "_ALL_" }
role_ids_by_name = MiqUserRole.in_my_region.where(:name => roles_names).pluck(:name, :id).to_h

roles.each do |role_name_or_id|
  if role_name_or_id.kind_of?(String) && role_name_or_id != "_ALL_"
    role = role_ids_by_name[role_name_or_id]
...
```

Fryguy:
I found an N+1 - not sure it's all that important, because the number of records here is likely reasonable that it wouldn't matter, but might be worth fixing anyway. I think there's a simpler way to partition the records into valid and invalid ids, but offhand I couldn't think of a nicer way to do it, so it's fine.

db/migrate/20260331000000_convert_role_names_to_ids_in_visibility.rb
jrafanie: I'll see if there's an easy way to fix it

jrafanie: done

Fryguy:
Backported to `tal` in commit bb191a49128ee8ce38b7396adad62372053073d2.

```text
commit bb191a49128ee8ce38b7396adad62372053073d2
Date:   Mon Apr 13 09:06:43 2026 -0400

    Merge pull request #852 from jrafanie/fix-custom-buttons-widgets-referencing-renamed-roles
    
    Convert role/group names/descriptions to IDs in visibility for CustomButton and MiqWidget
    
    (cherry picked from commit e0214f86cbb49b1444dc7ac674748de84ef254aa)
```

PR #896, opened 2026-08-04
Add mem/proc attributes to hardwares

shashank-gowda:
Currently hardware table stores only Installed Memory and Installed Processor values. The existing attributes are

memory_mb - > The total physical memory present in the Physical Host (Installed Memory)
cpu_total_cores - > The total physical cores present in the Physical Host (Installed Procs)

Adding 4 more values related to processor and memory.

memory_mb_configured -> Portion of installed memory that is licensed (Configured Memory)
cpu_configured_cores -> Portion of installed cores that is licensed (Configured Procs)
memory_mb_available -> Configured memory that is currently free and can be assigned to VMs (Available Memory)
cpu_available_cores -> Configured proc that is currently free and can be assigned to VMs (Available Procs)

### DIFF
--- a/db/migrate/20260804060955_add_mem_proc_attributes_to_hardwares.rb
+++ b/db/migrate/20260804060955_add_mem_proc_attributes_to_hardwares.rb
@@ -1,0 +1,8 @@
+class AddMemProcAttributesToHardwares < ActiveRecord::Migration[8.0]
+  def change
+    add_column :hardwares, :memory_mb_configured, :integer
+    add_column :hardwares, :cpu_configured_cores, :integer
+    add_column :hardwares, :memory_mb_available, :integer
+    add_column :hardwares, :cpu_available_cores, :integer
+  end
+end


Fryguy:
@agrare Please review.

A few questions

1. Do we need UI changes for these to present them? (I presume this was added for collection, so I expect a provider PR for that).
2. @agrare Is this something we have on other providers that would be useful to add collection for?
3. Is this duplicating any existing thing like reservations?

spec/migrations/20260804060955_add_mem_proc_attributes_to_hardwares_spec.rb
agrare, line 1: We usually don't require spec tests for simple schema changes like add_column since that is just testing active_record

agrare:
>  memory_mb_available -> Configured memory that is currently free and can be assigned to VMs (Available Memory)

This seems like something which would change relatively often, if I power a VM off does that VM's memory become available on the host?  Or is memory sliced up and statically assigned?

shashank-gowda:
@Fryguy Not planning to show these values in UI. Once this is merged will have the provider changes also.

 @agrare No, even if the VM is powered off the memory remains allocated to the VM and is not available to the host.

agrare:
> Is this something we have on other providers that would be useful to add collection for?

No I'm not aware of any other providers that would make use of this.

agrare (approved):

PR #904, opened 2026-08-14
Change the datatype of cpu_available_cores and cpu_configured_cores

shashank-gowda:
Change the datatype of cpu_available_cores column  and cpu_configured_cores from int to float.

@miq-bot add-label bug

### DIFF
--- a/db/migrate/20260814085218_change_cpu_cores_data_type_in_hardwares.rb
+++ b/db/migrate/20260814085218_change_cpu_cores_data_type_in_hardwares.rb
@@ -1,0 +1,11 @@
+class ChangeCpuCoresDataTypeInHardwares < ActiveRecord::Migration[8.0]
+  def up
+    change_column :hardwares, :cpu_available_cores, :float
+    change_column :hardwares, :cpu_configured_cores, :float
+  end
+
+  def down
+    change_column :hardwares, :cpu_available_cores, :integer
+    change_column :hardwares, :cpu_configured_cores, :integer
+  end
+end


db/migrate/20260814085218_change_cpu_available_cores_in_hardwares.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeCpuAvailableCoresInHardwares < ActiveRecord::Migration[8.0]
+  def up
+    remove_column :hardwares, :cpu_available_cores
+    add_column :hardwares, :cpu_available_cores, :float
+  end
+
+  def down
+    remove_column :hardwares, :cpu_available_cores
+    add_column :hardwares, :cpu_available_cores, :integer
+  end
```
agrare: Better off using `change` instead of `up` and `down` if you're using an action that is reversible.  See https://guides.rubyonrails.org/active_record_migrations.html#using-the-change-method for the list but tldr you just have to specify the data type of the column that you are removing and rails will know how to add it back on the down migration.

db/migrate/20260814085218_change_cpu_available_cores_in_hardwares.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeCpuAvailableCoresInHardwares < ActiveRecord::Migration[8.0]
+  def change
+    change_column :hardwares, :cpu_available_cores, :float
```
Fryguy: change_column is not a reversible method.  If you're using change_column, you need an up and down (in the previous commit you had use remove/add which is why @agrare suggested to use change).

I think change_column is a better choice than remove/add, so ultimately, up/down + change_column feels like the best combination.

Fryguy (approved):
LGTM - down to an integer _technically_ will truncate any floats, but the very next migration down will delete the column anyway, so it doesn't matter.

Fryguy:
@shashank-gowda I looked at https://github.com/ManageIQ/manageiq-schema/pull/896, and there were 2 cpu columns added - should the other one also be a float for consistency?

shashank-gowda:
@Fryguy @agrare **cpu_configured_cores** will not have a float value.

shashank-gowda:
@Fryguy @agrare I have changed **cpu_configured_cores** to float as well.

agrare:
> cpu_configured_cores will not have a float value.

> I have changed cpu_configured_cores to float as well.

@shashank-gowda which one is it?

shashank-gowda:
> > cpu_configured_cores will not have a float value.
> 
> > I have changed cpu_configured_cores to float as well.
> 
> @shashank-gowda which one is it?

@agrare As @Fryguy suggested to have the consistency across the new CPU columns, I have changed the data type of `cpu_configured_cores` as well.

agrare:
@shashank-gowda okay I was confused because you told @Fryguy  that it wouldn't be a float and then changed it.

agrare (approved):

PR #906, opened 2026-09-11
 feat: store PowerVS instance health state and reason in dedicated vms columns

ramyalu-ibm:
IBM Cloud Power VS returns a health object per instance alongside the power status:
"health": {
"lastUpdate": "2026-09-10T14:26:37.759526",
"reason": "rmc_state - inactive",
"status": "WARNING"
}

To capture this health information, two new columns are added to the vms table:

health_state — stores health.status ("OK", "WARNING", "CRITICAL")
health_details — stores health.reason ("rmc_state - inactive")

Associated PR link:
1. Mapped PR link : https://github.com/ManageIQ/manageiq/pull/23975
2. Provider PR link: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/579


db/migrate/20260901000000_add_health_state_to_vms.rb
```diff
@@ -0,0 +1,6 @@
+class AddHealthStateToVms < ActiveRecord::Migration[8.0]
+  def change
+    add_column :vms, :health_state,   :string
+    add_column :vms, :health_details, :string
```
agrare: @ramyalu-ibm we have the rest of the columns in the `vms` table commented with a helpful description of their purpose.  Can you add `:comment => "..."` to both of these `add_column` statements with a description of what they are for?  Make sure to keep it "cross-provider" and not Power specific.

ramyalu-ibm: @agrare Thank you for response,  
yeah sure, i will update the PR accordingly

agrare (approved):

Fryguy:
Should these stay raw values or should they be normalized? If raw, should we prefix that to the column name?

ramyalu-ibm:
> Should these stay raw values or should they be normalized? If raw, should we prefix that to the column name?
@Fryguy - The column names health_state and health_details were suggested by the team based on the discussion in this PR. Both store raw values directly from the provider API — no normalization is applied.
That said, we're open to renaming if you think raw_health_state better fits the existing convention. What would you suggest?

agrare:
@Fryguy I think we can normalize these (at least health_state since I could see that being tied to e.g. an icon or a filter).

`Ok`, `Warning`, `Error`, `Critical`, etc... seem like reasonable values.  @ramyalu-ibm do you have the full list of possible `health.status` values?

Fryguy:
Yeah I would say write normalized values now, and we can add a raw column later if necessary.

Whatever values we choose can be enumerated on the model level for validation.

ramyalu-ibm:
> @Fryguy I think we can normalize these (at least health_state since I could see that being tied to e.g. an icon or a filter).
> 
> `Ok`, `Warning`, `Error`, `Critical`, etc... seem like reasonable values. @ramyalu-ibm do you have the full list of possible `health.status` values?

@Fryguy it is the same values which you provided. - OK
WARNING
CRITICAL
these are the possible values, which i have seen in API response.

Fryguy:
We probably should map anything else to "unknown"

db/migrate/20260901000000_add_health_state_to_vms.rb
```diff
@@ -0,0 +1,6 @@
+class AddHealthStateToVms < ActiveRecord::Migration[8.0]
+  def change
+    add_column :vms, :health_state,   :string, :comment => "The health status of the VM (e.g. OK, WARNING, CRITICAL)."
```
agrare: Can you update the comment to include the actual normalized values that we'll support?

```suggestion
    add_column :vms, :health_state,   :string, :comment => "The health status of the VM (e.g. Ok, Warning, Error, Critical, Unknown)."
```

We can use this list in a core PR to validate health_state is one of these values then.

ramyalu-ibm: @agrare yeah sure, i will update PR.

agrare (approved):
LGTM

db/migrate/20260901000000_add_health_state_to_vms.rb
ramyalu-ibm: @Fryguy @agrare 
now we are directly maping health_state to health.status and health_details to health.reason,
we shouldn't do that?, is that approach is fine?
could you please provide conclusion on this it will be helpfull for further implementation.

agrare:
On a side note I just noticed `20260901000000_add_health_state_to_vms.rb` did you use `rails generate:migration` to create this and got really really lucky on the timestamp?  Or did some other tool generate this?

db/migrate/20260901000000_add_health_state_to_vms.rb
agrare: @ramyalu-ibm I will comment on your provider PR but you just need your parser to map to the normalized values.

ramyalu-ibm:
> On a side note I just noticed `20260901000000_add_health_state_to_vms.rb` did you use `rails generate:migration` to create this and got really really lucky on the timestamp? Or did some other tool generate this?

https://www.manageiq.org/docs/guides/providers/adding_a_new_model 
followed this document.

agrare:
And you generated this 10 days ago?  For example if I run `rails generate migration Abcd` right now I get `db/migrate/20260911164236_abcd.rb` since the current time in UTC was `16:42:36`

ramyalu-ibm:
> And you generated this 10 days ago?

Noo today only, i dont know how it took that timestamp. Is that wrong timestamp?

agrare:
That's really weird then, it should be something like `20260911164236`.  It isn't a problem since we don't have any older migrations after this value but in general the timestamp should be when you generated it.

ramyalu-ibm:
> That's really weird then, it should be something like `20260911164236`. It isn't a problem since we don't have any older migrations after this value but in general the timestamp should be when you generated it.

From next time i will be aware of these stuffs.

